Issue #64, opened 2025-10-26
Alter gateway type to be an Enum


thivindu:
### Description

Alter gateway type to be an enum rather than is_ai_gateway boolean.

key: "functional_type"
values: "regular", "ai", "event" etc

Issue #63, opened 2025-10-26
Add API endpoint to fetch the Gateway status


thivindu:
### Description

The Management portal needs a API endpoint to poll frequently to get the status of the gateways. Implement a lightweight API resource that returns the status of all gateways. If gatewayId is given as a query param return the status if that gateway only.

Issue #57, opened 2025-10-24
Add the status of the Gateway in the organization gateway listing


thivindu:
### Description

Implement the health check of the gateway. Include the status of the gateway in the gateway details.

PR #65, opened 2025-10-26
Implement gateway status feature

thivindu:
## Purpose
Currently the Management portal is unaware on the status of the registered gateways. Whether they are active, connected to the control plane or not.
The isAIGateway property is too constricting and needs expansion into more gateway types.

Fixes: [wso2/api-platform/issues#57](https://github.com/wso2/api-platform/issues/57)
Fixes: [wso2/api-platform/issues#63](https://github.com/wso2/api-platform/issues/63)
Fixes: [wso2/api-platform/issues#64](https://github.com/wso2/api-platform/issues/64)

## Approach
This PR adds functionality to keep track of the gateway status, whether it is alive or disconnected from the Control Plane or not. And adds a status polling endpoint for efficient monitoring.

The `/api/v1/gateways/status` implements the health check of the gateway. The management portal polls this lightweight API resource frequently that returns the status of all gateways. If gatewayId is given as a query param return the status if that gateway only. The status of the gateway and the other properties are included in the gateway detail fetching API calls.

Alter gateway type to be a string enum rather than is_ai_gateway which is boolean.
This enum can have the values: `regular,` `ai`, `event`.

### DIFF
--- a/platform-api/spec/impls/gateway-management/artifacts/openapi-gateways.yaml
+++ b/platform-api/spec/impls/gateway-management/artifacts/openapi-gateways.yaml
@@ -4,7 +4,7 @@
 # UPDATED: October 26, 2025
 # Changes: Added new gateway properties (isCritical, isAIGateway, description, vhost, isActive),
 # added gateway status endpoint for polling, updated all response schemas to include new fields.
-# Replaced isAIGateway boolean with gatewayType enum for better extensibility.
+# Replaced isAIGateway boolean with functionalityType enum for better extensibility.
 
 paths:
   /gateways:
@@ -297,7 +297,7 @@ components:
         - displayName
         - vhost
         - isCritical
-        - gatewayType
+        - functionalityType
       properties:
         name:
           type: string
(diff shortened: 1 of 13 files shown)


platform-api/spec/impls/gateway-management/artifacts/openapi-gateways.yaml
```diff
@@ -77,6 +78,47 @@ paths:
         '500':
           $ref: '#/components/responses/InternalServerError'
 
+  /gateways/status:
```
malinthaprasan: /gateways/status and   /gateways/{id} are both GET calls which conflicts each other. Generally its not recommended to have such resources as it can cause incosistent behaviours. (when fronted via gateways/LBs with resource policies etc they can conflict)

Shall we incoporate this into /gateways itself? (include the status inside each gateway model)

malinthaprasan: Cool

platform-api/src/resources/openapi.yaml
```diff
@@ -1458,10 +1493,15 @@ components:
           type: boolean
           description: Whether the gateway is critical for production
           example: true
-        isAiGateway:
+        gatewayType:
```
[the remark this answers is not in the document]
thivindu: Fixed via - https://github.com/wso2/api-platform/pull/65/commits/43a9eb27bdda97bcbb42eb49058116e9a763f961

PR #735, opened 2026-01-21
Correct response code considering the API associated gateway deletion

slahirucd7:
## Purpose
$subject

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests
   > Details about the test cases and coverage

## Security checks
 - Followed secure coding standards in http://wso2.com/technical-reports/wso2-secure-engineering-guidelines? yes/no
 - Ran FindSecurityBugs plugin and verified report? yes/no
 - Confirmed that this PR doesn't commit any keys, passwords, tokens, usernames, or other secrets? yes/no

## Summary by CodeRabbit

* **Bug Fixes**
  * When attempting to delete a gateway that still has associated APIs, the system now returns a clear conflict response (HTTP 409) instead of a generic error, providing more accurate status feedback while keeping the existing error message content.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/platform-api/src/internal/handler/gateway.go
+++ b/platform-api/src/internal/handler/gateway.go
@@ -259,7 +259,7 @@ func (h *GatewayHandler) DeleteGateway(c *gin.Context) {
 		}
 		if errors.Is(err, constants.ErrGatewayHasAssociatedAPIs) {
 			utils.LogError("Gateway has associated APIs during deletion", err)
-			c.JSON(http.StatusBadRequest, utils.NewErrorResponse(409, "Conflict",
+			c.JSON(http.StatusConflict, utils.NewErrorResponse(409, "Conflict",
 				"The gateway has associated APIs. Please remove all API associations before deleting the gateway"))
 			return
 		}


platform-api/src/internal/handler/gateway.go
```diff
@@ -259,7 +259,7 @@ func (h *GatewayHandler) DeleteGateway(c *gin.Context) {
 		}
 		if errors.Is(err, constants.ErrGatewayHasAssociatedAPIs) {
 			utils.LogError("Gateway has associated APIs during deletion", err)
-			c.JSON(http.StatusBadRequest, utils.NewErrorResponse(400, "Bad Request",
+			c.JSON(http.StatusBadRequest, utils.NewErrorResponse(409, "Conflict",
```
thivindu: Shouldn't this be a bad request considering this  is a delete operation?

nisan-abeywickrama: @thivindu , this cannot be a bad request since this is not due to a incorrect request syntax right? 
409 seems to be the best options since it is used under the following scenario, "The 409 (Conflict) status code indicates that the request could not be completed due to a conflict with the current state of the target resource. "

shanakama: Based on the logged error , `409` is ideal for the situation. As there are some dependencies (APIs here) associated with the target and then it exactly a conflict.

thivindu: > What happens if user sends an incorrect GW is ? What's the response code we are going to show

If the user tries to delete an incorrect GW, if it does not exists we send a 404 and if the id is malformed we send a 400.

> @thivindu , this cannot be a bad request since this is not due to a incorrect request syntax right? 409 seems to be the best options since it is used under the following scenario, "The 409 (Conflict) status code indicates that the request could not be completed due to a conflict with the current state of the target resource. "

Thanks for the clarification. 409 would be more suitable in this case.

slahirucd7: > > What happens if user sends an incorrect GW is ? What's the response code we are going to show
> 
> If the user tries to delete an incorrect GW, if it does not exists we send a 404 and if the id is malformed we send a 400.
> 
> > @thivindu , this cannot be a bad request since this is not due to a incorrect request syntax right? 409 seems to be the best options since it is used under the following scenario, "The 409 (Conflict) status code indicates that the request could not be completed due to a conflict with the current state of the target resource. "
> 
> Thanks for the clarification. 409 would be more suitable in this case.

Yes. 409 code should be available to send for an incorrect GW UUID (malformed uuid).

thivindu (approved):

Issue #1066, opened 2026-02-11
Use published sdk go module instead of using local latest sdk in gateway build


renuka-fernando:
## Problem

The gateway components (`policy-engine`, `gateway-builder`, `gateway-controller`) use `replace` directives in their `go.mod` files to point to the local SDK source:

```go
replace github.com/wso2/api-platform/sdk => ../../sdk
```

However, the SDK (`github.com/wso2/api-platform/sdk`) is a **user-facing, published module** with its own release tags (`sdk/v0.3.0` through `sdk/v0.3.9`). External policy authors depend on specific published versions (e.g., `sample-policies/count-letters` uses `v0.3.8`). Using `replace` in the gateway build means:

1. **No reproducibility** — two builds at different times can produce different binaries depending on local SDK state
2. **No compatibility signal** — policy authors can't tell which SDK version the runtime is built against
3. **Interface drift risk** — unreleased SDK changes compile fine in the gateway but may mismatch with what published policy authors use
4. **No release discipline** — SDK changes get consumed without going through a proper release

In contrast, the `common` module is internal and unpublished (`v0.0.0`), so using `replace` for it is correct.

### 1. Use published SDK version in go.mod (no replace)

Remove the SDK `replace` directive from all gateway `go.mod` files and use the published version:

- `gateway/gateway-runtime/policy-engine/go.mod`
- `gateway/gateway-builder/go.mod`
- `gateway/gateway-controller/go.mod`

The `common` module replace directives remain unchanged.

### 2. Keep SDK in Docker build context

The Dockerfiles and Makefiles continue to `COPY --from=sdk` and pass `--build-context sdk=../../sdk`. During normal builds, Go downloads the published SDK from the module proxy and the copied source is unused. This enables CI to inject `replace` directives when needed to test local SDK changes.

### 3. Separate CI workflow for SDK changes

Create a new `sdk-integration-test.yml` workflow that:
- Triggers on `sdk/**` changes
- Injects `go mod edit -replace` to point gateway modules to local SDK
- Builds gateway images and runs integration tests against the modified SDK

This runs independently from `gateway-integration-test.yml`:

| Scenario | gateway-integration-test | sdk-integration-test |
|---|---|---|
| Only gateway changes | Runs against published SDK | Does not trigger |
| Only SDK changes | Does not trigger | Runs with local SDK (replace) |
| Both change in same PR | Runs against published SDK | Runs with local SDK (replace) |

PR #1067, opened 2026-02-11
Use published SDK version instead of local replace directives

renuka-fernando:
## Summary

Fix https://github.com/wso2/api-platform/issues/1066

- Replace local SDK `replace` directives with published version `v0.3.9` in gateway go.mod files
- Add separate CI workflow (`sdk-integration-test.yml`) to validate SDK changes against gateway integration tests

## Problem

Gateway components used `replace github.com/wso2/api-platform/sdk => ../../sdk` to always build against local SDK source. Since the SDK is a user-facing published module with external consumers (policy authors), this caused non-reproducible builds and masked potential interface drift between the gateway runtime and published SDK versions.

## Changes

**go.mod (3 files):**
- `gateway/gateway-runtime/policy-engine/go.mod` — removed SDK replace directive
- `gateway/gateway-builder/go.mod` — removed SDK replace, updated `v0.0.0` → `v0.3.9`
- `gateway/gateway-controller/go.mod` — removed SDK replace, updated `v0.0.0` → `v0.3.9`
- `common` module replace directives kept as-is (internal, unpublished module)

**CI (1 new file):**
- `.github/workflows/sdk-integration-test.yml` — triggers on `sdk/**` changes, injects `go mod edit -replace` to test local SDK against gateway integration tests

**No changes to Dockerfiles or Makefiles** — SDK source continues to be copied into Docker builds via `--build-context sdk`. During normal builds Go downloads the published version from module proxy; during SDK CI testing the injected replace directive makes Go use the local copy.

## Summary by CodeRabbit

* **Chores**
  * CI: Expanded integration-test PR triggers to include shared/common code and added a new SDK integration workflow that runs Go/Docker builds, integration tests, and diagnostic log collection.
  * Dependencies: Upgraded the SDK and several libraries across components.
  * Build: Removed local module replacement overrides to standardize build behavior.

### DIFF
--- a/gateway/gateway-controller/go.mod
+++ b/gateway/gateway-controller/go.mod
@@ -24,7 +24,6 @@ require (
 	github.com/wso2/api-platform/common v0.0.0
 	github.com/wso2/api-platform/sdk v0.3.14
 	github.com/xeipuuv/gojsonschema v1.2.0
-	golang.org/x/crypto v0.47.0
 	google.golang.org/grpc v1.78.0
 	google.golang.org/protobuf v1.36.11
 	gopkg.in/yaml.v3 v3.0.1
@@ -92,6 +91,7 @@ require (
 	go.opentelemetry.io/otel/sdk/metric v1.39.0 // indirect
 	go.yaml.in/yaml/v2 v2.4.2 // indirect
 	golang.org/x/arch v0.23.0 // indirect
+	golang.org/x/crypto v0.47.0 // indirect
 	golang.org/x/net v0.49.0 // indirect
 	golang.org/x/sync v0.19.0 // indirect
 	golang.org/x/sys v0.40.0 // indirect


.github/workflows/sdk-integration-test.yml
```diff
@@ -0,0 +1,74 @@
...
+on:
+  workflow_dispatch:
+  pull_request:
+    branches:
+      - main
+    paths:
+      - 'sdk/**'
+      - '.github/workflows/sdk-integration-test.yml'
+
+jobs:
```
dushaniw: shall we set the concurrency settings as well to prevent running builds for staled commits

dushaniw: added for others in https://github.com/wso2/api-platform/pull/1128

PR #1805, opened 2026-04-28
Fix issue where wildcard policy not engaging for provider

Arshardh:


gateway/gateway-controller/pkg/utils/llm_transformer_test.go
```diff
@@ -1006,6 +1006,80 @@ func TestTransformProvider_ExpandsWildcardPolicyPathWithTemplateMappings(t *test
...
+	chatOp := findOperation(result.Spec.Operations, "/chat/completions", "POST")
+	require.NotNil(t, chatOp)
+	require.NotNil(t, chatOp.Policies)
+
+	policyNames := make(map[string]bool)
+	for _, policy := range *chatOp.Policies {
+		policyNames[policy.Name] = true
+	}
+
+	assert.True(t, policyNames["set-headers-all"], "/chat/completions should inherit the wildcard policy")
```
nimsara66: Shall we check the ordering as well?
Ideally based on the given ops, for `/chat/completions`, `set-headers-all` should come first right?

nimsara66 (approved):

PR #1855, opened 2026-05-03
Add application association support

Arshardh:


malinthaprasan (approved):

platform-api/src/internal/service/application.go
```diff
@@ -328,6 +394,24 @@ func (s *ApplicationService) AddMappedAPIKeys(appIDOrHandle string, req *api.Add
...
+	if err != nil {
+		return nil, err
+	}
+
+	targetUUIDs, err := s.resolveAssociationTargets(req.Associations, app, orgID)
+	if err != nil {
+		return nil, err
+	}
+
+	if err := s.appRepo.AddApplicationAssociations(app.UUID, targetUUIDs); err != nil {
```
malinthaprasan, line 408: Dont we need to communicate the associations to the gateway?

Arshardh: No, we only update the gateway when a key is mapped to an associated entity

Issue #1987, opened 2026-05-18
[Task]: Add API Platform Gateway- Production Deployment Guide


Thushani-Jayasekera:
### Please select the area the issue is related to

Area/Management (Management API or Management Portal UI)

### Please select the aspect the issue is related to

Aspect/Other (Anything else which does not match above categories)

### Description

$subject

Thushani-Jayasekera:
https://docs.google.com/document/d/1T3EKqt9SWnOFgXAtR7Oe1WWnmoGZ7gAmZGIqIajIUxg/edit?tab=t.0

Thushani-Jayasekera:
Locally tested setup:

  ```
╔══════════════════════════════════════════════════════════════════════════════════╗
  ║                    KIND DATAPLANE CLUSTER  (Kubernetes v1.35.0)                 ║
  ║                                                                                  ║
  ║  ┌─────────────────────┐  ┌─────────────────────┐  ┌─────────────────────────┐  ║
  ║  │  dataplane-         │  │  dataplane-worker   │  │  dataplane-worker2      │  ║
  ║  │  control-plane      │  │  172.26.0.3         │  │  172.26.0.4             │  ║
  ║  │  172.26.0.2         │  │                     │  │                         │  ║
  ║  │  (control-plane)    │  │  containerd 2.2.0   │  │  containerd 2.2.0       │  ║
  ║  └─────────────────────┘  └─────────────────────┘  └─────────────────────────┘  ║
  ║                                                                                  ║
  ╠══════════════════════════════════════ NAMESPACE: ap-gateway ═════════════════════╣
  ║                                                                                  ║
  ║  ┌─────────────────────────────────────────────────────────────────────────────┐ ║
  ║  │  INGRESS / EXTERNAL TRAFFIC                                                 │ ║
  ║  │                                                                             │ ║
  ║  │   Internet ──► LoadBalancer: my-gateway-gateway-runtime (PENDING ext-IP)   │ ║
  ║  │                  HTTP  :8080 → NodePort 31320                               │ ║
  ║  │                  HTTPS :8443 → NodePort 31608                               │ ║
  ║  │                  Admin :9003 → NodePort 32696                               │ ║
  ║  └──────────────────────────────┬──────────────────────────────────────────────┘ ║
  ║                                 │                                                ║
  ║  ┌──────────────────────────────▼──────────────────────────────────────────────┐ ║
[...]

PR #1998, opened 2026-05-21
Enhance gateway configuration by conditionally including OAuth2 client ID and secret in the Helm chart template

Thushani-Jayasekera:
## Purpose

This is not a mandatory fix - as we can ommit these values without adding even empty values to the values.yaml

https://github.com/wso2/api-platform/issues/1987

Without the conditionals, the template would render:

  apim_oauth2_client_id = ""
  apim_oauth2_client_secret = ""

An explicit empty string is a value — it gets written to the ConfigMap and the controller reads it from the TOML file. Most Go config frameworks (Viper, etc.) won't fall back to an env var when the key is present in the config file, even if the value is empty. The env var would be silently ignored.

By making the fields conditional, those lines are omitted entirely from config.toml. Only then — when the key is absent from the file — does the controller fall back to its env var override (APIP_GW_CONTROLPLANE_APIM_OAUTH2_CLIENT_ID).

So the conditionals are load-bearing. Without them: controller auth fails with an empty client ID. With them: the TOML is silent on those fields, env vars win.

### DIFF
--- a/kubernetes/helm/gateway-helm-chart/templates/gateway/gateway-config.yaml
+++ b/kubernetes/helm/gateway-helm-chart/templates/gateway/gateway-config.yaml
@@ -69,8 +69,12 @@ data:
     deployment_push_enabled = {{ $gc.controlplane.deployment_push_enabled }}
     sync_batch_size = {{ $gc.controlplane.sync_batch_size }}
     gateway_name = {{ $gc.controlplane.gateway_name | quote }}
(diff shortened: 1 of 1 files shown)


renuka-fernando (approved):

renuka-fernando:
> Most Go config frameworks (Viper, etc.) won't fall back to an env var when the key is present in the config file, even if the value is empty. The env var would be silently ignored.

We are using Koanf, it should override the config from the env var.

PR #2010, opened 2026-05-22
[WIP] docs(helm): add values reference for gateway chart

Thushani-Jayasekera:
## Purpose
https://github.com/wso2/api-platform/issues/1987

### DIFF
--- a/kubernetes/helm/gateway-helm-chart/README.md
+++ b/kubernetes/helm/gateway-helm-chart/README.md
@@ -133,7 +133,7 @@ All configurable values are documented in `values.yaml`. Component blocks are fu
 - `gateway.controller.upstreamCerts.*` – Custom CA certificates for upstream backend TLS verification.
 - `gateway.config.policy_engine.*` – policy engine configuration including xDS client settings and admin API.
 
-Refer to the inline comments inside `values.yaml` for a complete matrix of options and the expected data types for each block.
+Refer to the inline comments inside `values.yaml` for a complete matrix of options and the expected data types for each block. For a full parameter reference with types, defaults, and descriptions, see [values.md](./values.md).
 
 ## TLS Certificate Configuration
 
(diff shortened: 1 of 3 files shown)


renuka-fernando:
Thanks for putting this together @Thushani-Jayasekera — really appreciate the effort to make the chart's configuration more discoverable. 🙌

Before we merge, I'd like to gently raise a concern and propose an alternative for discussion.

**The concern: two sources of truth**

Our `values.yaml` is already quite well-commented (~655 lines, with descriptive comments on most blocks). Adding `values.md` means every future change to a parameter — a new key, a renamed field, a changed default — has to be made in two places. There's no tooling to enforce that, so drift is almost inevitable over time. CodeRabbit's review already surfaced a couple of small inconsistencies (`certificateProvider: secret` vs `existing-secret`, and a broken table row in the Router section), which I think nicely illustrates the maintenance cost.

**What most Helm charts do**

The community convention is to keep `values.yaml` as the single source of truth and **auto-generate** the parameter reference. Bitnami, Grafana, Prometheus community, cert-manager, ingress-nginx, ArgoCD, Cilium — all use [`helm-docs`](https://github.com/norwoodj/helm-docs) (or an equivalent generator) driven by annotated `values.yaml` comments.

**Proposed alternative**

Instead of `values.md`, could we:

1. Adopt `helm-docs` — most of our existing `values.yaml` comments just need a small `# --` prefix tweak to be picked up.
2. Use a `README.md.gotmpl` template so the hand-written narrative (TLS guide, etc.) stays curated, and only the parameter table is generated.
3. Add a small CI check (`git diff --exit-code` after running `helm-docs`) to prevent drift.

That way we get the same — arguably better — reference table you've built here, but it's guaranteed to stay in sync with `values.yaml`, and contributors only have to update one file.

Happy to help wire up the `helm-docs` setup if it's useful. WDYT? If there's a reason a hand-maintained `values.md` is preferred here that I'm missing, happy to hear it.

PR #2012, opened 2026-05-22
Add per-operation upstream override for REST API operations

mehara-rothila:
## Summary

Adds a per-operation `upstream` override on REST API operations: an operation can route its `main` and/or `sandbox` traffic to a different backend than the API-level upstream. Per-op targets are **ref-only**: each references a named entry in `spec.upstreamDefinitions` rather than carrying an inline URL. Operations without a per-op upstream fall back to the API-level upstream, exactly as before.

A per-op route **reuses the referenced `upstreamDefinition`'s cluster** (one cluster per definition, so operations sharing a `ref` share it) rather than minting a cluster per operation. API-level `main`/`sandbox` cluster names are **URL-stable** (derived from a hash of the API's identity, not the URL). Both are URL-independent, so URL edits update Envoy endpoints in place instead of recreating clusters.

## Why

Real-world APIs often need individual operations to reach different backend services. Today that needs path-based workarounds or splitting into multiple APIs. This lets the API definition declare per-operation routing directly, while backend URLs stay defined once in `upstreamDefinitions` and are referenced by name.

## Design decisions

**Ref-only at the operation level.** `operations[].upstream.main` / `.sandbox` carry only a `ref` to a named `upstreamDefinition`, with no inline `url` and no per-target timeout/hostRewrite. Backends are declared once in `spec.upstreamDefinitions`; connect timeout lives on the definition. (API-level `upstream` still accepts either `url` or `ref`, unchanged.)

**Per-op routes reuse the definition cluster.** A per-op `ref` does not create a new cluster. It points the operation's route at the cluster the referenced `upstreamDefinition` already owns (`upstream_<kind>_<apiID>_<defName>`, built once per definition). N operations sharing a `ref` cost one cluster, not N, and the route inherits the definition's authoritative base path. `cluster_header` stays on with that cluster as the default, so a dynamic-endpoint policy can still override the operation (precedence: op-policy > api-policy > per-op ref > api-level upstream).

**URL-stable API-level cluster naming.** API-level `main`/`sandbox` cluster names derive from the full `sha256(apiID)` (64 hex), via `clusterkey.HashedName(env, apiID)`, producing `main_<hash>` / `sandbox_<hash>`. The `env` is a plaintext prefix, not part of the hash, so `main` and `sandbox` share the same hash. The URL is intentionally excluded, so a URL edit keeps the same cluster name and Envoy updates its endpoints in place (no connection drain) rather than recreating the cluster. The clusters are `STRICT_DNS` with an inline load assignment. *Rollout:* a one-time cluster rebuild on the first config push after upgrade (URL-derived to identity-based names), handled gracefully by Envoy's CDS warming; stable thereafter.

**Single source of truth.** Ref resolution and connect-timeout parsing go through `pkg/utils/upstreamref` (`FindByName`, `ParseConnectTimeout`); API-level cluster-name hashing goes through `pkg/utils/clusterkey` (`Hash` returns the full SHA-256 hash, `HashedName` joins a prefix to it). Both are consumed by the validator, the RDC transformer, and the xDS translator so they cannot drift. The policy-engine kernel (a separate Go module that cannot import `clusterkey`) reconstructs the definition cluster name for dynamic-endpoint targets; a kernel unit test pins its output byte-for-byte to `clusterkey.DefinitionName` so the two modules cannot drift either.

## What changed (gateway-controller layer)

- **OpenAPI schema** (`management-openapi.yaml`): optional `upstream` on `Operation` resolving to `OperationUpstream` (`main`/`sandbox`), each an inline ref-only object (`{ ref }`, no named target schema). Wrapper and each sub-field locked with `additionalProperties: false`, plus `minProperties: 1` on the wrapper (an empty `upstream: {}` is rejected).
- **Validator** (`api_validator.go`): rejects an empty `upstream: {}` wrapper; requires at least one of `main`/`sandbox`; rejects a `ref` that does not resolve; and enforces one name contract (`^[a-zA-Z0-9\-_]+$`, max 100 chars) on per-op refs, API-level refs, and `upstreamDefinition` names so every valid definition name is referenceable.
- **Transform** (`transform/restapi.go`) and **xDS translator** (`xds/translator.go`): per-op routes reuse the referenced definition's cluster (`upstream_<kind>_<apiID>_<defName>`, built once per definition); API-level clusters use URL-stable names `main_<hash>`/`sandbox_<hash>`; operations without a per-op upstream fall back to the API-level cluster.
- **Shared utils** (`pkg/utils/upstreamref`, `pkg/utils/clusterkey`): new leaf packages. `upstreamref` handles ref resolution and connect-timeout parsing; the stdlib-only `clusterkey` handles URL-stable API-level cluster-key hashing (`Hash` / `HashedName`) and definition cluster naming (`DefinitionName`).

## Cluster names: before / after

For an API with UUID `0190b3e2-7b1c-7c2a-9b3d-1a2b3c4d5e6f` (`sha256(apiID)` = `54a9b3e5ce2b6ccb97168e5948a66f48e084213b38eb8c7dc01c6f624a63c2f2`):

| Cluster | Before | After |
|---|---|---|
| API-level main | `cluster_http_default-backend_8080` (URL-derived) | `main_54a9b3e5ce2b6ccb97168e5948a66f48e084213b38eb8c7dc01c6f624a63c2f2` (identity-based) |
| API-level sandbox | `cluster_http_sandbox-backend_8080` (URL-derived) | `sandbox_54a9b3e5ce2b6ccb97168e5948a66f48e084213b38eb8c7dc01c6f624a63c2f2` (identity-based) |
| Definition `user-service` | `upstream_RestApi_0190b3e2-7b1c-7c2a-9b3d-1a2b3c4d5e6f_user-service` | unchanged (same name, now built via the shared `clusterkey.DefinitionName` helper) |

Only the API-level names change. Editing a backend URL previously renamed the API-level cluster, so Envoy dropped and recreated it; with identity-based names the cluster name never changes and Envoy updates the endpoints in place. Definition cluster names were already URL-independent.

## Test coverage

- **Unit tests** across `pkg/config`, `pkg/transform`, `pkg/xds`, `pkg/utils/upstreamref`, and `pkg/utils/clusterkey`: deterministic hash tests (`Hash`), validator rejection tests (empty/unknown/pattern/length refs, with exact error reasons), URL-stable cluster-name contract tests (same name across a URL edit), definition-cluster reuse and cross-API isolation tests, per-op route tests asserting the exact referenced definition cluster, a policy chain-order test (API-level policies precede operation-level, so the operation-level policy wins as the last write), and a policy-engine kernel test pinning the `x-target-upstream` value for a dotted/coloned definition name byte-for-byte to the controller's `clusterkey.DefinitionName`.
- **Integration tests** (godog) across the per-op and URL-stable feature files: API-level fallback, per-op main/sandbox ref overrides, routing to two distinct real backends, validation rejections (strict 400), dynamic-endpoint precedence over per-op refs at both operation and API level, and URL-stable in-place endpoint updates. All passing against a local docker-compose stack.

## Example

One API exercising every capability: API-level fallback, per-op main override, per-op main+sandbox overrides, definition base path and connect timeout, host rewrite, and a dynamic-endpoint policy overriding a per-op ref.

```yaml
apiVersion: gateway.api-platform.wso2.com/v1
kind: RestApi
metadata:
  name: shop-api-v1.0
spec:
  displayName: Shop API
  version: v1.0
  context: /shop/$version
  vhosts:
    main: shop.example.com
    sandbox: sandbox.shop.example.com

  # Named backends, declared once, referenced everywhere
  upstreamDefinitions:
    - name: user-service
      basePath: /api/v2                 # prepended to requests routed to this backend
      timeout:
        connect: 5s                     # connect timeout lives on the definition (per-op targets are ref-only)
      upstreams:
        - url: http://user-service:8080
    - name: order-service
      timeout:
        connect: 500ms
      upstreams:
        - url: http://order-service:8080
    - name: promo-service
      upstreams:
        - url: http://promo-service:8080

  # API-level default upstream (url or ref; hostRewrite lives at this level)
  upstream:
    main:
      url: http://default-backend:8080
      hostRewrite: auto
    sandbox:
      ref: user-service                 # API-level upstream can also use a ref

  operations:
    # 1. No per-op upstream: falls back to the API-level upstream
    - method: GET
      path: /products

    # 2. Per-op main override: routes to user-service, inheriting its basePath and timeout
    - method: GET
      path: /users/{id}
      upstream:
        main:
          ref: user-service

    # 3. Per-op main and sandbox overrides: different backends per environment
    - method: POST
      path: /orders
      upstream:
        main:
          ref: order-service
        sandbox:
          ref: user-service

    # 4. Per-op ref plus dynamic-endpoint policy: the policy wins over the ref
    - method: GET
      path: /recommendations
      upstream:
        main:
          ref: user-service
      policies:
        - name: dynamic-endpoint
          version: v1
          params:
            targetUpstream: promo-service
```

Routing result:

- `GET /products` goes to `default-backend` (no per-op upstream, API-level fallback)
- `GET /users/{id}` goes to `user-service` under `/api/v2` (per-op main ref, inherits the definition's basePath)
- `POST /orders` goes to `order-service` on main and `user-service` on sandbox (per-op overrides per environment)
- `GET /recommendations` goes to `promo-service` (the operation-level dynamic-endpoint policy overrides the per-op ref)


mehara-rothila:
@coderabbitai review

mehara-rothila:
@coderabbitai review

mehara-rothila:
@coderabbitai review

mehara-rothila:
@coderabbitai review

mehara-rothila:
@coderabbitai review

mehara-rothila:
@coderabbitai review

mehara-rothila:
@coderabbitai review

ashera96:
@coderabbitai please review

renuka-fernando:
@coderabbitai full review

mehara-rothila:
@coderabbitai review

mehara-rothila:
@coderabbitai I'd like your opinion on a design decision in the operation-level upstream schema.

Context (gateway-controller/api/management-openapi.yaml):
- API-level `Upstream` allows `oneOf: [url, ref]`, so a backend can be given as a raw URL or as a `ref` to a named `upstreamDefinitions` entry.
- Operation-level `RestAPIOperationUpstream` has the same `main` and `sandbox` sub-fields as the API level, BUT each sub-field is a `RestAPIOperationUpstreamTarget`, which is `required: [ref]` (ref-only). A raw inline `url` is intentionally NOT allowed at the operation level.

Question: is restricting per-operation overrides to ref-only (no inline `url`) the right call, or should we mirror the API level and also allow `url` here?

Please weigh in on these specific angles by looking at the actual code:
1. Envoy/EDS cluster stability: does ref-only (one named definition = one reusable, stable cluster) actually avoid cluster churn vs. allowing an ad-hoc cluster per inline URL? Trace how the xDS translator builds clusters from upstream-definition refs.
2. Feature parity: `upstreamDefinitions` carry `basePath`, `timeout`, and weighted `upstreams` (load balancing). If we allowed a raw per-op `url`, would those settings be unreachable at the operation level?
3. Validation / single source of truth: is validating all backend URLs once in `spec.upstreamDefinitions` and referencing by name simpler and less error-prone than re-validating inline URLs on every operation?
4. Any downsides, edge cases, or inconsistencies you see in keeping operation-level ref-only while the API level stays url-or-ref?

Please confirm whether the ref-only restriction is sound, or flag concerns.

mehara-rothila:
@coderabbitai review the per-operation upstream schema (`RestAPIOperationUpstream` + `RestAPIOperationUpstreamTarget`) in `gateway/gateway-controller/api/management-openapi.yaml`. Is it well-formed and consistent with how the Go validator and transform/xds consume it? Flag any changes, or confirm it's fine. Give your own independent read.

gateway/gateway-controller/api/management-openapi.yaml
```diff
@@ -4313,6 +4313,41 @@ components:
...
+      additionalProperties: false
+      description: Per-operation upstream override. Each sub-field must reference a named entry in spec.upstreamDefinitions. Missing sub-fields fall back to API-level upstream. At least one of main or sandbox must be set.
+      anyOf:
+        - required: [main]
+        - required: [sandbox]
+      properties:
+        main:
+          description: Production vhost override. Must reference a named upstreamDefinition.
+          allOf:
+            - $ref: "#/components/schemas/RestAPIOperationUpstreamTarget"
```
renuka-fernando: Remove allof

gateway/gateway-controller/api/management-openapi.yaml
```diff
@@ -4313,6 +4313,41 @@ components:
...
+        sandbox:
+          description: Sandbox vhost override. Must reference a named upstreamDefinition.
+          allOf:
+            - $ref: "#/components/schemas/RestAPIOperationUpstreamTarget"
+
+    RestAPIOperationUpstreamTarget:
```
renuka-fernando: Resuse this target in upstream as well.

mehara-rothila: @coderabbitai design question on the upstream-ref name contract. Three schemas in `gateway/gateway-controller/api/management-openapi.yaml` carry the same ref/name rule (`pattern: ^[a-zA-Z0-9\-_]+$`, `minLength: 1`, `maxLength: 100`): `UpstreamDefinition.name`, the per-operation `RestAPIOperationUpstreamTarget.ref` (a ref-only object `{ ref }`), and the API-level `Upstream.ref` (a plain string property, which we just gave the same constraints inline). A reviewer suggested renaming `RestAPIOperationUpstreamTarget` to a generic name and reusing it for the API-level `Upstream.ref` so the contract is defined once. Note the per-op target is an **object** while `Upstream.ref` is a **string** property, so they can't share the same object schema directly. What's the cleanest way (if any) to define this ref contract once across an object field and a string field? And in this codebase, what would extracting a shared `$ref` string schema do to the generated Go types (oapi-codegen) and to the code that consumes `Ref`/`Name` as plain strings (e.g. `upstreamref.FindByName`, `strings.TrimSpace`)? Would you recommend extracting a shared schema, or keeping the constraints inline? Please give your own independent assessment.

mehara-rothila:
@coderabbitai Investigate the policy-building flow in the wso2/api-platform gateway-controller (Go) on the `feat/per-op-upstream-gateway` branch. I want an independent answer to whether `DerivePolicyFromAPIConfig` is actually used, and what the real policy path is.

Context , policy building appears to exist in two places:
1. `pkg/policy/builder.go` , `DerivePolicyFromAPIConfig(...)` returns `*models.StoredPolicyConfig` (held in `pkg/storage/policy_store.go`). Its doc comment claims it is used by APIDeploymentService, APIServer handlers, and main.go startup, but it also carries a TODO: "APIServer handlers (REST API path) - TODO: Refactor this to use the implementation."
2. `pkg/transform/restapi.go` , `RestAPITransformer.Transform` builds policy chains via `buildPolicyChain(...)` into `rdc.PolicyChains`, which `pkg/policyxds` consumes and pushes to the policy engine.

A grep across the whole gateway directory found NO callers of `DerivePolicyFromAPIConfig` (only its own definition in `builder.go`).

Please verify independently and report, with file:line evidence:
1. Is `DerivePolicyFromAPIConfig` (`builder.go`) called anywhere in the codebase? If yes, where exactly? If not, is it dead/orphaned code? Also check whether anything ever writes to `policy_store`.
2. What is the ACTUAL active path that builds per-route policy chains and feeds the policy engine for a deployed REST API (trace it from the deploy event to the policy engine)? Is it the `restapi.go` transform -> `rdc.PolicyChains` -> `policyxds` path, or `builder.go`, or both?
3. The per-op feature modified the sandbox vhost-gating in `builder.go` (added `perOpSandboxHasContent`). Is that copy on the active path, or did it edit an unused function? Is the equivalent gating in `restapi.go` (and `xds/translator.go`) the one actually exercised at runtime?
4. Bottom line: did the per-op work correctly wire per-op (incl. per-op sandbox) policy building through the active path, or is there a real gap (e.g., `builder.go` was supposed to be the path but is never called)? Are we missing anything, or is building it only in the transform path correct?

Give your own independent verdict.

gateway/gateway-controller/api/management-openapi.yaml
mehara-rothila: Done in `3e608d230`. Extracted a shared `UpstreamReference` string schema and pointed all three ref/name fields at it via a bare `$ref`: `UpstreamDefinition.name`, the API-level `Upstream.ref`, and the per-operation target's `ref`.

Since `Upstream.ref` is a string and the per-op target is an object (`{ ref }`), the shared piece is the scalar `UpstreamReference`, referenced by both. The `allOf` wrapper is gone, so each field is a direct `$ref` now.

oapi-codegen emits `type UpstreamReference = string`, so the generated Go and its consumers (`upstreamref.FindByName`, `strings.TrimSpace`) are unchanged.

mehara-rothila:
@coderabbitai review

I've applied a round of fixes from an external review on this per-operation upstream PR. Please give your own independent assessment of whether the current state is correct. Verify each of these against the actual code and flag anything wrong, inconsistent, or newly broken, with `file:line`:

- **Connect-timeout validation:** the regex is now `^\d+(\.\d+)?(ms|s|m|h)$` and the rejection message describes the single-unit, unsigned constraint. Is the validation correct and does it match the OpenAPI pattern?
- **Per-op definition cluster naming:** extracted into `clusterkey.DefinitionName`, used at all controller sites, with the two old sanitizers removed. Does it produce the exact same cluster name as before (no 503 / NoRoute risk)?
- **Removed field:** the unused `UpstreamCluster.ConnectTimeout` field and its reader were removed. Confirm nothing still depends on them.
- **Shared schema:** the `UpstreamReference` schema used by `UpstreamDefinition.name`, `Upstream.ref`, and the per-op target `ref`.

Don't assume the changes are right; give a skeptical, independent read.

gateway/it/features/per-op-upstream-basic.feature
```diff
@@ -0,0 +1,465 @@
...
+            basePath: /op-main
+        upstream:
+          main:
+            url: http://sample-backend:9080/api-main
+        operations:
+          - method: GET
+            path: /users
+            upstream:
+              main:
+                ref: op-main-svc
```
renuka-fernando: Let's add another operation with no ref, so this will test, `upstreamDefinitions` is defined and a operation is not using upstreamDefinitions, so the request should be proxied to upstream.main

gateway/it/features/per-op-upstream-basic.feature
```diff
@@ -0,0 +1,465 @@
...
+          main: per-op-basic-os-main.local
+          sandbox: per-op-basic-os-sandbox.local
+        upstreamDefinitions:
+          - name: op-sandbox-svc
+            upstreams:
+              - url: http://sample-backend:9080
+            basePath: /op-sandbox
+        upstream:
+          main:
+            url: http://sample-backend:9080/api-main
```
renuka-fernando: Maybe we can send a 400 bad request when sandbox vhost is defined and upstream.sandbox is not defined. Discussion: https://github.com/wso2/api-platform/discussions/369#discussioncomment-17188711

gateway/it/features/per-op-upstream-ref.feature
```diff
@@ -0,0 +1,338 @@
...
+      spec:
+        displayName: Per-Op-Ref-API
+        version: v1.0
+        context: /per-op/$version
+        vhosts:
+          main: per-op-main.local
+        upstreamDefinitions:
+          - name: users-svc
+            upstreams:
+              - url: http://sample-backend:9080
```
renuka-fernando:
```suggestion
              - url: http://sample-backend-b:9081
```

Let's use a different backend, so that we can confirm, the requests are not hitting the same Envoy cluster.

renuka-fernando: So we are not validating just from basepath

gateway/it/features/per-op-upstream-ref.feature
```diff
@@ -0,0 +1,338 @@
...
+    And I send a GET request to "http://localhost:8080/per-op/v1.0/orders"
+    Then the response should be successful
+    And the response should be valid JSON
+    And the JSON response field "path" should be "/order-svc/orders"
+
+    Given I authenticate using basic auth as "admin"
+    When I delete the API "per-op-ref-api-v1.0"
+    Then the response should be successful
+
+  Scenario: Per-operation sandbox ref routes independently of main
```
renuka-fernando: This is a duplicate test, already covered in basic feature file

gateway/it/features/per-op-upstream-ref.feature
```diff
@@ -0,0 +1,338 @@
...
+        operations:
+          - method: GET
+            path: /users
+            upstream:
+              main:
+                ref: non-existent-upstream
+      """
+    Then the response should be a client error
+    And the response should be valid JSON
+    And the JSON response field "status" should be "error"
```
renuka-fernando: Check status code as well, should be 400 bad request.

gateway/it/features/per-op-upstream-ref.feature
```diff
@@ -0,0 +1,338 @@
...
+            basePath: /ref-svc
+          - name: global-policy-svc
+            upstreams:
+              - url: http://sample-backend:9080
+            basePath: /global-policy-svc
+        upstream:
+          main:
+            url: http://sample-backend:9080/api-main
+        policies:
+          - name: dynamic-endpoint
```
renuka-fernando: Let's check the "Request Rewrite" policy as well, with rewrite path

gateway/it/features/per-op-upstream-ref.feature
mehara-rothila: @coderabbitai, Can u check and tell me have i addressed this correctly in the last commit

gateway/it/features/per-op-upstream-ref.feature
mehara-rothila: @coderabbitai, Can u check and tell me have i addressed this correctly on our PR

gateway/it/features/per-op-upstream-basic.feature
mehara-rothila: @coderabbitai , can u check commit 1372c5b5e and verify fixed this ?

mehara-rothila: @coderabbitai, if we add that 400 validation (reject a config where a sandbox vhost is set but there's no API-level upstream.sandbox and no per-op sandbox ref), would it conflict with any existing unit or integration tests in the repo?

gateway/it/features/per-op-upstream-basic.feature
mehara-rothila: Done , added a no-ref /orders operation to the "overrides API-level main" scenario. With upstreamDefinitions defined, that op falls back to API-level upstream.main, and the test asserts path: /api-main/orders

mehara-rothila:
@coderabbitai review

mehara-rothila:
@coderabbitai review

Please do a focused pass on code duplication only (not correctness - the code is
tested and green). For each changed file, and between changed files and the
pre-existing code around them, flag:

- logic written twice that could be one shared helper or constant
- repeated string/number literals that should be centralized
- loop-invariant work recomputed per iteration

For each finding: file:line, why it duplicates, and a concrete shape to fix it.

Known and intentionally left as-is, please skip these:
- clusterkey.DefinitionName kept separate from Hash/HashedName 
- perOpDefinitionClusterKey vs resolvePerOpDefinitionCluster (different outputs, cross-module)
- the policy-engine kernel's mirrored cluster-name builder (separate module, byte-identity pinned)

Cluster-name outputs must stay byte-identical - don't propose anything that changes them.

mehara-rothila:
@coderabbitai Thanks, but please re-run scoped to THIS PR's code only.

Only flag duplication where BOTH occurrences are in code this PR adds or modifies (check against the diff base d891a12e4). Do NOT flag pre-existing base functions or duplication between our code and base code - specifically skip validateUpstream, validateUpstreamUrl, validateUpstreamRef, validateUpstreamDefinitions (api_validator.go), resolveUpstreamURL (restapi.go), and resolveUpstreamCluster (translator.go) - those predate this PR and are out of scope.

In-scope surfaces are only: the upstreamref package, validateOperationUpstreamRef and validateUpstreamRefName, perOpDefinitionClusterKey, resolvePerOpDefinitionCluster, the sandbox helpers HasContent/SandboxActive, the clusterkey Hash/HashedName helpers, and the per-op route loops we added in restapi.go and translator.go.

Still exclude (by design): clusterkey.DefinitionName, the perOpDefinitionClusterKey vs resolvePerOpDefinitionCluster cross-module pair, and the policy-engine kernel mirror. Cluster-name outputs must stay byte-identical.

mehara-rothila:
@coderabbitai we're renaming the `OperationUpstream` schema in `gateway/gateway-controller/api/management-openapi.yaml` (the per-operation `{main, sandbox}` upstream override, each sub-field a ref to a named `upstreamDefinitions` entry). The same schema exists identically in the platform-api OpenAPI, so the chosen name must work on both sides.

Our candidates:
1. `OperationUpstream` (current)
2. `OperationUpstreamOverride`
3. `PerOperationUpstream`
4. `OperationUpstreamRef`
5. `OperationLevelUpstream`

Please (a) add any stronger options we've missed, then (b) rank ALL of them (ours + yours) best-to-worst with a one-line rationale each. Judge by: clarity of intent (it overrides the API-level upstream for a single operation), consistency with the sibling schemas (`Upstream`, `UpstreamDefinition`, `UpstreamReference`, `UpstreamTimeout`), and that it works as a schema name on both the gateway and platform-api sides.

mehara-rothila:
@coderabbitai please do an equivalence audit of this PR's recent rework.

Compare commit dac1f564a (the earlier verified state of this branch) against the current head 38958c316 and confirm the observable behavior is unchanged: for the same deployed API, identical routing results (which backend each operation reaches on main and sandbox vhosts, per-op ref reuse of definition clusters, fallbacks), identical path rewriting (definition basePath handling), identical timeout propagation, identical hostRewrite behavior, identical precedence (op-policy > api-policy > per-op ref > api-level upstream), and an identical validation accept/reject set.

The following changes in that range are INTENTIONAL, please verify they are implemented consistently but do not report them as regressions:
1. API-level cluster names moved from a truncated 24-hex hash to the full sha256(apiID) hex via clusterkey.Hash/HashedName; the hash input is still only the API id (URL-stable), main/sandbox share the fragment, and both xDS builders use the shared helper. Definition cluster names (clusterkey.DefinitionName) are unchanged.
2. OpenAPI schema renames: RestAPIOperationUpstream / RestAPIOperationUpstreamTarget consolidated into OperationUpstream with inline main/sandbox; the wire payload (operations[].upstream.main.ref etc.) and constraints (required ref, additionalProperties false, minProperties 1) are unchanged.
3. Validator refactors: the name contract (max 100, ^[a-zA-Z0-9\-_]+$) centralized into validateUpstreamRefName; error message labels now derive from the field path (wording changed, outcomes did not).
4. Sandbox detection consolidated into upstreamref.HasContent/SandboxActive (the xDS translator moved from nil-check to content-check, aligning with the transformer; unreachable for validator-passing inputs).
5. The latest commit consolidates the per-op override glue into routeSlot (transform) and routeParams (xDS) with a shared applyPerOpRef, changes the per-op ref parameters from *api.UpstreamReference to plain string (UpstreamReference is a string alias), drops a redundant Timeout nil-guard (the helper already returns nil for nil), and removes a duplicated nil-gate in the validator.

Please verify with file:line evidence and flag ONLY things outside this list: any route/cluster/policy-chain output that changed, any validation outcome that flipped, any error path lost, or any type/nil-safety issue introduced by the string-parameter change. If nothing is out of place, say explicitly that the range is behavior-preserving apart from the listed intentional changes.

mehara-rothila:
@coderabbitai review

mehara-rothila:
@coderabbitai full review

gateway/gateway-controller/pkg/utils/upstreamref/upstreamref.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+	d, err := time.ParseDuration(trimmed)
+	if err != nil {
+		return nil, fmt.Errorf("invalid timeout format: %w", err)
+	}
+	if d <= 0 {
+		return nil, fmt.Errorf("timeout must be positive, got: %v", d)
+	}
+	return &d, nil
+}
```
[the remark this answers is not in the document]
mehara-rothila: Not changed in this PR. The `<= 0` rejection matches the pre-existing `parseTimeout` contract on main; the helper centralizes that logic, it does not change it. Accepting `"0s"` first needs a decision on what zero maps to for a cluster connect timeout, since Envoy requires a positive value (map to the gateway default, or reject at validation instead). That contract decision is better made separately from this feature.

O-sura (dismissed):

mehara-rothila:
@coderabbitai review

mehara-rothila:
@coderabbitai review

tharindu1st (approved):

PR #2030, opened 2026-05-27
feat(auth): enhance JWT authentication with IDP support and Scope Validation integration

Thushani-Jayasekera:
## Purpose
$subject

This PR introduces a three-mode authentication architecture for the Platform API and integrates a unified RBAC layer across all modes.

Authentication mode selection is controlled via:
```
IDP_ENABLED
IDP_TYPE

```
These are evaluated at startup to determine the active authentication strategy.

#### Mode 1 — Simple JWT (default)

**Config**:

`IDP_ENABLED=false`

**Behavior**
Uses ThunderAuthMiddleware (renamed from the previous AuthMiddleware)
Validates JWT locally without external IDP calls

**Token validation**
`JWT_SKIP_VALIDATION=true` (default) - Signature validation is skipped. Intended for local development / test environments
`JWT_SKIP_VALIDATION=false` - HMAC signature validation using JWT_SECRET_KEY

**Notes**
Fully self-contained authentication flow
No runtime dependency on external identity provider

#### Mode 2 — Thunder IDP

**Config**:

```
IDP_ENABLED=true
IDP_TYPE=thunder
```

**Behavior**

JWT validated using Thunder JWKS endpoint: `THUNDER_JWKS_URL`
Uses shared authenticators.AuthMiddleware for JWKS-based validation
TLS verification for JWKS is intentionally skipped for local/self-signed setups

**Claim handling**
PlatformClaimsMiddleware maps token claims into Gin context:
- user_id
- username
- email
- organization
- scope
- platform_roles

**Authorization model**

Scope-based access control

Example scopes:

```
api-platform:gateway:create
api-platform:gateway:delete
```

#### Mode 3 — External IDP

**Config**:

```
IDP_ENABLED=true
IDP_TYPE=external
```

**Behavior**

JWT validated using external JWKS endpoint:
```
EXTERNAL_IDP_JWKS_URL Claim mapping (fully configurable)
EXTERNAL_IDP_ORGANIZATION_CLAIM_NAME
EXTERNAL_IDP_USER_ID_CLAIM_NAME
EXTERNAL_IDP_USERNAME_CLAIM_NAME
EXTERNAL_IDP_EMAIL_CLAIM_NAME
EXTERNAL_IDP_SCOPE_CLAIM_NAME
EXTERNAL_IDP_ROLES_CLAIM_PATH (supports nested paths like realm_access.roles)
```

**Role mapping**
External roles mapped to platform roles using:

`EXTERNAL_IDP_ROLE_MAPPINGS`

**Format**:

`idp-role=platform-role`

**Authorization model**
- Role-based access control (primary)
- Token roles (e.g., admin, developer) are mapped to platform permissions
- Scope-based checks used as fallback if roles are missing
- Supports multiple issuers and audience prefix matching

**RBAC Model**
```
| Platform Roles    | Role Access |
| admin |	Full access to all resources | 
| developer | Full CRUD + deploy access, excluding gateway management, dev portal admin, and subscription plan admin | 
| viewer	| Read-only access to all resources | 
```

**Permission Format**

All permissions follow:

`api-platform:<resource>:<action>`

Example:

api-platform:gateway:create
api-platform:subscription:delete

These are directly compatible with OAuth2 scopes in Thunder mode.

**Enforcement**
`RBAC_ENABLED=true (default) -` Enforces RequirePermission / RequireAnyPermission middleware across protected routes
`RBAC_ENABLED=false`- Skips authorization checks (useful for debugging / early deployment)

**Mode behavior**
Thunder mode: scope-based enforcement
External IDP mode: role-based enforcement with scope fallback

**Environment Variables**
```
| Variable	|  Default	Description | 
| IDP_ENABLED | 	false | 	Enables IDP-backed authentication
| IDP_TYPE	|  ""	|  thunder or external
| RBAC_ENABLED	|  true | 	Enables permission enforcement
| THUNDER_BASE_URL	|  http://localhost:8090	|  Thunder service endpoint
| THUNDER_CLIENT_ID / THUNDER_CLIENT_SECRET	|  "" | 	Service-to-service credentials
| THUNDER_JWKS_URL	|  ""	|  JWKS endpoint for Thunder
| THUNDER_ISSUER|  	thunder	|  Expected JWT issuer
| THUNDER_ORGANIZATION_CLAIM_NAME | 	organization	|  Org claim key
| EXTERNAL_IDP_JWKS_URL | 	""	|  External JWKS endpoint
| EXTERNAL_IDP_ISSUER | 	""	|  Allowed issuers (comma-separated)
| EXTERNAL_IDP_AUDIENCE	 |  "" |  	Allowed audiences (supports prefix match with *)
| EXTERNAL_IDP_ORGANIZATION_CLAIM_NAME | 	organization	|  Org claim key
| EXTERNAL_IDP_ROLES_CLAIM_PATH | 	roles	|  Path to roles claim
| EXTERNAL_IDP_ROLE_MAPPINGS | 	""	|  Role mapping configuration
```

**Important**
AuthMiddleware has been renamed to ThunderAuthMiddleware
Only internal server wiring (server.go) updated
RequireScope middleware has been removed
Replaced with:
RequirePermission
RequireAnyPermission

These now support both:

scope-based checks (Thunder mode)
role-based resolution (external IDP mode)

# 1. Authentication

  ## Mode A — Simple JWT (IDP_ENABLED=false, default) — ThunderAuthMiddleware → validateThunderToken

  **Validate JWT (parse structure)**
  Always done. Uses jwt.NewParser(jwt.WithoutClaimsValidation()).ParseUnverified when SkipValidation=true, or jwt.ParseWithClaims otherwise. A structurally malformed token is rejected in both cases.
  
  **Verify signature**
  - SkipValidation=true (default): skipped entirely — ParseUnverified is called, the signature bytes are not checked at all. Intended for local dev.
  - SkipValidation=false: HMAC signature verified using JWT_SECRET_KEY. The signing method is also asserted to be
  *jwt.SigningMethodHMAC; any other algorithm is rejected. (auth.go:117–122)
  
  **Verify issuer**
  Only when SkipValidation=false and TokenIssuer is non-empty — the iss claim is compared against the configured value. (auth.go:131–136)
  When SkipValidation=true, issuer is not checked.

  **Verify expiry**
  jwt.ParseWithClaims handles this automatically via the golang-jwt library (checks exp claim). But only runs when
  SkipValidation=false. When SkipValidation=true, jwt.WithoutClaimsValidation() is passed, so expiry is not checked.
  
  ---
 ##  Mode B — Thunder IDP (IDP_ENABLED=true, IDP_TYPE=thunder) — common/authenticators.JWTAuthenticator

  **Validate JWT** 
  jwt.ParseWithClaims with j.jwks.Keyfunc — keys are fetched from THUNDER_JWKS_URL on startup and refreshed every 10 minutes (jwt_authenticator.go:128)
  
  **Verify signature**
  Always done — the JWKS keyfunc supplies the public key for RSA/EC signature verification. Note: TLS verification of the JWKS endpoint itself is disabled (self-signed cert support), but token signature verification is unaffected.(jwt_authenticator.go:75–78)  

  **Verify issuer**
  Checked against THUNDER_ISSUER using claims.GetIssuer(). Fails with an error if they don't match.(jwt_authenticator.go:143–149)

  **Verify expiry**
  jwt.ParseWithClaims checks exp automatically (no WithoutClaimsValidation). A comment notes leeway can be added for clock drift but is not currently applied. (jwt_authenticator.go:127)
  
  ---
  ## Mode C — External IDP (IDP_ENABLED=true, IDP_TYPE=external) — same JWTAuthenticator
  
  Identical flow to Thunder IDP — same Authenticate method, just configured with EXTERNAL_IDP_JWKS_URL and EXTERNAL_IDP_ISSUER. Additionally validates audience if EXTERNAL_IDP_AUDIENCE is set. (jwt_authenticator.go:152–161)
  
  ---
 #  2. Authorization
  
  Entry point — checkPermission (authorization.go:78)

  Every protected route goes through RequirePermission(perm) → checkPermission. The logic is:

  1. If RBAC_ENABLED=false → allow
  2. If AuthzSkipKey is set in context → allow   (set when ScopeClaim is not configured in IDP mode)
  3. Scope check: does the space-separated scope string contain perm.Scope()?
  4. Role fallback: do any platform_roles in context grant perm via RolePermissions map?

  Steps 3 and 4 are the same code path for all three modes — the difference is what ends up in the scope and platform_roles context keys.

  ---
  Scopes
  
  What is a scope? A Permission constant (e.g. GatewayCreate) has a .Scope() method that prefixes it:
  "api-platform:gateway:create". (permissions.go:28)
  
  Thunder IDP / Simple JWT: Thunder issues these fine-grained scopes directly in the scope claim. hasScope splits the claim string and does an exact string match. This is the primary authorization path.

 External IDP: The IDP likely does not issue api-platform:* scopes. Scope check runs first but typically misses, and the role fallback carries the decision.

  ---
  Roles
  
  Roles (admin, developer, viewer) are the three canonical platform roles defined in predefined_roles.go.

  Simple JWT mode: platform_roles is set to strings.Fields(scope) — so the scope string itself is treated as roles. This only works as a role check if Thunder puts role names (not fine-grained scopes) in the scope claim. In practice, when Thunder issues api-platform:* scopes, these won't match any role name, so HasPermissionForRoles would return false and the scope check in step 3 carries the decision instead.

  Thunder IDP mode: PlatformClaimsMiddleware is wired without RolesClaimPath (auth.go:239 called with empty path) → platform_roles
   is empty. Scope check in step 3 is expected to handle everything.

  External IDP mode: PlatformClaimsMiddleware is wired with EXTERNAL_IDP_ROLES_CLAIM_PATH (e.g. realm_access.roles) and
  EXTERNAL_IDP_ROLE_MAPPINGS. IDP role values are mapped to admin/developer/viewer and stored in platform_roles.
  HasPermissionForRoles then looks up the permission set for each role in RolePermissions map.

  ---
  Permissions (the RBAC model)

  RolePermissions statically maps each role to a slice of Permission constants:
  - admin → all permissions
  - developer → all resource CRUD/deploy, no gateway manage / devportal admin / subscription plan admin
  - viewer → all *:read permissions
  
  HasPermissionForRoles iterates the roles, looks up their permission sets, and checks if perm is present. There is no dynamic, database-driven permission assignment — the mapping is fully compile-time.
  
  ---
  Gaps worth noting
  
```
  ┌───────────────────────┬───────────────────────────────────────────────────────────────────────────────────────────────────┐
  │         Area          │                                               Issue                                               │
  ├───────────────────────┼───────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ Expiry in simple JWT  │ When SkipValidation=true (the default), expiry is not checked — any non-expired-looking token     │
  │ mode                  │ passes.                                                                                           │
  ├───────────────────────┼───────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ Issuer in simple JWT  │ Also skipped when SkipValidation=true.                                                            │
  │ mode                  │                                                                                                   │
  ├───────────────────────┼───────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ Signing method in     │ Hardcoded to HMAC; Thunder IDP actually issues RSA-signed tokens, so SkipValidation=false in      │
  │ simple JWT mode       │ simple mode would fail against a real Thunder token.                                              │
  ├───────────────────────┼───────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ platform_roles in     │ Set to strings.Fields(scope) — if Thunder issues api-platform:gateway:create style scopes these   │
  │ simple JWT mode       │ won't match any role, so role-based fallback silently contributes nothing (which is correct, but  │
  │                       │ potentially confusing).                                                                           │
  └───────────────────────┴
```


platform-api/src/internal/rbac/predefined_roles.go
```diff
@@ -0,0 +1,103 @@
...
+// PermissionsForRoles returns the deduplicated set of permissions granted to the given roles.
+func PermissionsForRoles(roles []string) map[Permission]struct{} {
+	result := make(map[Permission]struct{})
+	for _, role := range roles {
+		for _, perm := range RolePermissions[role] {
+			result[perm] = struct{}{}
+		}
+	}
+	return result
+}
```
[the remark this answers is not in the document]
Thushani-Jayasekera: resolve.go has implementation

platform-api/src/internal/handler/api.go
```diff
@@ -992,27 +993,27 @@ func (h *APIHandler) RegisterRoutes(r *gin.Engine) {
...
 	apiGroup := r.Group("/api/v1/rest-apis")
 	{
-		apiGroup.POST("", h.CreateAPI)
+		apiGroup.POST("", middleware.RequirePermission(rbac.APICreate), h.CreateAPI)
 		apiGroup.GET("", h.ListAPIs)
 		apiGroup.GET("/:apiId", h.GetAPI)
-		apiGroup.PUT("/:apiId", h.UpdateAPI)
-		apiGroup.DELETE("/:apiId", h.DeleteAPI)
+		apiGroup.PUT("/:apiId", middleware.RequirePermission(rbac.APIUpdate), h.UpdateAPI)
+		apiGroup.DELETE("/:apiId", middleware.RequirePermission(rbac.APIDelete), h.DeleteAPI)
```
Thushani-Jayasekera: middleware to parse openapi and validate scopes from that.

platform-api/src/resources/openapi.yaml
```diff
@@ -262,6 +306,7 @@ paths:
         Deletes a specific project by its UUID. Access is validated against the organization 
         in the JWT token.
       operationId: DeleteProject
+      x-required-roles: [admin, developer]
```
Thushani-Jayasekera: define scopes

PR #2046, opened 2026-05-29
Update Devportal API Spec

ShakyaPr:
### Purpose

- Restructures the Devportal API Spec `devportal-openapi-spec-v1.yaml` with below things:
  - New URL shape for org-scoped resources: `/o/{orgId}/devportal/v1/{resource}` (was `/devportal/organizations/{orgId}/{resource}`)
  - Naming normalization — kebab-case resources, consistent camelCase path params, Api casing in operationIds.
  - **Subscription-policies lookup redesign** — name lookup is now a collection filter, not an overloaded path segment.

### URL shape

- `servers.url` is now `https://devportal.api-platform.io` (prod) plus the local HTTP/HTTPS variants. The `/devportal` base path is gone from `servers.url` and reintroduced after `{orgId}` in each path.
- All org-scoped paths rewritten: `/organizations/{orgId}/X` → `/o/{orgId}/devportal/v1/X` (~50 paths).
- Org-collection meta endpoints (`POST /organizations`, `GET/PATCH/DELETE /organizations/{orgId}`) kept at the root — they manage the org collection itself, not resources within an org.
- Non-org paths (`/login`, `/temp-arazzo-file`, `/applications/*`) untouched; these will be dropped from the spec in a follow-up when the legacy-only routes are removed.

### Naming changes

| Before | After |
|---|---|
| `/identityProvider` (singular, camelCase) | `/identity-providers` (plural, kebab) |
| `/events`, `/events/{eventId}` | `/webhook-events`, `/webhook-events/{eventId}` |
| `/deliveries/{deliveryId}/retry` | `/webhook-deliveries/{deliveryId}/retry` |
| `{subscriptionId}` (1 path, 1 param ref) | `{subId}` — consolidated on existing `SubId` component |
| `{name}` (view name, 3 paths) | `{viewName}` — consolidated on existing `ViewName` component |
| `{policyIdOrName}` (overloaded) | `{policyId}` (see subscription policies redesign) |

Component cleanup: removed unused `Name` and duplicate `SubscriptionId` parameter components; renamed `PolicyIdOrName` → `PolicyId`.

OperationIds normalized so `API` is rendered as `Api` (14 IDs across API metadata, API content, API flows): e.g. `createAPIMetadata` → `createApiMetadata`, `getAPIFlow` → `getApiFlow`. The PUT API-content endpoint is renamed from `updateAPIContent` to `replaceApiContent` to better reflect its replace-semantics. `OAuth*` operationIds left as-is (proper noun).

### Subscription policies — lookup-by-name redesign

- The old {policyIdOrName} path overloaded one segment to mean either UUID or name (with GET/DELETE interpreting it differently). Replaced with the standard REST split:

  - GET `/subscription-policies` — new listSubscriptionPolicies operation with optional ?name= filter. Returns an array (0 or 1 item when filtered). New SubscriptionPolicyListResponse component added.
  - GET `/subscription-policies/{policyId}` — ID-only, single-resource fetch. 
  - DELETE `/subscription-policies/{policyId}` — ID-only, single-resource delete.
Callers that only have a name now do list?name=X then operate on the returned policyId. New endpoint will need a matching controller method when the spec-driven router is enabled.

### DIFF
--- a/portals/developer-portal/docs/devportal-openapi-spec-v1.yaml
+++ b/portals/developer-portal/docs/devportal-openapi-spec-v1.yaml
@@ -24,13 +24,15 @@ info:
     Fine-grained Developer Portal API for managing organizations, identity providers, providers,
     API metadata and content, applications, subscriptions, application credentials, billing, invoices, usage, and API flows.
 
-    All paths are served under the `/devportal` base path. Operations declare the least-privilege
-    OAuth2 scopes required for each resource action, and selected billing-related endpoints also
-    support API key authentication where configured.
+    Organization-scoped resources are served under `/o/{orgId}/devportal/v1`. Operations declare
+    the least-privilege OAuth2 scopes required for each resource action, and selected billing-related
+    endpoints also support API key authentication where configured.
 servers:
-  - url: http://localhost:3000/devportal
+  - url: https://devportal.api-platform.io
+    description: Production
+  - url: http://localhost:3000
     description: Local development server
-  - url: https://localhost:{port}/devportal
+  - url: https://localhost:{port}
     description: Local HTTPS server
     variables:
       port:
@@ -171,7 +173,7 @@ paths:
         - OAuth2Security:
             - dp:org_delete
             - dp:org_manage
-  /organizations/{orgId}/identityProvider:
+  /o/{orgId}/devportal/v1/identity-providers:
     parameters:
       - $ref: "#/components/parameters/OrgId"
     post:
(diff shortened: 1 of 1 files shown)


portals/developer-portal/docs/devportal-openapi-spec-v1.yaml
```diff
@@ -375,7 +377,7 @@ paths:
           $ref: "#/components/responses/BadRequest"
         "404":
           $ref: "#/components/responses/TextMessageResponse"
-  /organizations/{orgId}/provider:
+  /o/{orgId}/devportal/v1/provider:
```
Piumal1999, line 380:
```suggestion
  /o/{orgId}/devportal/v1/providers:
```

PR #2058, opened 2026-05-30
fix(controller): apply most-specific match for overlapping LLM policies

### LINKED ISSUE #2057: Overlapping LLM policy paths/methods stack instead of most-specific
### Description

When the same policy is attached to multiple overlapping `paths` (and/or `methods`) on an `LlmProvider` or `LlmProxy`, the gateway applies **all** matching policy instances to a request instead of only the most specific match. The most restrictive instance then dominates, so the route is governed by the wrong policy.

**Example** — `advanced-ratelimit` attached to an `LlmProvider` on two paths:

```yaml
policies:
  - name: advanced-ratelimit
    version: v1
    paths:
      - path: /chat/completions
        methods: [POST]
        params: { quotas: [{ name: request-limit, limits: [{ duration: 1h, limit: 4 }] }] }
      - path: /*
        methods: ['*']
        params: { quotas: [{ name: request-limit, limits: [{ duration: 1h, limit: 1 }] }] }
```

**Expected:** `POST /chat/completions` is governed only by its own quota (4/hour); every other path is governed by `/*` (1/hour).

**Actual:** the `/chat/completions` route's policy chain contains **both** `advanced-ratelimit` instances (visible in the policy-engine config dump). The `/*` quota (1/hour) is also enforced on `/chat/completions`, so the 2nd request returns `429` — the route is wrongly limited to 1/hour instead of 4/hour.

**Steps to reproduce:**
1. Create an `LlmProvider` with the policy configuration above.
2. Send repeated `POST` requests to `/<context>/chat/completions`.
3. The 2nd request returns `429` (`X-RateLimit-Quota: wildcard-quota`) instead of being allowed up to 4/hour.

**Root cause:** the controller's LLM transformer (`gateway/gateway-controller/pkg/utils/llm_transformer.go`) fans a wildcard-path policy out onto every more-specific operation, so the specific route's chain ends up with multiple same-named instances, and the policy engine executes the whole chain.

**Expected resolution — most specific match wins, per policy name:**
- **Path** first: concrete > wildcard prefix, longer > shorter (`/chat/completions` > `/chat/*` > `/*`).
- then **Method**: the narrower HTTP method set wins (`[POST]` is more specific than `[GET, POST]`, which is more specific than `*`).
- Different policy **names** still layer (e.g. a global `set-headers` on `/*` plus a path-specific rate limit).

**Scope:** affects `LlmProvider` and `LlmProxy` (both use the LLM transformer). `RestApi` is not affected — its operations carry explicit per-operation policies with no wildcard fan-out.

### Version

_No response_

### Related Issue

_No response_

renuka-fernando:
## Purpose

Resolves #2057.

On an `LlmProvider` or `LlmProxy`, the same policy can be attached to several `paths`/`methods`. The LLM transformer fanned every matching attachment onto each operation, so a route's policy chain stacked multiple instances and the most restrictive one silently dominated — for example `POST /chat/completions` was limited by a `/*` quota (1/h) instead of its own (4/h).

## Goals

Resolve overlapping policy attachments to the most specific match per route, so each path+method is governed only by its most specific rule. Within a single policy block this collapses to one effective policy per route; separate policy blocks (even with the same name) still each contribute and layer.

## Approach

- In `gateway/gateway-controller/pkg/utils/llm_transformer.go`, before layering a policy onto a target path, skip it when a strictly more specific attachment in the **same policy block** also covers that path+method (`moreSpecificPolicyAttachmentCovers`).
- Specificity ordering (`isMoreSpecificAttachment`): **path first** — concrete beats wildcard, longer prefix beats shorter (`isMoreSpecificPath`); then **method** — the narrower HTTP method set wins, so `[POST]` is more specific than `[GET, POST]`, which is more specific than `*` (`isStrictMethodSubset`).
- Resolution is scoped to a single block, so separate `spec.policies[]` entries — even ones with the same name, such as two `advanced-ratelimit` blocks — each resolve their own most-specific match and all layer onto the route.
- Wired into all three transform paths — `transformProvider` (`allow_all` and `deny_all`) and `transformProxy` — and the result is independent of declaration order.
- Scope: `LlmProvider` and `LlmProxy` only. `RestApi` is unaffected because its operations carry explicit per-operation policies with no wildcard fan-out.

## User stories

As an API developer, when I attach a policy to a specific path/method and a broader fallback (such as `/*`), I expect the specific rule to govern that path and the fallback to govern everything else, rather than both stacking on the specific route. When I declare two separate policy blocks of the same type, I expect both to apply, each resolved to its own most specific match.

## Documentation

N/A. This corrects internal policy-resolution behaviour; the `LlmProvider`/`LlmProxy` `spec.policies[].paths` format is unchanged.

## Automation tests

- Unit tests
  > Added in `pkg/utils/llm_transformer_test.go`: `MostSpecificPathWinsForSamePolicy`, `MostSpecificPathWinsAcrossNestedWildcards`, `MostSpecificPathWinsAcrossMethodsAndOrder`, `MostSpecificMethodWinsOnSamePath`, `NarrowerMethodSetWins`, `PathDominatesMethodSpecificity`, `DifferentPoliciesResolveIndependently`, `DenyAllModeMostSpecificWins`, `TemplateExpansionDedupesSameNamePolicy`, and `SeparatePoliciesOfSameNameBothApply`. The existing `PolicyOrderDoesNotAffectWildcardCoverage` and `ExpandsWildcardPolicyPathWithTemplateMappings` still pass. `go test ./pkg/utils/...` is green.
- Integration tests
  > Added `gateway/it/features/llm-policy-path-specificity.feature` with 7 scenarios validated end-to-end via response headers and 429 enforcement: specific-vs-wildcard, nested paths (4/2/1), per-method on the same path, concrete-method-beats-`*`, narrower-method-set, an `LlmProxy` scenario, and two same-name `set-headers` blocks both applying. Full run (with `health` + `api_deploy` warm-up): 16 scenarios / 329 steps passing.

## Security checks

- Ran FindSecurityBugs plugin and verified report? **N/A** — FindSecurityBugs is a Java/SpotBugs plugin and this change is Go.

## Samples

All samples attach policies to an `LlmProvider` via `spec.policies[].paths` (each entry is `path` + `methods` + a per-hour `limit`). `=>` shows the resolved per-route limit after this fix; before the fix every overlapping route collapsed to the most restrictive limit.

Sample 1 — specific path beats the wildcard fallback:
```
/chat/completions  [POST] -> 4/h
/*                 [*]    -> 1/h
=> POST /chat/completions: 4/h ; every other path: 1/h
```

Sample 2 — nested wildcards (not just `/*`):
```
/chat/completions  [*] -> 4/h
/chat/*            [*] -> 2/h
/*                 [*] -> 1/h
=> /chat/completions: 4/h ; /chat/<other>: 2/h ; everything else: 1/h
```

Sample 3 — different methods on the same path, plus wildcards:
```
/chat/completions  [POST] -> 4/h
/chat/completions  [GET]  -> 10/h
/chat/*            [*]     -> 2/h
/*                 [*]     -> 1/h
=> POST /chat/completions: 4/h ; GET /chat/completions: 10/h ; /chat/<other>: 2/h ; else: 1/h
```

Sample 4 — a concrete method beats the `*` method on the same path:
```
/chat/completions  [*]   -> 4/h
/chat/completions  [GET] -> 10/h
=> GET /chat/completions: 10/h ; every other method on /chat/completions: 4/h
```

Sample 5 — a narrower method set beats a broader one on the same path:
```
/chat/completions  [GET, POST] -> 3/h
/chat/completions  [POST]      -> 100/h
=> GET /chat/completions: 3/h ; POST /chat/completions: 100/h
```

Sample 6 — same resolution applies to an `LlmProxy`'s own policies:
```
proxy policies:  /chat/completions [*] -> 4/h ;  /* [*] -> 1/h
=> proxy POST /chat/completions: 4/h ; every other proxy path: 1/h
```

Sample 7 — two separate blocks of the same policy name both apply, each most-specific within itself:
```
block 1 (advanced-ratelimit):  /chat/completions [*] -> 4/h , [GET] -> 10/h , /chat/* [*] -> 2/h , /* [*] -> 1/h
block 2 (advanced-ratelimit):  /* [*] -> 100/day
=> POST /chat/completions: 4/h AND 100/day ; GET /chat/completions: 10/h AND 100/day ; /chat/<other>: 2/h AND 100/day ; else: 1/h AND 100/day
```

Declaration order of the `paths` entries does not affect any of the above. Across separate blocks, every block's most-specific match is applied.

## Related PRs

N/A

## Test environment

- Go 1.23+ (gateway-controller module); `go test ./pkg/utils/...`.
- Gateway integration suite (Godog + Docker Compose: gateway-controller, gateway-runtime with Envoy + policy-engine, echo/mock backends) using coverage-instrumented images.
- Docker via Colima on macOS (darwin/arm64).

### DIFF
--- a/gateway/gateway-controller/pkg/utils/llm_transformer.go
+++ b/gateway/gateway-controller/pkg/utils/llm_transformer.go
@@ -246,7 +246,7 @@ func (t *LLMProviderTransformer) transformProxy(proxy *api.LLMProxyConfiguration
 							if attachedPolicyPaths[targetPath] {
 								continue
 							}
-							if moreSpecificPolicyAttachmentCovers(targetPath, policyMethod, attachment, *proxy.Spec.Policies) {
+							if moreSpecificPolicyAttachmentCovers(targetPath, policyMethod, attachment) {
 								continue
 							}
 							targetKey := pathMethodKey{path: targetPath, method: policyMethod}
@@ -467,7 +467,7 @@ func (t *LLMProviderTransformer) transformProvider(provider *api.LLMProviderConf
 								if attachedPolicyPaths[targetPath] {
 									continue
 								}
-								if moreSpecificPolicyAttachmentCovers(targetPath, policyMethod, attachment, *provider.Spec.Policies) {
+								if moreSpecificPolicyAttachmentCovers(targetPath, policyMethod, attachment) {
 									continue
 								}
 								targetKey := pathMethodKey{path: targetPath, method: policyMethod}
@@ -557,7 +557,7 @@ func (t *LLMProviderTransformer) transformProvider(provider *api.LLMProviderConf
 								if attachedPolicyPaths[targetPath] {
 									continue
 								}
-								if moreSpecificPolicyAttachmentCovers(targetPath, policyMethod, attachment, *provider.Spec.Policies) {
+								if moreSpecificPolicyAttachmentCovers(targetPath, policyMethod, attachment) {
 									continue
 								}
 								targetKey := pathMethodKey{path: targetPath, method: policyMethod}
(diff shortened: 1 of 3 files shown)


renuka-fernando:
@coderabbitai full review

Krishanx92 (approved):

PR #2066, opened 2026-05-31
fix(operator): honor HTTPRoute path.type for PathPrefix

### LINKED ISSUE #2021: Operator HTTPRoute ignores path.type — PathPrefix degrades to Exact
### Description

The Gateway Operator's HTTPRoute → RestApi mapper reads only `match.path.value` and ignores `match.path.type`. As a result, every HTTPRoute path is effectively treated as `Exact`, even when the user explicitly sets `type: PathPrefix`.

**Where it happens**

`kubernetes/gateway-operator/internal/controller/httproute_mapper.go:173-192` — the loop over `rule.Matches` reads `m.Path.Value` but never inspects `m.Path.Type`. The verbatim value is written into `Operation.Path`.

The gateway-controller treats an `Operation.Path` as a prefix route only when it ends in `/*` (`gateway/gateway-controller/pkg/xds/translator.go:224, 1749-1750`). Since the operator never appends `/*` for `PathPrefix` matches, prefix semantics are lost.

**Reproduction**

Apply an HTTPRoute with:

```yaml
matches:
  - path:
      type: PathPrefix
      value: /foo
```

- Request `GET /foo` → 200 (matches)
- Request `GET /foo/bar` → 404 (does NOT match)

**Expected**

Per the Gateway API spec, `PathPrefix` should match `/foo` **and** any path under `/foo/`. The operator should emit an `Operation.Path` of `/foo/*` (or equivalent) when `path.type == PathPrefix`.

**Workarounds**

- Enumerate every concrete operation path explicitly on the HTTPRoute.
- Use the `RestApi` CR directly, where `path: /foo/*` already works.
- `value: /foo/*` on an HTTPRoute does **not** work — Gateway API CRD validation rejects `*` in `Exact` / `PathPrefix` values.

**Proposed fix**

In `httproute_mapper.go`, when `m.Path.Type` is `PathMatchPathPrefix` and the value is not `/`, append `/*` to `pathVal` before emitting the `Operation`. `Exact` keeps current behavior. `RegularExpression` should be rejected with a clear `Invalid` config error (or mapped explicitly if we want to support it).

### Version

_No response_

### Related Issue

_No response_

renuka-fernando:
## Purpose

The Gateway Operator's HTTPRoute → RestApi mapper read only `match.path.value` and ignored `match.path.type`, so every HTTPRoute path was treated as `Exact` even when the user set `type: PathPrefix`.
As a result, a `PathPrefix` of `/foo` matched `GET /foo` but returned 404 for `GET /foo/bar`, breaking prefix-based routing (required when API Platform runs as the edge gateway in OpenChoreo, which uses prefix-based routing in its default component types).
Resolves #2021.

## Goals

Honor `match.path.type` so `PathPrefix` produces a real prefix route, keep `Exact` behavior unchanged, and reject unsupported `RegularExpression` matches with a clear error.

## Approach

- Add `restAPIPathForHTTPRouteMatch`, which maps a Gateway API path match to an `Operation.Path` based on `path.type` instead of reading the value alone.
- `PathPrefix /foo` now emits `/foo/*`, the gateway-controller's wildcard-prefix convention (the xDS translator routes a `/*`-suffixed path as a prefix).
- `PathPrefix /` stays `/` (root catch-all, never `//*`); `Exact` is emitted verbatim (unchanged behavior); an unset `path.type` defaults to `PathPrefix` per the Gateway API spec.
- `RegularExpression` is rejected as an `Invalid` config error, because the gateway-controller has no raw-regex path support (it supports only exact paths, the `/*` wildcard, and `{param}` templating).
- The rule-matches loop now delegates to the helper rather than reading `path.value` directly.

## User stories

As a platform user deploying APIs through Gateway API `HTTPRoute`s, I want a `PathPrefix` match to route the prefix and everything beneath it, so that I do not have to enumerate every sub-path or fall back to the `RestApi` CR.

## Documentation

N/A — this is a bug fix that aligns operator behavior with the Gateway API spec and the existing `api.yaml` path semantics (`concepts/api-yaml-specification.md`); no new user-facing configuration is introduced.

## Automation tests

- Unit tests
  - Updated three assertions in `httproute_mapper_test.go` that encoded the old `Exact`-degraded behavior (`PathPrefix` now expects the `/*` suffix).
  - Added `TestBuildAPIConfigFromHTTPRoute_PathMatchType` covering `PathPrefix → /*`, nested prefix, trailing slash, `PathPrefix /` → `/`, `Exact` verbatim, unset-type default, and `RegularExpression` rejection.
  - `go test ./internal/controller/ -run TestBuildAPIConfigFromHTTPRoute` passes.
- Integration tests
  - Validated end-to-end on a local Kubernetes cluster (Colima) with the Gateway Operator running locally and reaching the gateway-controller via kubefwd.
  - The operator mapped a `PathPrefix /hello` HTTPRoute and pushed `operations[].path == /hello/*` to the gateway-controller.
  - `GET /hello-context/hello`, `GET /hello-context/hello/world`, and `GET /hello-context/hello/a/b/c` all returned 200 (before the fix the sub-paths returned 404).

## Security checks

- Ran FindSecurityBugs plugin and verified report? N/A — FindSecurityBugs is a Java/SpotBugs plugin and this change is Go code.

## Samples

The change is exercised by the existing Gateway API demo HTTPRoute (`kubernetes/helm/resources/gateway-api-operator-demo/04-httproute.yaml`), which uses a `PathPrefix` match:

```yaml
metadata:
  annotations:
    gateway.api-platform.wso2.com/context: "/hello-context"
spec:
  rules:
    - matches:
        - path:
            type: PathPrefix
            value: /hello
          method: GET
```

With the operator running and the route deployed, the gateway-controller receives `operations: [{ method: GET, path: /hello/* }]`, and all of the following return 200:

```bash
curl -k https://<gateway-runtime>:8443/hello-context/hello         # 200
curl -k https://<gateway-runtime>:8443/hello-context/hello/world   # 200 (was 404 before the fix)
curl -k https://<gateway-runtime>:8443/hello-context/hello/a/b/c   # 200
```

## Related PRs

N/A

## Test environment

- Go 1.26.3 (darwin/arm64)
- macOS; local Kubernetes via Colima
- Gateway Helm chart 1.1.1 (appVersion 1.1.0); gateway-controller and gateway-runtime images 1.1.0
- Gateway API standard CRDs v1.3.0
- JDK: N/A (Go component)


renuka-fernando:
The operator IT failure here (`K8s Gateway API HTTPRoute update test` → `PUT /k8s/put` returns **405**) is **not** an operator-side bug. It's a gateway-controller xDS translator issue: for a wildcard operation path `/foo/*`, the controller stripped the matched `/foo` prefix from the upstream path, so `PUT /k8s/put` reached the backend as `PUT /` (which httpbin rejects with 405). The `GET /k8s/get` smoke test only passed by luck, since httpbin's `/` happens to answer GET.

Tracked separately as #2071 and fixed in #2072 (preserves the matched prefix on the upstream; the bare `/*` catch-all and `/` root are unchanged).

Once #2072 merges, this branch should be rebased on the updated `main` so the operator IT re-runs against the fixed controller — `PUT /k8s/put` will then reach `/put` and return 200.

CrowleyRajapakse (approved):

PR #2106, opened 2026-06-04
fix(controller): enforce clean absolute basePath in upstreamDefinitions

### LINKED ISSUE #2105: [Bug]: malformed upstreamDefinitions basePath breaks routing
On deploy, the gateway accepts an `upstreamDefinitions[].basePath` with no value validation. A malformed basePath is accepted at deploy time (HTTP 201) but produces a broken route at runtime:

- basePath without a leading slash (e.g. `svc`): every request to the operation returns HTTP 400, and while that API is deployed any further deploy is rejected with HTTP 409, so one bad config blocks the deploy pipeline until it is removed.

- basePath with a trailing slash (e.g. `/svc/`): the upstream path becomes a double slash (`/svc//item`) and the backend answers with a 307 redirect instead of serving the request.

- basePath with surrounding or whitespace-only value (e.g. `" /svc "`, `"   "`): the value is consumed verbatim by the routing layers and produces a malformed path.

basePath is authoritative for the upstream path, so an invalid value should be rejected at deploy time with a clear message, the same way a path inside an upstream URL is already rejected ("URL must not include a path; set the base path in upstreamDefinitions[].basePath instead").

mehara-rothila:
## Purpose

Fixes #2105.

On deploy, the gateway accepted an `upstreamDefinitions[].basePath` with no value validation, so a malformed basePath deploys (HTTP 201) but breaks the route at runtime:

- without a leading slash (e.g. `svc`): every request to the operation returns HTTP 400, and while that API is deployed any further deploy is rejected with HTTP 409 (the bad config blocks the deploy pipeline until it is removed).
- with a trailing slash (e.g. `/svc/`): the upstream path becomes a double slash (`/svc//item`) and the backend returns a 307 redirect instead of serving the request.
- with a surrounding or whitespace-only value (e.g. `" /svc "`, `"   "`): the value is consumed verbatim by the xDS/transform layers and produces a malformed path.

## Approach

Validate the raw `basePath` value in `validateUpstreamDefinitions` (the raw value, because the xDS/transform layers consume it verbatim) and reject at deploy time with a clear message:

- a value with leading or trailing whitespace
- a value that does not start with `/`
- a value that ends with a trailing `/` (root `/` stays allowed)

A nil or empty basePath is left untouched; the runtime defaults it to root.

## Tests

Unit tests in `validator_test.go` cover the missing leading slash, trailing slash, whitespace cases (`" "`, `"   "`, `" /svc"`, `"/svc "`, `"/svc/ "`), and valid edge cases (`/`, empty, `/svc`, `/api/v2`). `go test ./pkg/config/` passes.

### DIFF
--- a/gateway/gateway-controller/pkg/config/api_validator.go
+++ b/gateway/gateway-controller/pkg/config/api_validator.go
@@ -354,6 +354,34 @@ func (v *APIValidator) validateUpstreamDefinitions(definitions *[]api.UpstreamDe
 			}
 		}
 
+		// basePath, when set, must be a clean absolute path: it has to start with
+		// "/", carry no trailing "/", and have no surrounding whitespace. The raw
+		// value is validated (not a trimmed copy) because the xDS/transform layers
+		// consume it verbatim, so a relative "svc" (Envoy 400 plus wedged deploys),
+		// a trailing "/svc/" (double slash "/svc//users"), or a padded " /svc " must
+		// all be rejected at deploy time rather than producing a malformed route.
+		if def.BasePath != nil {
+			bp := *def.BasePath
+			if bp != "" {
+				if strings.TrimSpace(bp) != bp {
+					errors = append(errors, ValidationError{
+						Field:   fmt.Sprintf("spec.upstreamDefinitions[%d].basePath", i),
+						Message: "basePath must not contain leading or trailing whitespace",
+					})
+				} else if !strings.HasPrefix(bp, "/") {
+					errors = append(errors, ValidationError{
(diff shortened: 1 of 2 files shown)


mehara-rothila:
@coderabbitai review

mehara-rothila:
@coderabbitai review

mehara-rothila:
Closing this PR. After further discussion we've decided not to move forward with adding dedicated basePath value validation at this time. Thanks to everyone who reviewed. Closing the linked issue #2105 as well.

PR #2137, opened 2026-06-10
feat(gateway): URL-stable identity-based cluster naming for API-level upstreams

mehara-rothila:
## Purpose

API-level Envoy cluster names were derived from the backend URL (`cluster_<scheme>_<sanitized-host>`). Because the name depended on the URL, editing a backend URL renamed the cluster: the old cluster was destroyed, a new one created, and routes repointed, draining in-flight connections on every URL change. This makes a routine config edit disruptive to live traffic.

## Goals

Make an API's main/sandbox cluster names stable for the lifetime of the API, so a backend-URL edit becomes an endpoint update on the same named cluster instead of a cluster rename. As a side effect, fix a latent collision where a main and sandbox upstream on the same URL, or two different APIs sharing a backend URL, collapsed into one shared cluster.

## Approach

Derive the cluster name from the API identity instead of the URL: `main_<hash>` / `sandbox_<hash>`, where the hash is the first 12 bytes of `sha256(apiID)` and the env is a plaintext `main_`/`sandbox_` prefix (not part of the hash). A single shared helper (`pkg/utils/clusterkey`) is used by both xDS builders (the Envoy xDS translator and the RuntimeDeployConfig transform path), so the two paths cannot diverge, and the two hand-rolled URL-sanitizing functions are removed. `ClusterKey` and `EnvoyClusterName` are unified to the same string so the policy engine never points Envoy at an unknown cluster (which would surface as 503 NoRoute).

Rollout is a one-time, in-memory effect: on the next config regeneration each API's clusters are renamed once, delivered atomically (new cluster plus repointed routes in a single xDS snapshot, validated by `snapshot.Consistent()`), so live traffic is not dropped. There is no database migration, the name is derived at translate time and never persisted.

## User stories

N/A. This is an internal gateway cluster-naming change with no user-facing API or behavior change.

## Documentation

N/A for external product docs. In-repo, the cluster-naming section of `gateway/spec/impls/1-basic-gateway-with-controller/data-model.md` is updated from the old URL-sanitized scheme to the identity-derived scheme.

## Automation tests

- Unit tests
  > New `pkg/utils/clusterkey` tests with known-answer vectors (including a UUID-shaped apiID). Translator and transformer tests assert URL-independence, that the name carries no URL-derived information, that `ClusterKey == EnvoyClusterName`, and that a main-only API registers no sandbox cluster or route. Covers the new helper and both builders' naming paths.
- Integration tests
  > `gateway/it/features/api-level-url-stable.feature`: a main upstream URL update changing both host and path, a sandbox URL update, default-cluster resolution under cluster_header routing, main and sandbox on the same backend host getting separate identity-named clusters, and two APIs on the same backend host routing independently (with a post-delete check that removing one does not disturb the other). The main and sandbox scenarios assert cluster identity through the Envoy admin endpoint, so they fail under the previous naming scheme.

## Security checks

- Ran FindSecurityBugs plugin and verified report? **N/A** (this change is in a Go module; FindSecurityBugs is a Java/SpotBugs plugin)

## Samples

No new configuration or API surface is introduced (cluster naming is internal), so this shows the observable behavior with any REST API.

Deploy an API with a main upstream:

```yaml
spec:
  upstream:
    main:
      url: http://backend-a:8080
  operations:
    - method: GET
      path: /users
```

The gateway names the Envoy cluster from the API identity, e.g. `main_5b50e1f10eb52155ebe4d947` (the `main_` prefix plus the first 12 bytes of `sha256(apiID)`), not from the URL.

Now change the upstream to a different host and port:

```yaml
spec:
  upstream:
    main:
      url: http://backend-b:9090
```

The cluster name stays `main_5b50e1f10eb52155ebe4d947`, so the route keeps pointing at the same named cluster and Envoy updates the endpoint in place. Under the old scheme the name would have changed from `cluster_http_backend-a_8080` to `cluster_http_backend-b_9090`, renaming the cluster on every URL edit.

## Related PRs

N/A.

## Test environment

GitHub Actions CI on Linux: Go unit tests and the gateway integration-test suite. No JDK, database, or browser dependency for this change.


mehara-rothila:
@coderabbitai full review

mehara-rothila:
@coderabbitai please do a full review of this PR , concrete suggestions, a risk analysis, and anything else you can surface. Please read the actual code, not just the description.

**Correctness & design**
- The cluster name is `<env>_<fragment>` where `fragment = first 8 bytes of sha256(apiID)` , the apiID ONLY. The env ("main"/"sandbox") is a plaintext prefix, NOT part of the hash, so main and sandbox of the same API share the fragment and are distinguished only by the prefix. Trace `clusterkey.APILevel` / `clusterkey.APILevelName` and both call sites: `pkg/xds/translator.go` (`resolveUpstreamCluster`) and `pkg/transform/restapi.go` (`addUpstreamCluster`). Is this sound?
- Confirm both xDS builders produce identical names for the same (env, apiID), so they can never disagree.
- Verify the env prefix alone guarantees main and sandbox never collide even though they share the hash fragment.
- Verify `ClusterKey` and `EnvoyClusterName` are the same string in the RDC path, and nothing still references the old URL-derived name.

**Risk analysis**
- Collision at 64 bits: realistic probability of two DISTINCT apiIDs hashing to the same 8-byte fragment (the only thing keeping different APIs apart), and what happens if they do , silent cross-wiring or fail-closed? Should the truncation be longer?
- Migration/one-time effects: existing APIs get clusters renamed on next redeploy. Any window where a route points at a not-yet-created cluster (503 NoRoute)? Is "new cluster + repointed routes" truly atomic in one snapshot?
- Observability: the `upstream_cluster` stat label changes from `cluster_<scheme>_<host>` to `<env>_<hash>`. What breaks, and is it called out?
- Cluster-count increase (two APIs sharing a backend URL no longer share a cluster): any resource concern at scale?
- Does removing `sanitizeClusterName` / `sanitizeEnvoyClusterName` drop any sanitizing some input still needs?

**Edge cases**
- Empty/unusual apiID, an API with only sandbox (no main), main vs sandbox pointing at identical URLs.

**Tests**
- Is unit + integration coverage enough to lock the contract? Anything missing , e.g. a negative test proving the old URL-derived name is gone, and a test proving main/sandbox of one API differ only by prefix?

mehara-rothila:
@coderabbitai please do a full review of this PR in its current state (head has changed since your last pass, it now also includes two new integration scenarios in api-level-url-stable.feature and minor doc fixes).

In particular, review all of the code comments across the PR for verbosity and redundancy. Several doc/inline comments (for example the doc comment on APILevel in pkg/utils/clusterkey/clusterkey.go, which is about 13 lines over a 2-line function) restate design rationale that is already in the PR description. Compared to the commenting conventions in neighbouring gateway-controller files, are any comments too long or redundant? Where you find some, suggest a concise replacement.

Also please cover, as a normal full review:
- correctness and any logic issues across the diff,
- the two new integration scenarios (main and sandbox on the same backend host; two APIs on the same backend host) , are they correct and asserting the right thing,
- any remaining risks, edge cases, or test gaps.

Base all style judgments on the repo's actual conventions, not a generic standard.

mehara-rothila:
@coderabbitai addressed your review in d79e44580:

- Trimmed the over-verbose doc comments to the repo norm: clusterkey.APILevel (13 to 3 lines), the package comment, resolveUpstreamCluster, and the TestAPILevel comment.
- Added a UUIDv7-shaped known-answer vector in clusterkey_test.go, so the algorithm is pinned against a realistic apiID.
- Scenario 5: added a post-delete routing check (delete API-B, assert API-A still routes), proving the two APIs own independent clusters.

Two points needed no code change: the "should be unchanged" step already uses exact set equality (strings.Join over sorted sets in steps_envoy_admin.go), and the "two distinct main_ count" assertion was intentionally skipped because a global /clusters count is unreliable (other deployed APIs also produce main_<hash> clusters); the post-delete check covers that property robustly.

Please do a final full review of this head, and as part of it verify one design question by reading the actual code across the repo (gateway-controller and platform-api):

The API-level cluster name hashes the API's UUID (clusterkey.APILevel(apiID), where apiID = cfg.UUID in xds and rdc.Metadata.UUID in transform). A reviewer noted that "users can give api ids," so please confirm precisely:
1. Where is the UUID that reaches clusterkey.APILevel actually assigned, and can a client or end user supply the hashed UUID itself (as opposed to the user-facing handle / id field)?
2. Can an empty, duplicate, or user-controlled UUID reach clusterkey.APILevel through any create/deploy path (the platform-api control plane, the gateway-controller deploy event, or direct gateway YAML via the gateway.api-platform.wso2.com/artifact-id annotation)?
3. Given that, is leaving clusterkey.APILevel without an in-function empty/uniqueness guard safe, or is there a path that genuinely needs one?

Please cite file:line for the assignment, validation, and uniqueness constraints, give a clear verdict on the three risks (empty apiID, duplicate UUID, UUID changing across a URL edit), and flag anything else you find in this final pass.

mehara-rothila:
@coderabbitai one final verification pass on head 10548bd0a, please, focused on production safety for EXISTING deployments. Read the actual code (the snapshot-building and startup config-load paths, the xDS translator, and pkg/policyxds).

This PR changes API-level Envoy cluster names from URL-derived (cluster_<scheme>_<host>) to identity-derived (main_<hash> / sandbox_<hash>, hash = sha256(apiID)[:12]). Please confirm whether rolling this out can crash, disrupt, or drop traffic for already-running APIs, and answer concretely:

1. Rollout / one-time rename: when the upgraded controller regenerates configs, existing clusters rename once. Is the new cluster plus the repointed routes delivered ATOMICALLY in a single xDS snapshot, so Envoy warms the new cluster and switches routes without dropping in-flight requests? Trace how the startup/bootstrap path loads existing configs and whether it produces one consolidated snapshot or N separate updates.

2. CDS-vs-RDS ordering: is there a real window where Envoy applies the new clusters before the repointed routes (route briefly pointing at a removed cluster -> 503 NoRoute), or is that mitigated by single-snapshot delivery and cluster warming? Is this risk new to this PR or inherent to xDS?

3. Connection handling: do in-flight requests on the old cluster drain gracefully during the rename, or can they be cut?

4. Persistence: does ANY persisted state (database, on-disk cache, control-plane records) store cluster names such that this needs a migration or backfill? Confirm the name is purely derived in memory and nothing in the DB references it. (The diff contains no .sql/schema changes; please confirm there is no hidden persistence dependency.)

5. Operational side effects: the upstream_cluster stat label changes once, and cluster count can rise when two APIs shared a backend URL. Confirm these are observability/resource effects only, not correctness or crash risks.

6. Revertability: is the change safe to roll back (reverting simply regenerates the old derived names), with no data left in an inconsistent state?

Give a clear verdict: does deploying this break or disrupt existing running APIs, and is anything required of operators beyond updating dashboards keyed on the old cluster names? Flag anything else you find in this final pass.

mehara-rothila:
@coderabbitai quick clarification on the operator note for #2137. The bundled Grafana dashboards in gateway/observability/grafana/dashboards/ (gateway-controller.json, infrastructure-overview.json, policy-engine.json) appear NOT to use the per-cluster upstream_cluster label. Can you confirm: do any shipped dashboards or alerts in this repo key on cluster names (upstream_cluster / cluster_http_)? So is the "update dashboards" action actually required for anything in-repo, or only for custom user-built per-cluster dashboards? Cite file:line.

mehara-rothila:
@coderabbitai is everything covered here, or is there anything still left to address?

gateway/gateway-controller/pkg/transform/restapi_test.go
```diff
@@ -383,7 +384,190 @@ func TestRestAPITransformer_SandboxRouteClusterHeader(t *testing.T) {
...
+func TestRestAPITransformer_APILevelClusterNameShape(t *testing.T) {
+	transformer := NewRestAPITransformer(testRouterCfg(), &config.Config{}, map[string]models.PolicyDefinition{})
+	cfg := makeRestAPIWithOps([]api.Operation{
+		{Method: "GET", Path: "/users"},
+	})
+
+	rdc, err := transformer.Transform(cfg)
+	require.NoError(t, err)
+
+	expectedMain := "main_" + clusterkey.APILevel(cfg.UUID)
```
renuka-fernando: let's use a hard-coded `expectedMain`. Otherwise, we won't catch if the func `clusterkey.APILevel` is changed.

mehara-rothila: @coderabbitai I hard-coded the expected cluster names in restapi_test.go as suggested (commit e03752f6c). Can you confirm that comment is resolved, and let me know if anything else is still pending on this PR?

mehara-rothila:
Closing in favor of #2012. The maintainer decided to incorporate this API-level URL-stable cluster naming into the per-operation upstream PR rather than land it separately, so #2012 now carries the same `clusterkey` hashing (`APILevel` / `APILevelName`) and the `api-level-url-stable.feature` integration scenarios. The cluster-naming work is fully covered there, so tracking it as part of #2012 from here.

PR #2144, opened 2026-06-10
fix(gateway): normalize operation HTTP method case

### LINKED ISSUE #2140: [Bug]: Lowercase HTTP methods cause Envoy routing 404

Area/Management (Management API or Management Portal UI), Area/Gateway (Routing, API deployment in gateway etc.)

### Description

API configuration validation allowed lowercase HTTP methods (such as `get`, `post`, etc.) to pass successfully. However, because these methods were stored as lowercase strings, the xDS translator generated Envoy route matches targeting lowercase methods. Standard HTTP clients send uppercase methods (e.g. `GET`), causing route-matching to fail with a `404 Not Found`.

## Root Cause

In [api_validator.go](https://github.com/wso2/api-platform/blob/main/gateway/gateway-controller/pkg/config/api_validator.go), the method validation was checking if `strings.ToUpper(string(op.Method))` was a valid HTTP method:
```go
} else if !validMethods[strings.ToUpper(string(op.Method))] {
    // ...
}
```
This correctly identified `get` as representing a valid method name, but the config struct itself was left unmodified (storing the lowercase `"get"`).

At translation time, [translator.go](https://github.com/wso2/api-platform/blob/main/gateway/gateway-controller/pkg/xds/translator.go) mapped this directly into Envoy's `:method` exact matcher:
```go
r.Match = &route.RouteMatch{
    Headers: []*route.HeaderMatcher{{
        Name: ":method",
        HeaderMatchSpecifier: &route.HeaderMatcher_StringMatch{
            StringMatch: &matcher.StringMatcher{
                MatchPattern: &matcher.StringMatcher_Exact{
                    Exact: method, // matched "get" exactly
                },
            },
        },
    }},
}
```

Since HTTP requests carry uppercase `:method` headers (like `"GET"`), Envoy's exact string matching failed.

## Impact

Any API configured with lowercase HTTP methods (e.g., `get` instead of `GET`) would successfully deploy to the gateway controller database but experience routing failures (`404 Not Found`) at runtime.

### Steps to Reproduce

1. Deploy an API configuration where one of the HTTP operations uses a lowercase method (e.g., `method: get`):
   ```yaml
   apiVersion: gateway.api-platform.wso2.com/v1alpha1
   kind: RestApi
   metadata:
     name: weather-api
   spec:
     context: /weather
     upstream:
       main:
         url: http://weather-backend:8080
     operations:
       - method: get
         path: /today
   ```

2. The API deployment succeeds with `201 Created` because the validator accepts lowercase methods.

3. Make an HTTP request through Envoy (the router) to the deployed route:
   ```bash
   curl -i -X GET http://localhost:8081/weather/today
   ```

4. **Expected Result:** The request is successfully routed to the backend upstream.
5. **Actual Result:** Envoy returns a `404 Not Found` response because the route only matches the literal lowercase `get`, while the request carries `GET`.

Severity/Minor (Non-critical functionality. Can be fixed in future releases)

*   **Operating System:** Windows
*   **Go Version:** `go1.26.2`
*   **Git Version:** `2.53.0.windows.1`
*   **Docker Version:** `29.1.4-rd, build 3c6914c`
*   **Envoy Router Version:** `v1.35.3` (envoyproxy/envoy:v1.35.3)
*   **Gateway-Controller Version:** `0.0.1-SNAPSHOT`

mehara-rothila:
## Summary

The xDS translator maps an operation's method straight into Envoy's case-sensitive `:method` exact matcher, so an API declaring a lower or mixed case method (for example `get`) deployed successfully but returned `404 Not Found` for every request, since HTTP clients send uppercase methods. The translator now normalizes the method to uppercase when building the route, so any-case methods route correctly.

## Why normalize instead of reject

An earlier revision of this PR rejected non-uppercase methods at validation time. That was changed to normalization because rejecting risks existing deployments: an API already running with a lowercase method would start failing validation on its next update or restart, a new failure it did not have before, while still not routing. Normalizing instead heals those configurations (404 becomes 200) on the next xDS rebuild without a redeploy, and adds no new failure mode for anyone. Uppercasing the method is also consistent with existing handling in the codebase: the on-prem APIM sync already uppercases the method for its verb field.

## Notes for reviewers

- Normalization is applied at the source in both REST route builders (`createRouteFromRDC` and `createRoute`), so the route name and the `:method` matcher stay consistent.
- Existing configurations with non-uppercase methods are healed on the next xDS rebuild. No migration or redeploy is required.
- The validator again accepts any letter case (it uppercases before the enum check), matching the router's actual matching once the method is normalized.
- The earlier deploy-time rejection and its unit tests are reverted in this PR.

## Test coverage

- New integration tests in `gateway/it/features/http-method-case.feature`: a lowercase method routes, a mixed-case method routes, and two methods on one path (mixed case) both deploy and route.
- `go build` and the `pkg/config` validator unit tests pass locally; the integration scenarios run in CI.

Fixes #2140


mehara-rothila:
@coderabbitai full review

mehara-rothila:
@coderabbitai review

mehara-rothila:
@coderabbitai review

mehara-rothila:
@coderabbitai review

mehara-rothila:
@coderabbitai does uppercasing the method only in the translator cover every route path, or could a method still reach Envoy un-normalized (policies, request transformation, route name)?

mehara-rothila:
@coderabbitai good catch. Addressed by normalizing the method inside `GenerateRouteName`, the single function that builds both the Envoy route name and the policy-engine key (per its own doc comment). Route names and policy keys are now uppercased consistently with the `:method` matcher across all paths (`createRoute`, the RDC transform path, and `policy/builder.go`), so the policy chain binds for lowercase/mixed-case methods too. Added `TestGenerateRouteName_UppercasesMethod`. Fixed in 53e969201.

mehara-rothila:
@coderabbitai please do a full review of the latest commits and tell us whether the HTTP method normalization fix is now complete, or if anything else is still needed. Specifically check: the :method matcher, route names and policy keys (GenerateRouteName), request transformation, and whether the unit + integration tests are sufficient.

tharindu1st (dismissed):

ashera96 (dismissed):

PR #2146, opened 2026-06-10
fix(gateway): reject identical main and sandbox vhosts

### LINKED ISSUE #2145: [Bug]: xDS paths disagree on same-vhost configs

Area/Gateway (Routing, API deployment in gateway etc.)

### Description

When a RestApi configures both a main and a sandbox upstream and both vhosts resolve to the same value, the two xDS builders disagree about whether the configuration is valid. The RuntimeDeployConfig transform path rejects it with an explicit guard, but the Envoy xDS translator path has no equivalent check and deploys it. The deployment API returns `201 Created`, while the transform failure is only logged and swallowed, so the main xDS gets routes for the API but the policy xDS gets no configuration at all. With the policy engine running fail-closed, every request to the API then returns `500`.

## Root Cause

The transform path guards against the collision in [restapi.go](https://github.com/wso2/api-platform/blob/main/gateway/gateway-controller/pkg/transform/restapi.go):

```go
// Guard: sandbox and main vhosts must differ, otherwise sandbox routes would
// overwrite main routes (same route key) and the sandbox patch would leave only
// a sandbox-cluster route with no main-cluster route at all.
if hasSandbox && effectiveMainVHost == effectiveSandboxVHost {
    return nil, fmt.Errorf("sandbox upstream is configured but resolves to the same vhost %q as the main upstream; configure distinct vhosts to avoid route conflicts", effectiveMainVHost)
}
```

The Envoy xDS translator in [translator.go](https://github.com/wso2/api-platform/blob/main/gateway/gateway-controller/pkg/xds/translator.go) has no equivalent guard: it builds the main routes and then appends sandbox routes for the same vhost, producing route entries with identical match criteria.

Because the transform failure does not fail the deployment, the request still returns `201`, and the error only appears in the controller log:

```
level=ERROR msg="Failed to upsert runtime config from replica sync"
error="failed to transform config: sandbox upstream is configured but resolves to the same vhost \"collide.local\" as the main upstream; configure distinct vhosts to avoid route conflicts"
```

## Impact

- An invalid configuration is accepted with `201` instead of being rejected at deploy time.
- The main xDS and policy xDS become inconsistent: Envoy has routes for the API, the policy engine has nothing.
- Every request to the API returns `500` at runtime (verified live), with no indication to the API author beyond a swallowed log line.

### Steps to Reproduce

1. Deploy a RestApi whose main and sandbox vhosts are identical:
   ```yaml
   apiVersion: gateway.api-platform.wso2.com/v1alpha1
   kind: RestApi
   metadata:
     name: same-vhost-demo-v1.0
   spec:
     displayName: Same-Vhost-Demo
     version: v1.0
     context: /same-vhost-demo/$version
     vhosts:
       main: collide.local
       sandbox: collide.local
     upstream:
       main:
         url: http://backend:8080/main-be
       sandbox:
         url: http://backend:8080/sb-be
     operations:
       - method: GET
         path: /endpoint
   ```

2. The deployment succeeds with `201 Created`.

3. Send a request through the router using the shared vhost:
   ```bash
   curl -i -H "Host: collide.local" http://localhost:8080/same-vhost-demo/v1.0/endpoint
   ```

4. **Expected Result:** The deployment is rejected with a `400` and the same clear error message the transform path produces, from both xDS paths consistently.
5. **Actual Result:** The deployment returns `201`, every request returns `500`, and the transform-path rejection only appears as a swallowed error in the controller logs.

Severity/Minor (Non-critical functionality. Can be fixed in future releases)

*   **Operating System:** Windows
*   **Go Version:** `go1.26.2`
*   **Git Version:** `2.53.0.windows.1`
*   **Docker Version:** `29.1.4-rd, build 3c6914c`
*   **Envoy Router Version:** `v1.35.3` (envoyproxy/envoy:v1.35.3)
*   **Gateway-Controller Version:** `0.0.1-SNAPSHOT`

mehara-rothila:
## Purpose

A RestApi with a sandbox upstream whose vhost resolves to the same value as the main vhost was rejected by the RuntimeDeployConfig transform path but accepted by the Envoy xDS translator. Because per-API translation errors are only logged and swallowed, the deployment returned `201 Created` while the policy xDS received no configuration for the API. With the policy engine running fail-closed, every request to the API then returned `500` at runtime, with no signal to the API author beyond a swallowed log line.

Resolves #2145

## Goals

Reject the same-vhost collision consistently and at deploy time, returning a clear `400` to the author instead of accepting a misconfiguration that silently fails at runtime. Ensure the two xDS builders can never disagree about whether a configuration is valid, since that disagreement was the root cause.

## Approach

The collision is now caught at every layer it can arise:

- Vhost resolution, which already froze router defaults into fully-omitted vhosts on create, now also fills blank or individually-omitted fields, and runs on the update (PUT) path in addition to create. Validation therefore always compares the effective vhosts, whether they come from explicit values, partial configurations, or router defaults.
- The API validator rejects identical main and sandbox vhosts (when a sandbox upstream is configured) with a `400` at deploy time. The comparison is case-insensitive and whitespace-trimmed, since Envoy matches domains case-insensitively.
- Both xDS builders (the Envoy translator and the RuntimeDeployConfig transform path) now apply the same case-insensitive guard, kept identical to each other, so they reject any remaining collision consistently and the inconsistent main-xDS / policy-xDS state can no longer occur.
- A gateway configured with identical main and sandbox vhost defaults logs a startup warning instead of failing to boot, so main-only APIs keep working while sandbox-enabled APIs are rejected per API.

## User stories

As an API developer, when I deploy an API whose main and sandbox vhosts collide, I want it rejected immediately with a clear error, rather than accepted and then failing every request at runtime.

## Documentation

N/A. This changes internal validation and xDS-translation behavior only; there is no user-facing API or product-documentation impact.

## Automation tests

 - Unit tests
   > Added validator cases (identical, case-insensitively identical, distinct, no-sandbox-upstream, absent vhosts); vhost-resolution cases (omitted/blank sandbox and blank main filled from defaults, existing sentinel/nil behaviour unchanged); and translator cases (explicit identical, case-only, and defaults-derived collisions all rejected). All existing unit tests in the affected packages (`pkg/config`, `pkg/utils`, `pkg/xds`, `pkg/transform`) pass.
 - Integration tests
   > No new integration tests; the change is covered by unit tests at the validator, the resolver, and both xDS builders. The existing integration suite is unaffected.

## Security checks

 - Ran FindSecurityBugs plugin and verified report? N/A (Go module; FindSecurityBugs is a Java/SpotBugs tool)

## Samples

Before the fix, this configuration deployed with `201` and then returned `500` on every request:

```yaml
apiVersion: gateway.api-platform.wso2.com/v1alpha1
kind: RestApi
metadata:
  name: same-vhost-demo-v1.0
spec:
  displayName: Same-Vhost-Demo
  version: v1.0
  context: /same-vhost-demo/$version
  vhosts:
    main: collide.local
    sandbox: collide.local
  upstream:
    main:
      url: http://backend:8080/main-be
    sandbox:
      url: http://backend:8080/sb-be
  operations:
    - method: GET
      path: /endpoint
```

After the fix, the same deployment is rejected at admission:

```
spec.vhosts: Main and sandbox vhosts must differ when a sandbox upstream is configured
```

## Related PRs

None.

## Test environment

 - Go (version per `gateway/gateway-controller/go.mod`)
 - Reproduced and verified on the gateway integration Docker stack (Linux containers) via Rancher Desktop on a Windows host
 - Controller storage backend: SQLite (default)

### DIFF
--- a/gateway/gateway-controller/cmd/controller/main.go
+++ b/gateway/gateway-controller/cmd/controller/main.go
@@ -124,7 +124,7 @@ func main() {
 	}
 
 	if strings.EqualFold(strings.TrimSpace(cfg.Router.VHosts.Main.Default), strings.TrimSpace(cfg.Router.VHosts.Sandbox.Default)) {
-		log.Warn("router.vhosts.main.default and router.vhosts.sandbox.default are identical: sandbox routes would collide with main routes, so REST API deployments will be rejected until the defaults differ")
+		log.Warn("router.vhosts.main.default and router.vhosts.sandbox.default are identical: sandbox routes would collide with main routes, so any REST API deployed without explicit distinct vhosts will be rejected until the defaults differ")
 	}
 
 	// In immutable mode, delete any stale SQLite files before opening the DB to
--- a/gateway/gateway-controller/cmd/controller/main.go
+++ b/gateway/gateway-controller/cmd/controller/main.go
@@ -124,7 +124,7 @@ func main() {
 	}
 
 	if strings.EqualFold(strings.TrimSpace(cfg.Router.VHosts.Main.Default), strings.TrimSpace(cfg.Router.VHosts.Sandbox.Default)) {
-		log.Warn("router.vhosts.main.default and router.vhosts.sandbox.default are identical: sandbox routes would collide with main routes, so APIs with sandbox upstreams will be rejected until the defaults differ")
+		log.Warn("router.vhosts.main.default and router.vhosts.sandbox.default are identical: sandbox routes would collide with main routes, so REST API deployments will be rejected until the defaults differ")
 	}
 
 	// In immutable mode, delete any stale SQLite files before opening the DB to
--- a/gateway/gateway-controller/cmd/controller/main.go
+++ b/gateway/gateway-controller/cmd/controller/main.go
@@ -38,10 +38,10 @@ import (
 	"github.com/wso2/api-platform/gateway/gateway-controller/pkg/policyxds"
 	"github.com/wso2/api-platform/gateway/gateway-controller/pkg/service/restapi"
 	"github.com/wso2/api-platform/gateway/gateway-controller/pkg/storage"
-	"github.com/wso2/api-platform/gateway/gateway-controller/pkg/webhooksecretxds"
 	"github.com/wso2/api-platform/gateway/gateway-controller/pkg/transform"
 	"github.com/wso2/api-platform/gateway/gateway-controller/pkg/utils"
 	"github.com/wso2/api-platform/gateway/gateway-controller/pkg/version"
+	"github.com/wso2/api-platform/gateway/gateway-controller/pkg/webhooksecretxds"
 	"github.com/wso2/api-platform/gateway/gateway-controller/pkg/xds"
 )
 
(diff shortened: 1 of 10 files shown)


mehara-rothila:
@coderabbitai full review

mehara-rothila:
@coderabbitai full review

PR #2155, opened 2026-06-11
feat(platform-api): support upstreamDefinitions pool and per-operation upstream overrides

mehara-rothila:
## Purpose

Add control-plane support for reusable upstream definitions and per-operation upstream routing on REST APIs, so individual operations of an API can route to different backends. This is the platform-api counterpart of the gateway-side change in PR #2012 (per-operation upstream routing in the gateway controller), which this PR pairs with: the platform stores, validates, and emits the configuration; the gateway enforces and routes it.

## Approach

- **Schema**: `RESTAPI` (and the create/update requests) gain an optional `upstreamDefinitions` pool of named, reusable upstreams (`name`, `basePath`, `timeout.connect`, weighted `upstreams[]`), and each operation gains an optional `upstream` override whose `main`/`sandbox` reference a pool entry by name (`ref` only, no inline URLs). The API-level `upstream.main`/`upstream.sandbox` become a url-or-ref union. Field names and constraints mirror the gateway's management OpenAPI contract exactly.
- **No DB migration**: the new fields ride in the existing `rest_apis.configuration` JSON blob.
- **Validation** (`validateUpstreamRefs`, run on create and on the merged config on update): definition names must be unique and match the shared `UpstreamReference` contract (1-100 chars, `^[a-zA-Z0-9\-_]+$`); definition URLs must be http/https host[:port] only (no path, query, or fragment); `basePath` must start with `/` and must not end with `/` (same regex as the gateway validator; empty is treated as unset since `omitempty` drops it from the deployment YAML); weights must be 0-100; `timeout.connect` must be a single unsigned ms/s/m/h unit and positive; every API-level and per-operation ref must resolve to a declared definition; the API-level union must set exactly one of url or ref. Violations map to HTTP 400.
- **Deployment YAML**: definitions and per-operation refs are emitted with the exact field names and casing the gateway parses; omitted optionals are omitted, zero weights are preserved. WebSub artifacts never emit definitions.
- **Update semantics**: partial updates merge first, then revalidate, so removing a pool entry that a stored operation still references is rejected rather than persisted as a dangling ref.

## Example

A REST API with a reusable pool, an API-level default upstream (url-or-ref), and one operation overridden to a pool entry:

```yaml
upstream:
  main:
    url: http://main:8080          # API-level default (a direct url, or a ref into the pool)

upstreamDefinitions:                # reusable, named upstream pool
  - name: alt-backend
    basePath: /api/v2
    timeout:
      connect: 5s
    upstreams:
      - url: http://alt:9090        # host and port only; path comes from basePath
        weight: 80

operations:
  - method: GET
    path: /whoami
    upstream:
      main:
        ref: alt-backend            # this operation routes to alt-backend
  - method: GET
    path: /ping                     # no override, falls back to the API-level upstream
```

## Validation parity with the gateway

Independent cross-checks of this branch against the gateway branch confirmed the accept/reject sets match for names, refs, URLs, weights, basePath, and timeout format, with two deliberate, safe differences: the platform rejects non-positive connect timeouts such as `0s` (the gateway's management validation accepts them but its translation layer then fails, so the platform enforces the effective contract), and the platform rejects an API-level upstream with neither url nor ref (previously accepted and then failed at deploy). A known gateway-internal item (weighted targets applying only on one translation path) is pre-existing on main and tracked separately.

## Testing

- Unit: `validateUpstreamRefs` covers names, duplicates, URLs (path/query/fragment/scheme/host), weights (0, 100, out-of-range), basePath (leading/trailing slash, bare `/`, space, query, empty-as-unset), timeout formats, and ref resolution for API-level and per-operation targets.
- Handler integration (real route/handler/service/repository stack over SQLite): unresolved refs and invalid basePath return 400 on create; a valid create returns 201; an update that dangles a stored ref returns 400.
- Deployment YAML round-trip tests for definitions, per-operation refs, WebSub omission, and model round-trip.

## Note on strictness

The service now rejects an API-level upstream with neither `url` nor `ref` on create and on the merged configuration of updates, closing a previously accepted configuration that could only fail at deploy time. Practical HTTP behavior is unchanged, since the handlers already required at least one upstream endpoint on both create and update.

### DIFF
--- a/platform-api/internal/dto/api.go
+++ b/platform-api/internal/dto/api.go
@@ -137,16 +137,16 @@ type Vhosts struct {
 
 // APIYAMLData represents a basic spec section of the API deployment YAML
 type APIYAMLData struct {
-	DisplayName         string                   `yaml:"displayName"`
-	Version             string                   `yaml:"version"`
-	Context             string                   `yaml:"context"`
-	SubscriptionPlans   []string                 `yaml:"subscriptionPlans,omitempty"`
-	Vhosts              *Vhosts                  `yaml:"vhosts,omitempty"`
-	Upstream            *UpstreamYAML            `yaml:"upstream,omitempty"`
-	UpstreamDefinitions []UpstreamDefinitionYAML `yaml:"upstreamDefinitions,omitempty"`
-	Policies            []Policy                 `yaml:"policies,omitempty"`
-	Operations          []api.OperationRequest   `yaml:"operations,omitempty"`
-	Channels            []api.ChannelRequest     `yaml:"channels,omitempty"`
+	DisplayName         string                 `yaml:"displayName"`
+	Version             string                 `yaml:"version"`
+	Context             string                 `yaml:"context"`
+	SubscriptionPlans   []string               `yaml:"subscriptionPlans,omitempty"`
+	Vhosts              *Vhosts                `yaml:"vhosts,omitempty"`
+	Upstream            *UpstreamYAML          `yaml:"upstream,omitempty"`
+	UpstreamDefinitions []ReusableUpstreamYAML `yaml:"upstreamDefinitions,omitempty"`
+	Policies            []Policy               `yaml:"policies,omitempty"`
+	Operations          []api.OperationRequest `yaml:"operations,omitempty"`
+	Channels            []api.ChannelRequest   `yaml:"channels,omitempty"`
 }
 
 // UpstreamYAML represents the upstream configuration for API deployment YAML
@@ -161,21 +161,21 @@ type UpstreamTarget struct {
 	Ref string `yaml:"ref,omitempty"`
 }
 
-// UpstreamDefinitionYAML represents a reusable named upstream definition in the deployment YAML
-type UpstreamDefinitionYAML struct {
+// ReusableUpstreamYAML represents a named entry in the deployment YAML's upstreamDefinitions pool.
+type ReusableUpstreamYAML struct {
 	Name      string                `yaml:"name"`
 	BasePath  string                `yaml:"basePath,omitempty"`
 	Timeout   *UpstreamTimeoutYAML  `yaml:"timeout,omitempty"`
 	Upstreams []UpstreamBackendYAML `yaml:"upstreams"`
 }
 
-// UpstreamBackendYAML represents a single backend target within an upstream definition.
+// UpstreamBackendYAML represents a weighted backend target within a reusable upstream.
 type UpstreamBackendYAML struct {
 	URL    string `yaml:"url"`
 	Weight *int   `yaml:"weight,omitempty"`
 }
 
-// UpstreamTimeoutYAML represents the timeout configuration for an upstream definition.
+// UpstreamTimeoutYAML represents the timeout configuration for a reusable upstream.
 type UpstreamTimeoutYAML struct {
 	Connect string `yaml:"connect,omitempty"`
 }
--- a/platform-api/internal/dto/api.go
+++ b/platform-api/internal/dto/api.go
@@ -161,9 +161,7 @@ type UpstreamTarget struct {
 	Ref string `yaml:"ref,omitempty"`
 }
 
-// UpstreamDefinitionYAML represents a reusable named upstream definition in the
-// deployment YAML (the API-level upstreamDefinitions array the gateway resolves
-// refs against).
+// UpstreamDefinitionYAML represents a reusable named upstream definition in the deployment YAML
 type UpstreamDefinitionYAML struct {
 	Name      string                `yaml:"name"`
 	BasePath  string                `yaml:"basePath,omitempty"`
--- a/platform-api/internal/dto/api.go
+++ b/platform-api/internal/dto/api.go
@@ -137,15 +137,16 @@ type Vhosts struct {
 
 // APIYAMLData represents a basic spec section of the API deployment YAML
 type APIYAMLData struct {
-	DisplayName       string                 `yaml:"displayName"`
-	Version           string                 `yaml:"version"`
-	Context           string                 `yaml:"context"`
-	SubscriptionPlans []string               `yaml:"subscriptionPlans,omitempty"`
-	Vhosts            *Vhosts                `yaml:"vhosts,omitempty"`
-	Upstream          *UpstreamYAML          `yaml:"upstream,omitempty"`
-	Policies          []Policy               `yaml:"policies,omitempty"`
-	Operations        []api.OperationRequest `yaml:"operations,omitempty"`
-	Channels          []api.ChannelRequest   `yaml:"channels,omitempty"`
+	DisplayName         string                   `yaml:"displayName"`
+	Version             string                   `yaml:"version"`
+	Context             string                   `yaml:"context"`
+	SubscriptionPlans   []string                 `yaml:"subscriptionPlans,omitempty"`
+	Vhosts              *Vhosts                  `yaml:"vhosts,omitempty"`
+	Upstream            *UpstreamYAML            `yaml:"upstream,omitempty"`
+	UpstreamDefinitions []UpstreamDefinitionYAML `yaml:"upstreamDefinitions,omitempty"`
+	Policies            []Policy                 `yaml:"policies,omitempty"`
+	Operations          []api.OperationRequest   `yaml:"operations,omitempty"`
+	Channels            []api.ChannelRequest     `yaml:"channels,omitempty"`
 }
 
 // UpstreamYAML represents the upstream configuration for API deployment YAML
@@ -160,10 +161,30 @@ type UpstreamTarget struct {
 	Ref string `yaml:"ref,omitempty"`
 }
 
+// UpstreamDefinitionYAML represents a reusable named upstream definition in the
+// deployment YAML (the API-level upstreamDefinitions array the gateway resolves
+// refs against).
+type UpstreamDefinitionYAML struct {
+	Name      string                `yaml:"name"`
+	BasePath  string                `yaml:"basePath,omitempty"`
+	Timeout   *UpstreamTimeoutYAML  `yaml:"timeout,omitempty"`
+	Upstreams []UpstreamBackendYAML `yaml:"upstreams"`
+}
+
+// UpstreamBackendYAML represents a single backend target within an upstream definition.
+type UpstreamBackendYAML struct {
+	URL    string `yaml:"url"`
+	Weight *int   `yaml:"weight,omitempty"`
+}
+
+// UpstreamTimeoutYAML represents the timeout configuration for an upstream definition.
+type UpstreamTimeoutYAML struct {
+	Connect string `yaml:"connect,omitempty"`
+}
+
 // APIListResponse represents a paginated list of APIs (constitution-compliant)
 type APIListResponse struct {
 	Count      int        `json:"count" yaml:"count"`           // Number of items in current response
 	List       []*API     `json:"list" yaml:"list"`             // Array of API objects
 	Pagination Pagination `json:"pagination" yaml:"pagination"` // Pagination metadata
 }
-
--- a/platform-api/internal/dto/api.go
+++ b/platform-api/internal/dto/api.go
@@ -143,7 +143,7 @@ type APIYAMLData struct {
 	SubscriptionPlans   []string               `yaml:"subscriptionPlans,omitempty"`
 	Vhosts              *Vhosts                `yaml:"vhosts,omitempty"`
 	Upstream            *UpstreamYAML          `yaml:"upstream,omitempty"`
-	UpstreamDefinitions []ReusableUpstreamYAML `yaml:"upstreamDefinitions,omitempty"`
+	UpstreamDefinitions []ReusableUpstream     `yaml:"upstreamDefinitions,omitempty"`
 	Policies            []Policy               `yaml:"policies,omitempty"`
 	Operations          []api.OperationRequest `yaml:"operations,omitempty"`
 	Channels            []api.ChannelRequest   `yaml:"channels,omitempty"`
@@ -161,22 +161,22 @@ type UpstreamTarget struct {
 	Ref string `yaml:"ref,omitempty"`
 }
 
-// ReusableUpstreamYAML represents a named entry in the deployment YAML's upstreamDefinitions pool.
-type ReusableUpstreamYAML struct {
-	Name      string                `yaml:"name"`
-	BasePath  string                `yaml:"basePath,omitempty"`
-	Timeout   *UpstreamTimeoutYAML  `yaml:"timeout,omitempty"`
-	Upstreams []UpstreamBackendYAML `yaml:"upstreams"`
+// ReusableUpstream represents a named entry in the deployment YAML's upstreamDefinitions pool.
+type ReusableUpstream struct {
+	Name      string            `yaml:"name"`
+	BasePath  string            `yaml:"basePath,omitempty"`
+	Timeout   *UpstreamTimeout  `yaml:"timeout,omitempty"`
+	Upstreams []UpstreamBackend `yaml:"upstreams"`
 }
 
-// UpstreamBackendYAML represents a weighted backend target within a reusable upstream.
-type UpstreamBackendYAML struct {
+// UpstreamBackend represents a weighted backend target within a reusable upstream.
+type UpstreamBackend struct {
 	URL    string `yaml:"url"`
 	Weight *int   `yaml:"weight,omitempty"`
 }
 
-// UpstreamTimeoutYAML represents the timeout configuration for a reusable upstream.
-type UpstreamTimeoutYAML struct {
+// UpstreamTimeout represents the timeout configuration for a reusable upstream.
+type UpstreamTimeout struct {
 	Connect string `yaml:"connect,omitempty"`
 }
 
--- a/platform-api/internal/dto/api.go
+++ b/platform-api/internal/dto/api.go
@@ -143,7 +143,7 @@ type APIYAMLData struct {
 	SubscriptionPlans   []string               `yaml:"subscriptionPlans,omitempty"`
 	Vhosts              *Vhosts                `yaml:"vhosts,omitempty"`
 	Upstream            *UpstreamYAML          `yaml:"upstream,omitempty"`
-	UpstreamDefinitions []ReusableUpstream     `yaml:"upstreamDefinitions,omitempty"`
+	UpstreamDefinitions []NamedUpstream        `yaml:"upstreamDefinitions,omitempty"`
 	Policies            []Policy               `yaml:"policies,omitempty"`
 	Operations          []api.OperationRequest `yaml:"operations,omitempty"`
 	Channels            []api.ChannelRequest   `yaml:"channels,omitempty"`
@@ -161,8 +161,8 @@ type UpstreamTarget struct {
 	Ref string `yaml:"ref,omitempty"`
 }
 
-// ReusableUpstream represents a named entry in the deployment YAML's upstreamDefinitions pool.
-type ReusableUpstream struct {
+// NamedUpstream represents a named entry in the deployment YAML's upstreamDefinitions pool.
+type NamedUpstream struct {
 	Name      string            `yaml:"name"`
 	BasePath  string            `yaml:"basePath,omitempty"`
 	Timeout   *UpstreamTimeout  `yaml:"timeout,omitempty"`
--- a/platform-api/internal/model/api.go
+++ b/platform-api/internal/model/api.go
@@ -43,14 +43,15 @@ type API struct {
 }
 
 type RestAPIConfig struct {
-	Name              string         `json:"name,omitempty"`
-	Version           string         `json:"version,omitempty"`
-	Context           *string        `json:"context,omitempty"`
-	Transport         []string       `json:"transport,omitempty"`
-	Upstream          UpstreamConfig `json:"upstream,omitempty"`
-	Policies          []Policy       `json:"policies,omitempty"`
-	Operations        []Operation    `json:"operations,omitempty"`
-	SubscriptionPlans []string       `json:"subscriptionPlans,omitempty"`
+	Name                string             `json:"name,omitempty"`
+	Version             string             `json:"version,omitempty"`
+	Context             *string            `json:"context,omitempty"`
+	Transport           []string           `json:"transport,omitempty"`
+	Upstream            UpstreamConfig     `json:"upstream,omitempty"`
+	UpstreamDefinitions []ReusableUpstream `json:"upstreamDefinitions,omitempty"`
+	Policies            []Policy           `json:"policies,omitempty"`
+	Operations          []Operation        `json:"operations,omitempty"`
+	SubscriptionPlans   []string           `json:"subscriptionPlans,omitempty"`
 }
 
 // TableName returns the table name for the API model
@@ -84,9 +85,10 @@ type Channel struct {
 
 // OperationRequest represents operation request details
 type OperationRequest struct {
-	Method   string   `json:"method,omitempty"`
-	Path     string   `json:"path,omitempty"`
-	Policies []Policy `json:"policies,omitempty"`
+	Method   string             `json:"method,omitempty"`
+	Path     string             `json:"path,omitempty"`
+	Policies []Policy           `json:"policies,omitempty"`
+	Upstream *OperationUpstream `json:"upstream,omitempty"`
 }
 
 // ChannelRequest represents channel request details
--- a/platform-api/internal/model/api.go
+++ b/platform-api/internal/model/api.go
@@ -43,15 +43,15 @@ type API struct {
 }
 
 type RestAPIConfig struct {
-	Name                string             `json:"name,omitempty"`
-	Version             string             `json:"version,omitempty"`
-	Context             *string            `json:"context,omitempty"`
-	Transport           []string           `json:"transport,omitempty"`
-	Upstream            UpstreamConfig     `json:"upstream,omitempty"`
-	UpstreamDefinitions []ReusableUpstream `json:"upstreamDefinitions,omitempty"`
-	Policies            []Policy           `json:"policies,omitempty"`
-	Operations          []Operation        `json:"operations,omitempty"`
-	SubscriptionPlans   []string           `json:"subscriptionPlans,omitempty"`
+	Name                string          `json:"name,omitempty"`
+	Version             string          `json:"version,omitempty"`
+	Context             *string         `json:"context,omitempty"`
+	Transport           []string        `json:"transport,omitempty"`
+	Upstream            UpstreamConfig  `json:"upstream,omitempty"`
+	UpstreamDefinitions []NamedUpstream `json:"upstreamDefinitions,omitempty"`
+	Policies            []Policy        `json:"policies,omitempty"`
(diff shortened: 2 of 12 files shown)


mehara-rothila:
@coderabbitai review

platform-api/resources/openapi.yaml
```diff
@@ -8404,6 +8409,60 @@ components:
...
+    UpstreamTimeout:
+      title: Upstream Timeout
+      type: object
+      description: Timeout configuration for upstream requests
+      properties:
+        connect:
+          type: string
+          description: Connection timeout duration (e.g., "5s", "500ms")
+          pattern: '^\d+(\.\d+)?(ms|s|m|h)$'
+          example: 5s
```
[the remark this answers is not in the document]
mehara-rothila: Both intentional. This schema is a deliberate mirror of the gateway's authoritative `UpstreamDefinition`/`UpstreamTimeout`: same single-unit `connect` pattern, same `format: uri`, no `basePath` pattern.

The gateway enforces host-only URLs and duration parsing in Go (`api_validator.go`), not in the schema. I'm keeping the schema identical and adding the matching value-level validation in the follow-up validation commit on this branch.

platform-api/resources/openapi.yaml
```diff
@@ -8404,6 +8409,60 @@ components:
...
+            type: object
+            additionalProperties: false
+            required:
+              - url
+            properties:
+              url:
+                type: string
+                format: uri
+                description: Backend URL (host and port only; path comes from basePath)
+                example: http://prod-backend-1:5000
```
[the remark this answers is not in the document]
mehara-rothila: Same intent as the timeout thread: this mirrors the gateway's authoritative `UpstreamDefinition`, which keeps `format: uri` with no `basePath` pattern and enforces host-only URLs in Go (`api_validator.go`), not the schema. The matching value-level validation lands in the follow-up validation commit on this branch.

platform-api/resources/openapi.yaml
```diff
@@ -10228,6 +8409,60 @@ components:
           description: Authentication value (API key, Bearer token, or Base64 encoded credentials for basic auth)
           example: my-api-key-value
 
+    ReusableUpstream:
```
renuka-fernando:
```suggestion
    UpstreamDefinition:
```

What about UpstreamDefinition here?
cc: @tharindu1st

platform-api/resources/openapi.yaml
```diff
@@ -10228,6 +8409,60 @@ components:
...
+      description: A reusable named upstream definition. Referenced by name from an API-level upstream.ref.
+      properties:
+        name:
+          type: string
+          description: Unique identifier for this upstream definition
+          minLength: 1
+          maxLength: 100
+          pattern: '^[a-zA-Z0-9\-_]+$'
+          example: my-upstream-1
+        basePath:
```
renuka-fernando, line 6866: pattern for basePath?

platform-api/src/resources/openapi.yaml
```diff
@@ -8006,6 +6326,11 @@ components:
           example: "2023-10-12T10:30:00Z"
         upstream:
           $ref: "#/components/schemas/Upstream"
```
renuka-fernando: upstream should have to support ref as well right?

platform-api/resources/openapi.yaml
mehara-rothila: Removed

platform-api/resources/openapi.yaml
mehara-rothila: @tharindu1st @renuka-fernando this is a **3-way name swap**, not a simple rename.

**Today in `openapi.yaml`:**

| Schema | Role | Line |
|---|---|---|
| `Upstream` | the `{main, sandbox}` **wrapper** | [main L6791](https://github.com/wso2/api-platform/blob/main/platform-api/resources/openapi.yaml#L6791) |
| `UpstreamDefinition` | the `url`/`ref` **endpoint** | [main L6802](https://github.com/wso2/api-platform/blob/main/platform-api/resources/openapi.yaml#L6802) |
| `ReusableUpstream` | the **pool entry** (added by this PR) | [PR head L8447](https://github.com/wso2/api-platform/blob/9ceebf6d5b6fb719082b6356d1fbdcc4b5053455/platform-api/src/resources/openapi.yaml#L8447) |

(`Upstream` is shared by API / LLMProvider / MCPServer / WebSub, so it has to stay. `ReusableUpstream` is the one schema this PR introduces, so it's the only row not yet on `main`.)

**Two things point the same way:**
- The gateway-controller names the **pool entry** `UpstreamDefinition` ([management-openapi.yaml#L4210](https://github.com/wso2/api-platform/blob/main/gateway/gateway-controller/api/management-openapi.yaml#L4210)).
- Our **internal model already** names the endpoint `UpstreamEndpoint` (see [platform-api/internal/model/upstream.go#L27](https://github.com/wso2/api-platform/blob/main/platform-api/internal/model/upstream.go#L27)).

We can't fully match the gateway since it has no wrapper (it inlines `{main, sandbox}`).

**Proposal:** keep the `Upstream` wrapper, then rename:
- endpoint `UpstreamDefinition` → `UpstreamEndpoint`
- pool entry `ReusableUpstream` → `UpstreamDefinition`

This aligns the schema with both the gateway's pool name and our internal model. Good to proceed?

mehara-rothila:
@coderabbitai a naming sanity-check on the upstream schemas in this PR, please.

Context: this PR (#2155) is the **platform-api / control-plane** side of the per-operation upstream feature. Its gateway / data-plane companion is **#2012** (https://github.com/wso2/api-platform/pull/2012), which adds the matching schemas to `gateway/gateway-controller/api/management-openapi.yaml`. The two sides must use matching schema shapes and, ideally, matching names. Please review the naming with both in mind.

Current state in `platform-api/src/resources/openapi.yaml` (this PR):
- `Upstream` , the `{main, sandbox}` wrapper, shared by RESTAPI, LLMProvider, MCPServer and WebSub, so it cannot be renamed.
- `UpstreamDefinition` , a single endpoint (`url` or `ref` + `auth`), used only by `Upstream.main`/`sandbox`.
- `ReusableUpstream` , a named pool entry (`name`, `basePath`, `timeout`, `upstreams[]`), held by the new `upstreamDefinitions` array.

How the gateway side names the same things:
- On `main`, `gateway/gateway-controller/api/management-openapi.yaml` names the **pool entry** `UpstreamDefinition` (the `upstreamDefinitions` array and `UpstreamDefinition` schema around L4155 / L4210). The gateway has no `Upstream` wrapper; it inlines `{main, sandbox}`.
- #2012 additionally adds `UpstreamReference` (the ref-name string, reused by both the pool entry's `name` and per-operation refs) and `OperationUpstream` (the per-operation `{main, sandbox}` override). These are on the #2012 branch, not yet on `main`.

The clash: our field is `upstreamDefinitions`, but its items are `ReusableUpstream`, while a different schema here is named `UpstreamDefinition`. The gateway's pool name (`UpstreamDefinition`) is already taken on our side by the endpoint schema.

Proposed rename, keeping `Upstream` as-is:
- `ReusableUpstream` -> `UpstreamDefinition` (so the `upstreamDefinitions` items match both the field name and the gateway pool name)
- endpoint `UpstreamDefinition` -> `UpstreamEndpoint`

Please advise:
1. Is this the clearest naming, or do you have a better suggestion?
2. Any `$ref`, collision, or backward-compatibility risk we are missing?
3. Does this align the control-plane (#2155) and data-plane (#2012) naming as closely as the shared `Upstream` wrapper allows?

mehara-rothila:
@coderabbitai we're renaming the `ReusableUpstream` schema in `platform-api/src/resources/openapi.yaml` (the pool entry held by `upstreamDefinitions`: `name`, `basePath`, `timeout`, weighted `upstreams[]`). The gateway-controller names this same concept `UpstreamDefinition`. Note platform-api already uses `UpstreamDefinition` for a different schema (the url/ref endpoint under `Upstream.main`/`sandbox`), so adopting `UpstreamDefinition` for the pool would rename that endpoint schema to `UpstreamEndpoint`.

Our candidates:
1. `UpstreamDefinition` (matches the gateway pool name; requires endpoint -> `UpstreamEndpoint`)
2. `ReusableUpstream` (current)
3. `NamedUpstream`
4. `UpstreamPoolEntry`
5. `SharedUpstream`

Please (a) add any stronger options, then (b) rank ALL of them best-to-worst with a one-line rationale each. Judge by: cross-plane consistency with the gateway (which calls this `UpstreamDefinition`), clarity, and fit with the sibling schemas (`Upstream`, `UpstreamReference`, `OperationUpstream`, `UpstreamTimeout`).

mehara-rothila:
@coderabbitai for the agreed rename, please show a focused before/after of ONLY the affected schema blocks, not the full file: the `Upstream` wrapper's `main`/`sandbox` refs, the endpoint schema (`UpstreamDefinition` -> `UpstreamEndpoint`), the pool schema (`ReusableUpstream` -> `UpstreamDefinition`), and the `RESTAPI.upstreamDefinitions.items` ref. Keep it to those blocks so the concrete diff is easy to review.

mehara-rothila:
@coderabbitai full review

This is the control-plane (platform-api) side of per-operation upstream routing. The data-plane contract it must match is in PR #2012 (gateway-controller/pkg/config/api_validator.go and management-openapi.yaml). Please verify each of these and flag any divergence:

- Validation parity with the gateway validator: upstreamDefinition name pattern/length, duplicate names, backend URL host-only http/https (path, query, or fragment rejected), weight 0-100, connect timeout format + positive + blank-treated-as-unset, API-level url-xor-ref, and that every API-level and per-operation ref resolves to a declared pool entry.
- Validation runs on create and on the merged result of an update (an update must not be able to remove a pool entry a surviving operation still references), and surfaces as HTTP 400 on create, update, project import, and OpenAPI import.
- Persistence round-trip through the rest_apis.configuration JSON blob: no field dropped, and rows written before this change still decode.
- Deployment-YAML emission matches the gateway shape (spec.upstreamDefinitions and operations[].upstream.main/sandbox.ref) and is emitted for REST APIs only, not WebSub.
- Nil-safety on the generated OperationUpstream main/sandbox pointers, and that api/generated.go is regeneration-only (no hand edits).
- Whether the added tests actually exercise these paths.

mehara-rothila:
@coderabbitai we've decided to keep the pool entry schema named ReusableUpstream and are not renaming it to UpstreamDefinition. My earlier comment that called the rename "agreed" was premature; on reflection we are not doing it in this PR.

Reasoning: adopting UpstreamDefinition for the pool would also force renaming the pre-existing UpstreamDefinition endpoint schema (the url/ref endpoint under Upstream.main/sandbox) to UpstreamEndpoint, which churns a large amount of pre-existing code (dto, model, service, utils, and their tests) that is unrelated to this feature. This PR is scoped to adding the upstreamDefinitions pool and per-operation upstream overrides, and we want to keep that scope tight. The schema shapes already match the gateway contract in #2012; only the pool entry name differs, and ReusableUpstream reads clearly for a reusable named entry in the pool.

Please note this decision and do not re-flag the ReusableUpstream -> UpstreamDefinition rename in future reviews of this PR. If the maintainers prefer the rename, we will do it as a separate follow-up.

platform-api/src/internal/handler/api.go
```diff
@@ -159,6 +159,12 @@ func (h *APIHandler) CreateAPI(c *gin.Context) {
 				err.Error()))
 			return
 		}
+		if errors.Is(err, constants.ErrInvalidUpstreamDefinition) {
+			h.slogger.Error("Invalid upstream definition", "organizationId", orgId, "error", err)
+			c.JSON(http.StatusBadRequest, utils.NewErrorResponse(400, "Bad Request",
+				err.Error()))
+			return
```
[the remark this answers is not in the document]
mehara-rothila: @coderabbitai We're keeping the detailed message here intentionally. This detail is built exclusively from the request itself (validation field paths like upstreamDefinitions[0].url and the client's own submitted values); no internal state can reach this branch, since repository and unexpected errors fall through to the fixed-message 500 below. Returning err.Error() for 400 validation failures is this handler's established convention (the adjacent ErrSubscriptionPlanNotFoundOrInactive branch does the same), and an actionable message telling the caller which field failed is the purpose of this control-plane validation. Please record this decision for this PR.

platform-api/src/internal/service/api_test.go
```diff
@@ -499,3 +515,530 @@ func createStatusPtr(s string) *api.CreateRESTAPIRequestLifeCycleStatus {
...
+	if _, err := sqlDB.Exec("PRAGMA foreign_keys = ON"); err != nil {
+		t.Fatalf("enable foreign keys: %v", err)
+	}
+	db := &database.DB{DB: sqlDB}
+
+	// Resolve the schema path relative to this test file so the helper does not depend
+	// on the test's working directory.
+	_, thisFile, _, _ := runtime.Caller(0)
+	schemaPath := filepath.Join(filepath.Dir(thisFile), "..", "database", "schema.sqlite.sql")
+	schema, err := os.ReadFile(schemaPath)
```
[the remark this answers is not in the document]
mehara-rothila: @coderabbitai This trim-for-validation-only behavior is deliberate gateway parity, and a padded reference is not an invalid value downstream. The merged gateway (PR #2012) trims the ref at deploy-time validation (api_validator.go:274 and :714) and again at runtime resolution: upstreamref.FindByName trims both the ref and each definition name before comparing, so a whitespace-padded ref validates, resolves, and routes end-to-end exactly as an unpadded one. Rejecting it here would make the control plane stricter than the data plane and could newly fail an already-stored, working configuration on its next update, and normalizing the stored value would silently change what reads return for a value the client submitted. We are keeping the platform aligned with the gateway's operative behavior. Please record this decision for this PR.

platform-api/src/internal/service/api.go
```diff
@@ -657,6 +663,182 @@ func (s *APIService) validateSubscriptionPlans(planNames *[]string, orgUUID stri
...
+// names a declared upstreamDefinition. field is the full path of the ref, for example
+// "upstream.main.ref" or "operations[0].upstream.main.ref".
+func resolveUpstreamRef(ref, field string, defined map[string]bool) error {
+	refName := strings.TrimSpace(ref)
+	if err := validateUpstreamReferenceName(refName, field); err != nil {
+		return err
+	}
+	if !defined[refName] {
+		return fmt.Errorf("%s references upstream definition %q which is not declared in upstreamDefinitions", field, refName)
+	}
```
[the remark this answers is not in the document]
mehara-rothila: @coderabbitai This trim-for-validation-only behavior is deliberate gateway parity, and a padded reference is not an invalid value downstream. The merged gateway (PR #2012) trims the ref at deploy-time validation (api_validator.go:274 and :714) and again at runtime resolution: upstreamref.FindByName trims both the ref and each definition name before comparing, so a whitespace-padded ref validates, resolves, and routes end-to-end exactly as an unpadded one. Rejecting it here would make the control plane stricter than the data plane and could newly fail an already-stored, working configuration on its next update, and normalizing the stored value would silently change what reads return for a value the client submitted. We are keeping the platform aligned with the gateway's operative behavior. Please record this decision for this PR.

platform-api/src/internal/service/api.go
```diff
@@ -657,6 +663,182 @@ func (s *APIService) validateSubscriptionPlans(planNames *[]string, orgUUID stri
...
+		}
+		parsed, err := url.Parse(*endpoint.Url)
+		if err != nil {
+			return fmt.Errorf("%s.url is invalid: %w", field, err)
+		}
+		if parsed.Scheme != "http" && parsed.Scheme != "https" {
+			return fmt.Errorf("%s.url must use http or https scheme", field)
+		}
+		if parsed.Host == "" {
+			return fmt.Errorf("%s.url must include a host", field)
```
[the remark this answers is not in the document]
mehara-rothila: @coderabbitai We're keeping the parsed.Host check to stay in exact parity with the merged gateway validator, which checks parsedURL.Host == "" in both of its paths (api_validator.go:222 and :368) and never consults Hostname(). Switching the platform to Hostname() would reject input such as http://:8080 that the gateway currently accepts at deploy, making the control plane stricter than the data plane and risking a 400 on the next update of an already-stored configuration. If this hardening is wanted, the right sequence is to add the Hostname() check to the gateway validator first and mirror it here in lockstep as a follow-up. Please record this decision for this PR.

mehara-rothila:
@coderabbitai review

mehara-rothila:
@coderabbitai Please review the rebased changes and additionally verify cross-PR consistency with PR #2012, which implements the gateway-side counterpart of this feature (per-operation upstream routing in the gateway controller). Specifically, check whether the validation rules and wire format in this PR match the gateway contract on that PR's head:

- upstreamDefinitions entry rules: name pattern and length, basePath pattern, timeout.connect format, upstreams url restrictions (host and port only, no path, query, or fragment), and the weight range
- the API-level upstream url-or-ref union (exactly one of url or ref)
- operations[].request.upstream main and sandbox refs resolving against declared definitions
- the deployment YAML field names and casing this PR emits versus what the gateway parser expects

Please report any mismatch you find between the two sides, and any issues in this PR's own changes (including the error handling migration to apperror.ValidationFailed and the new handler integration tests).

mehara-rothila:
@coderabbitai Thank you for the thorough cross-check. On the weight description, I believe the alignment should go in the other direction: the gateway-side text is the stale one.

On the current gateway head, the primary translation path applies weights. RestAPITransformer builds an endpoint for every target in a definition, carrying each weight, into the RuntimeDeployConfig (gateway-controller/pkg/transform/restapi.go, the endpoints loop), and the translator emits weighted clusters via createWeightedCluster (gateway-controller/pkg/xds/translator.go), which sets LoadBalancingWeight for weights greater than zero. TranslateConfigs uses this transformer path first for the RestApi kind; the legacy path that reads only the first target runs only as a fallback. We also verified this empirically on a gateway built from PR #2012's head: a per-operation ref to a pool weighted 80/20 distributed 30 requests exactly 24/6 across the two targets.

The "Reserved for future multi-target load balancing; not applied yet" description in the gateway schema predates createWeightedCluster and is not part of PR #2012's diff, so this PR's wording reflects current behavior, and the gateway description should be refreshed in a separate follow-up instead. "Only the first target is used" remains true only for the legacy fallback path and the API-level slot cluster, which is tracked as a known follow-up item.

I will fix the leftover comment reference to the removed sentinel in api_upstream_integration_test.go, thank you for catching it.

mehara-rothila:
@coderabbitai full review

Please review thoroughly, focusing on:
1. Gateway parity: does platform-api validation exactly mirror the gateway for the upstreamDefinitions pool and per-op refs (name pattern ^[a-zA-Z0-9\-_]+$ max 100, weight 0 to 100, connect-timeout single-unit format, basePath regex, API-level url-or-ref union)? Flag anything the platform accepts that the gateway would 400 at deploy, or vice versa.
2. Reusable-upstream naming: is the rename complete and consistent (ReusableUpstreamsAPIToModel/ModelToAPI, ReusableUpstreamYAML, newAPIOperationUpstreamRef)? Confirm retained names are correct and unchanged: the UpstreamDefinitions struct field, the upstreamDefinitions/upstreams wire keys, api.ReusableUpstream, and the pre-existing api.UpstreamDefinition endpoint type. No YAML/JSON tag changes.
3. Round-trip integrity: create to model to JSON persistence to GET, and BuildAPIDeploymentYAML emission of API-level refs, per-op refs, the pool, timeout, every target url and weight (including an explicit weight:0 preserved via the *int field).
4. Merged-update safety: an update removing a pool entry still referenced by a per-op ref must be rejected (no dangling refs).
5. Backward compatibility: existing APIs without upstreamDefinitions/per-op upstream must be unaffected, and old stored JSON blobs must read back unchanged.
6. Test coverage adequacy for all the above.

Review only the changes in this PR, and do not report pre-existing issues in code outside the diff.

mehara-rothila:
@coderabbitai Following up on the timeout.connect: "0s" point, I want to verify what the gateway actually does with it end to end, not only at the management validator. Please trace it in the gateway-controller code (PR #2012 / main) and answer precisely with function names and file:line, based only on the code:

1. Take one upstreamDefinition with timeout.connect: "0s" and follow it from validation through to the built Envoy cluster config. List every place the connect timeout is parsed or checked, and state exactly what each one does with "0s" (accept, reject, or transform).
2. At the stage that builds the actual deploy/xDS cluster config for that definition, which function parses the definition's timeout.connect, and does it accept a zero/non-positive value or return an error? Quote that function.
3. Is the connect timeout parsed by the same code path as the route/request ("resilience") timeout, or a different one? For each, what does "0s" mean and is it accepted?
4. What constraint does Envoy place on a cluster connect_timeout, and is 0 a valid value there?
5. Putting 1 to 4 together: does a gateway deployment of a definition with timeout.connect: "0s" actually succeed, or does it fail at some stage? If the platform rejects "0s" at create and update, is that matching the gateway's deploy-time behavior or diverging from it?

Please follow it all the way to the deployed cluster config rather than stopping at the validator, and report whatever the code actually shows.

platform-api/internal/dto/api.go
```diff
@@ -137,15 +137,16 @@ type Vhosts struct {
...
-	Policies          []Policy               `yaml:"policies,omitempty"`
-	Operations        []api.OperationRequest `yaml:"operations,omitempty"`
-	Channels          []api.ChannelRequest   `yaml:"channels,omitempty"`
+	DisplayName         string                 `yaml:"displayName"`
+	Version             string                 `yaml:"version"`
+	Context             string                 `yaml:"context"`
+	SubscriptionPlans   []string               `yaml:"subscriptionPlans,omitempty"`
+	Vhosts              *Vhosts                `yaml:"vhosts,omitempty"`
+	Upstream            *UpstreamYAML          `yaml:"upstream,omitempty"`
+	UpstreamDefinitions []ReusableUpstreamYAML `yaml:"upstreamDefinitions,omitempty"`
```
ashera96: Shall we rename this `ReusableUpstreamYAML` to something more aligning like `[]UpstreamDefinition`? Any limitation in doing a rename like that?

platform-api/internal/dto/api.go
```diff
@@ -160,10 +161,28 @@ type UpstreamTarget struct {
 	Ref string `yaml:"ref,omitempty"`
 }
 
+// ReusableUpstreamYAML represents a named entry in the deployment YAML's upstreamDefinitions pool.
+type ReusableUpstreamYAML struct {
+	Name      string                `yaml:"name"`
+	BasePath  string                `yaml:"basePath,omitempty"`
+	Timeout   *UpstreamTimeoutYAML  `yaml:"timeout,omitempty"`
```
ashera96: Do we need the `YAML` part in this type? Can we maybe use just `UpstreamTimeout`?

platform-api/internal/model/api.go
```diff
@@ -43,14 +43,15 @@ type API struct {
...
-	Upstream          UpstreamConfig `json:"upstream,omitempty"`
-	Policies          []Policy       `json:"policies,omitempty"`
-	Operations        []Operation    `json:"operations,omitempty"`
-	SubscriptionPlans []string       `json:"subscriptionPlans,omitempty"`
+	Name                string             `json:"name,omitempty"`
+	Version             string             `json:"version,omitempty"`
+	Context             *string            `json:"context,omitempty"`
+	Transport           []string           `json:"transport,omitempty"`
+	Upstream            UpstreamConfig     `json:"upstream,omitempty"`
+	UpstreamDefinitions []ReusableUpstream `json:"upstreamDefinitions,omitempty"`
```
ashera96: If above was renamed, check here as well.

mehara-rothila:
@coderabbitai review

platform-api/internal/dto/api.go
mehara-rothila: `UpstreamDefinition` is already a pre-existing platform-api schema/type, used for the API-level `upstream.main`/`sandbox` endpoint (the `url | ref | auth` union). It predates this PR:
https://github.com/wso2/api-platform/blob/be3e3a00bffaa638d18569cde826d92d45669d88/platform-api/resources/openapi.yaml#L6703-L6719

That is why the new pool entry is `ReusableUpstream`: renaming it to `UpstreamDefinition` would collide with that existing type in both the OpenAPI components and the generated Go.

For context, the gateway's `management-openapi.yaml` uses the opposite terminology, `Upstream` for the endpoint and `UpstreamDefinition` for the pool entry, so the two planes already name these concepts differently, independent of this PR.

platform-api/internal/dto/api.go
mehara-rothila: Done in dfbfcd7c3. Dropped the suffix (`ReusableUpstreamYAML`/`UpstreamBackendYAML`/`UpstreamTimeoutYAML` to `ReusableUpstream`/`UpstreamBackend`/`UpstreamTimeout`), which also lines these DTO types up with the unsuffixed names already used in the model and generated `api` packages. Field tags are unchanged, so the emitted deployment YAML is identical.

platform-api/resources/openapi.yaml
```diff
@@ -6731,6 +6737,89 @@ components:
...
+                example: 80
+
+    UpstreamTimeout:
+      title: Upstream Timeout
+      type: object
+      description: Timeout configuration for upstream requests
+      properties:
+        connect:
+          type: string
+          description: Connection timeout duration (e.g., "5s", "500ms")
```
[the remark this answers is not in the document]
mehara-rothila: Done in e667035b8. Updated the `connect` description in the OpenAPI schema (and the matching generated doc comment) to state that the value must be positive (non-zero), aligning the published contract with the runtime validation.

platform-api/internal/utils/api.go
```diff
@@ -318,6 +321,43 @@ func (u *APIUtil) upstreamAuthToModel(auth *api.UpstreamAuth) *model.UpstreamAut
...
+	}
+	models := make([]model.ReusableUpstream, 0, len(*definitions))
+	for _, definition := range *definitions {
+		m := model.ReusableUpstream{
+			Name:     definition.Name,
+			BasePath: defaultStringPtr(definition.BasePath),
+		}
+		if definition.Timeout != nil && definition.Timeout.Connect != nil {
+			m.Timeout = &model.UpstreamTimeout{Connect: *definition.Timeout.Connect}
+		}
```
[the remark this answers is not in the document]
mehara-rothila: Good catch. Fixed in e667035b8: the API to model conversion now trims `connect` and treats a blank or whitespace-only value as unset, matching how the validation already treats it, so it is no longer persisted as a meaningless value in the deployment YAML. Added a unit test covering both the blank and padded cases.

mehara-rothila:
@coderabbitai review

platform-api/internal/model/api.go
mehara-rothila: The model is already `ReusableUpstream`, and dfbfcd7c3 renames the DTO to match, so both layers are aligned now. Kept as `ReusableUpstream` rather than `UpstreamDefinition` for the reason in the thread above.

mehara-rothila:
@coderabbitai review

platform-api/internal/dto/api.go
mehara-rothila: Renamed as suggested. `ReusableUpstream` is now `NamedUpstream` across the OpenAPI spec, the generated types, and the model, DTO and converter layers. Field names and the wire format are unchanged. Done in ba2840023.

ashera96 (approved):

PR #2185, opened 2026-06-13
implement multi platform api support

tharindu1st:
implement multi platform api support


tharindu1st:
@coderabbitai review

platform-api/src/internal/service/event_dispatcher.go
```diff
@@ -0,0 +1,196 @@
...
+
+// OnGatewayConnected is called whenever a WebSocket connection for gatewayID is registered
+// on this instance. On the first connection, it subscribes to the EventHub for that gateway
+// and starts a dispatch goroutine. Subsequent connections only increment the ref count.
+func (d *EventDispatcher) OnGatewayConnected(gatewayID string) {
+	d.mu.Lock()
+	defer d.mu.Unlock()
+
+	d.refCount[gatewayID]++
+	if d.refCount[gatewayID] > 1 {
```
VirajSalaka: Do we really need to restrict this one GatewayID? May be we can restrict at 2 at least.

platform-api/src/internal/service/event_dispatcher.go
```diff
@@ -0,0 +1,196 @@
...
+		return // already subscribed
+	}
+
+	if err := d.hub.RegisterGateway(gatewayID); err != nil {
+		d.logger.Error("EventDispatcher: failed to register gateway", "gatewayID", gatewayID, "error", err)
```
VirajSalaka, line 87: Should we return here?

platform-api/src/internal/service/event_dispatcher.go
VirajSalaka: Anyway when there are multiple platform-apis running, the two gateway controller replicas distribute among the two and counter would be shared anyway. 
As future work we need to have proper connection limiting here.

platform-api/src/internal/service/event_dispatcher.go
```diff
@@ -0,0 +1,196 @@
...
+	sub, ok := d.subs[gatewayID]
+	if !ok {
+		return
+	}
+
+	if err := d.hub.Unsubscribe(gatewayID, sub.ch); err != nil {
+		d.logger.Warn("EventDispatcher: failed to unsubscribe from gateway", "gatewayID", gatewayID, "error", err)
+	}
+	close(sub.done)
+	delete(d.subs, gatewayID)
```
VirajSalaka, line 134: Depending on the decision to include more than 1 connection, the logic here would change too.

platform-api/src/internal/service/gateway_events.go
```diff
@@ -372,115 +267,43 @@ func (s *GatewayEventsService) broadcastEventWithUserID(gatewayID, userId, event
...
+		EventID:             correlationID,
+		EventData:           string(eventJSON),
 	}
 
-	s.slogger.Debug("Broadcast summary", "gatewayID", gatewayID, "correlationId", correlationID, "type", eventType, "total", len(connections), "success", successCount, "failed", failureCount)
-
-	if successCount == 0 {
-		return fmt.Errorf("failed to deliver %s event to any connection: %w", eventType, lastError)
+	if err := s.hub.PublishEvent(gatewayID, hubEvent); err != nil {
+		s.slogger.Error("Failed to publish gateway event",
```
VirajSalaka, line 285: When there is no gateway is registered at that point, this would return an error. May be we can accept that as we do not want to keep unnecessary data in events table. But change the log line to INFO or WARN.

We can improve the underlying eventhub implementation as well.

VirajSalaka (dismissed):

VirajSalaka (approved):

Krishanx92:
Fix: https://github.com/wso2/api-platform/issues/2266

PR #2205, opened 2026-06-16
Add Agent SKILL for OpenAPI Editing and defining scopes

Thushani-Jayasekera:
$subject


.agents/skills/platform-api-scopes/SKILL.md
```diff
@@ -0,0 +1,166 @@
+---
+name: platform-api-scopes
```
renuka-fernando: Shall we make this generic, not only for the platform API?

renuka-fernando (approved):

PR #2228, opened 2026-06-19
Secret management for Platform API

### LINKED ISSUE #2149: [Feature]: Secret Management Feature for Platform API

Area/Other (Anything else which does not match above categories)

### Suggested Feature

Need to implement a secret management feature for Platform API.

This should cover the following;

- Simple secure vault implementation which encrypts and stores the secrets for on-prem
- Pluggable secret management solution to support multiple vendors like HarshiCorp vault in cloud

The following points will be considered when implementing the solution.

- There will be a Go interface to access secret manager - Get/Add/Update/Remove secret operations
- This interface will support both on-prem and cloud solutions (Support for multiple secret providers)
- Need to track the secret type - File based secrets (certs), String based secrets (api keys, passwords) - This will be useful when doing cert management feature to fetch what are the certs expiring in a given date.
- Need to add a field to track expiry - Will be useful when storing certs, JWTs, etc
- Need to consider organisation and projects (For platform GW, it has a flat hierarchy and store the secrets only with the GW Id) - Add organisation and project fields when defining the DB schema for Platform API
- Secret names are defined by end users. When a secret is given by a user we need to resolve the secret value considering the organisation and project of the user

### Related Issues

https://github.com/wso2/api-platform/issues/837

npamudika:
## Purpose
Fix https://github.com/wso2/api-platform/issues/2149

Introduces an end-to-end **Secrets Management** feature for the WSO2 API Platform. Currently, sensitive credentials (API keys, tokens, certificates) are stored in plain text inside artifact configuration blobs. This creates security risks: credentials are exposed in the database, audit logs, and API responses. There is also no lifecycle management (rotation, deprecation, delete protection) and no way for gateway controllers to efficiently sync only the secrets they need.

## Goals
- Provide a secure encrypted secret store with AES-GCM-256 encryption via a pluggable vault (`InHouseVault`)
- Expose CRUD REST endpoints (`POST/GET/PUT/DELETE /api/v1/secrets`) for org-scoped secret management
- Allow artifact configurations (REST APIs, LLM providers, LLM proxies, MCP servers) to reference secrets via `{{ secret "handle" }}` placeholders instead of plain-text credentials
- Block deletion of secrets that are actively referenced by any artifact (deployed or undeployed)
- Provide an internal gateway sync endpoint (`GET /api/internal/v1/secrets`) that returns only the secrets referenced by a specific gateway's deployed artifacts, with optional decrypted values for bulk startup sync

## Approach
**Storage**: A new `secrets` table stores `ciphertext` (AES-GCM-256 BLOB), a `sha256:` hash for change detection, `value_scope` (always `ORG_SHARED`), `status` (`ACTIVE`/`DEPRECATED`), and `provider` (`IN_BUILT`). Secrets are never physically deleted — `DELETE` sets `status = DEPRECATED`.

**Reference tracking**: A new `artifact_secret_refs` table with composite PK `(organization_id, artifact_uuid, secret_handle, gateway_id)` tracks every `{{ secret "..." }}` placeholder across all artifact types. Two row types coexist:
- `gateway_id = ''` — artifact-level rows written on artifact create/update (used for delete protection — blocks deletion even of undeployed drafts)
- `gateway_id = <uuid>` — deployment-snapshot rows written atomically inside the `SetCurrentWithDetails` transaction on deploy; cleared on undeploy/archive

**Delete protection**: `FindRefs` queries `artifact_secret_refs` by `(org_id, secret_handle)` — a single indexed lookup replacing the previous `LIKE`-based scans across multiple tables.

**Gateway sync**: `GetSecretHandlesByGateway` queries `artifact_secret_refs WHERE org = ? AND gateway_id = ?` (indexed). `ListByHandles` filters to `value_scope = 'ORG_SHARED'` only. The `?includeValues=true` parameter enables startup bulk fetch (decrypts all secrets server-side), avoiding N round trips for `/value` calls.

**Validation**: `ValidateSecretRefs` is called at artifact create/update time; rejects configs that reference non-existent or DEPRECATED secret handles.

See https://docs.google.com/document/d/1I-ls5QUvDpGLmYPe39ai_i3AEL_awv8vQbi83jOnoRg/edit?tab=t.w3lefdjelsk5#heading=h.qxwb86l2rlwz for the full feature spec including database DDL, API shapes, flow diagrams, and gateway sync sequence.

## User stories
- **As a platform operator**, I can store an API key as a secret once and reference it by handle in multiple artifact configs, so credentials are never stored in plain text.
- **As a platform operator**, I can rotate a secret's value via `PUT /api/v1/secrets/{handle}` and all gateways will automatically pick up the new value on the next sync cycle (detected via hash change).
- **As a platform operator**, I am prevented from deleting a secret that is still referenced by any artifact config (deployed or draft), with a clear 409 response listing the referencing artifacts.
- **As a gateway controller**, I can call `GET /api/internal/v1/secrets` with my API key and receive only the secrets referenced by my deployed artifacts, with decrypted values available for bulk startup sync.

## Documentation
https://docs.google.com/document/d/1I-ls5QUvDpGLmYPe39ai_i3AEL_awv8vQbi83jOnoRg/edit?tab=t.w3lefdjelsk5#heading=h.qxwb86l2rlwz — full feature specification including API contracts, database schema, encryption design, gateway sync protocol, and flow diagrams.

### Unit tests
- **`src/internal/repository/secret_test.go`** — 26 tests covering `SecretRepo` CRUD, `List` pagination, `updatedAfter` filter, `ListByHandles` with scope filter, `FindRefs` (artifact-level and gateway-level rows), `extractSecretHandles` (regex, deduplication), `upsertArtifactSecretRefs` (insert on create, replace on update, clear when no secrets), `upsertDeploymentSecretRefs` (deploy, undeploy, isolation from artifact-level rows)
- **`src/internal/service/secret_service_test.go`** — 21 tests covering `Create` (value scope always `ORG_SHARED`, type defaulting, encryption failure propagation), `List` (pagination shape), `Get`/`Update`, `Delete` (blocked by artifact-level ref, blocked by deployment ref, success, soft-delete contract), `ValidateSecretRefs` (missing handle, deduplication), `Decrypt` (DEPRECATED rejection)
- **`src/internal/service/llm_secret_validation_test.go`** — 4 tests covering LLM provider create/update with valid and invalid `{{ secret "..." }}` placeholders

### Integration tests
- **`src/internal/handler/secret_integration_test.go`** — 19 handler-level tests using real SQLite + AES-GCM vault + Gin router. Covers the full public API (`/api/v1/secrets`): create (201, 409 duplicate, 401 no auth), list (pagination object shape, `?limit`/`?offset`, `?updatedAfter` filter, invalid timestamp → 400), get (200 metadata-only, 404 not found), update (200 new value returned), delete (204 unreferenced, 409 with references array, 404 not found), cross-org isolation (org B sees empty list, org B gets 404 for org A secret), soft-delete verification (DB row retained with `status=DEPRECATED`, `Decrypt` returns "secret is deprecated"), and `ValidateSecretRefs` rejection of DEPRECATED handle references.
- **`src/internal/handler/gateway_secret_integration_test.go`** — 18 handler-level tests for the internal GW API (`/api/internal/v1/secrets`). Covers: authentication (no key → 401, invalid key → 401), list scoping (only gateway's deployed secrets returned, `?updatedAfter` filter, invalid timestamp → 400, no `value` field by default), multi-gateway isolation (GW-A secret not visible to GW-B, shared handle visible to both), undeploy removes secrets from list, deduplication of same handle across multiple artifacts, `GET /:id/value` (200 plaintext, 200 after rotation, 404 not found), `?includeValues=true` bulk fetch (ACTIVE secrets include `value`, DEPRECATED secrets omit `value`, empty list when no deployments, `hash` present alongside `value`).

**Test scenario traceability**: https://docs.google.com/spreadsheets/d/1FfXxjNbNh1SgjgxabPG8VcKLkITfqQTKwH61VVmfDkk/edit?gid=750717062#gid=750717062 — 85 test cases with `Test Result` and `Automated Test` columns. **64 PASS** (automated), **21 MANUAL** (gateway controller sync behaviour, process startup, and AI Workspace frontend orchestration flows).

## Security checks
- Ran FindSecurityBugs plugin and verified report? **yes**

Notes:
- Secrets are encrypted with AES-GCM-256 before persistence; the `ciphertext` column stores binary blobs only — plaintext is never written to the database
- The encryption key (`PLATFORM_SECRET_ENCRYPTION_KEY`) is injected via environment variable; an ephemeral key is auto-generated (with a warning log) if unset — operators must set a persistent key for production
- Secret values are returned in API responses only at creation time (`POST`) and rotation time (`PUT`); all `GET` responses omit the `value` field
- The internal GW endpoint is authenticated via gateway API key (SHA-256 hashed token lookup), not JWT — it is not accessible via the public API path

## Samples
N/A — this is a platform infrastructure feature with no standalone samples.

## Related PRs
N/A

## Test environment
- **OS**: macOS 14 (Darwin 24.6.0)
- **Go**: 1.23
- **Database**: SQLite (in-memory, for tests); PostgreSQL-compatible schema also provided (`schema.postgres.sql`)
- **Tested via**: `go test ./src/internal/...` — all packages pass

### DIFF
--- a/gateway/gateway-controller/pkg/utils/api_utils.go
+++ b/gateway/gateway-controller/pkg/utils/api_utils.go
@@ -1174,9 +1174,9 @@ func (s *APIUtilsService) CheckArtifactsExist(artifactIDs []string) ([]string, e
 // PlatformSecretMeta holds the metadata returned by GET /api/internal/v1/secrets.
 // Value is non-nil only when the request included ?includeValues=true (startup bulk fetch).
 type PlatformSecretMeta struct {
-	ID          string  `json:"id"`
-	Handle      string  `json:"name"`
-	DisplayName string  `json:"displayName"`
+	ID          string  `json:"uuid"`
+	Handle      string  `json:"handle"`
+	DisplayName string  `json:"name"`
 	Hash        string  `json:"hash"`
 	Status      string  `json:"status"`
 	Value       *string `json:"value,omitempty"`
@@ -1241,13 +1241,13 @@ func (s *APIUtilsService) FetchPlatformSecrets(updatedAfter *time.Time, includeV
 }
 
 // FetchPlatformSecretValue fetches the decrypted plaintext value of a secret from the
-// Platform API internal endpoint GET /internal/v1/secrets/:id/value.
+// Platform API internal endpoint GET /internal/v1/secrets/:handle/value.
 // The Gateway authenticates using the same api-key header used for all Platform API calls.
-func (s *APIUtilsService) FetchPlatformSecretValue(secretID string) (string, error) {
-	valueURL := s.getBaseURL() + "/secrets/" + secretID + "/value"
+func (s *APIUtilsService) FetchPlatformSecretValue(secretHandle string) (string, error) {
+	valueURL := s.getBaseURL() + "/secrets/" + secretHandle + "/value"
 
 	s.logger.Debug("Fetching platform secret value",
-		slog.String("secret_id", secretID),
+		slog.String("secret_handle", secretHandle),
 		slog.String("url", valueURL),
 	)
 
--- a/gateway/gateway-controller/pkg/utils/api_utils.go
+++ b/gateway/gateway-controller/pkg/utils/api_utils.go
@@ -1170,3 +1170,114 @@ func (s *APIUtilsService) CheckArtifactsExist(artifactIDs []string) ([]string, e
 
 	return existingIDs, nil
 }
+
+// PlatformSecretMeta holds the metadata returned by GET /api/internal/v1/secrets.
+// Value is non-nil only when the request included ?includeValues=true (startup bulk fetch).
+type PlatformSecretMeta struct {
+	ID          string  `json:"id"`
+	Handle      string  `json:"name"`
+	DisplayName string  `json:"displayName"`
+	Hash        string  `json:"hash"`
+	Status      string  `json:"status"`
+	Value       *string `json:"value,omitempty"`
+}
+
+type platformSecretsListResponse struct {
+	List  []PlatformSecretMeta `json:"list"`
+	Count int                  `json:"count"`
+}
+
+type platformSecretValueResponse struct {
+	Value string `json:"value"`
+}
+
+// FetchPlatformSecrets retrieves secrets from the Platform API internal endpoint.
+// If updatedAfter is non-nil, only secrets modified after that time are returned.
+// If includeValues is true, decrypted plaintext values are included in the response
+// (used on startup for a single bulk fetch instead of N per-secret round trips).
+func (s *APIUtilsService) FetchPlatformSecrets(updatedAfter *time.Time, includeValues bool) ([]PlatformSecretMeta, error) {
+	secretsURL := s.getBaseURL() + "/secrets"
+
+	params := url.Values{}
+	if updatedAfter != nil {
+		params.Set("updatedAfter", updatedAfter.UTC().Format(time.RFC3339))
+	}
+	if includeValues {
+		params.Set("includeValues", "true")
+	}
+	if len(params) > 0 {
+		secretsURL += "?" + params.Encode()
+	}
+
+	s.logger.Info("Fetching platform secrets",
+		slog.String("url", secretsURL),
+		slog.Bool("includeValues", includeValues),
+	)
+
+	req, err := http.NewRequest(http.MethodGet, secretsURL, nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create secrets request: %w", err)
+	}
+	req.Header.Add("api-key", s.config.Token)
+	req.Header.Add("Accept", "application/json")
+
+	resp, err := s.client.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch platform secrets: %w", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		bodyBytes, _ := io.ReadAll(resp.Body)
+		return nil, fmt.Errorf("platform secrets request failed with status %d: %s", resp.StatusCode, string(bodyBytes))
+	}
+
+	var listResp platformSecretsListResponse
+	if err := json.NewDecoder(resp.Body).Decode(&listResp); err != nil {
+		return nil, fmt.Errorf("failed to decode platform secrets response: %w", err)
+	}
+
+	s.logger.Info("Successfully fetched platform secrets",
+		slog.Int("count", len(listResp.List)),
+		slog.Bool("includeValues", includeValues),
+	)
+
+	return listResp.List, nil
+}
+
+// FetchPlatformSecretValue fetches the decrypted plaintext value of a secret from the
+// Platform API internal endpoint GET /internal/v1/secrets/:id/value.
+// The Gateway authenticates using the same api-key header used for all Platform API calls.
+func (s *APIUtilsService) FetchPlatformSecretValue(secretID string) (string, error) {
+	valueURL := s.getBaseURL() + "/secrets/" + secretID + "/value"
+
+	s.logger.Debug("Fetching platform secret value",
+		slog.String("secret_id", secretID),
+		slog.String("url", valueURL),
+	)
+
+	req, err := http.NewRequest(http.MethodGet, valueURL, nil)
+	if err != nil {
+		return "", fmt.Errorf("failed to create secret value request: %w", err)
+	}
+	req.Header.Add("api-key", s.config.Token)
+	req.Header.Add("Accept", "application/json")
+
+	resp, err := s.client.Do(req)
+	if err != nil {
+		return "", fmt.Errorf("failed to fetch platform secret value: %w", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		bodyBytes, _ := io.ReadAll(resp.Body)
+		return "", fmt.Errorf("platform secret value request failed with status %d: %s", resp.StatusCode, string(bodyBytes))
+	}
+
+	var valueResp platformSecretValueResponse
+	if err := json.NewDecoder(resp.Body).Decode(&valueResp); err != nil {
+		return "", fmt.Errorf("failed to decode platform secret value response: %w", err)
+	}
+
+	return valueResp.Value, nil
+}
--- a/gateway/gateway-controller/pkg/utils/api_utils.go
+++ b/gateway/gateway-controller/pkg/utils/api_utils.go
@@ -1237,11 +1237,6 @@ func (s *APIUtilsService) FetchPlatformSecrets(updatedAfter *time.Time, includeV
 		return nil, fmt.Errorf("failed to decode platform secrets response: %w", err)
 	}
 
-	s.logger.Info("Successfully fetched platform secrets",
-		slog.Int("count", len(listResp.List)),
-		slog.Bool("includeValues", includeValues),
-	)
-
 	return listResp.List, nil
 }
 
--- a/platform-api/src/config/config.go
+++ b/platform-api/src/config/config.go
@@ -23,6 +23,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"log/slog"
+	"os"
 	"reflect"
 	"strings"
 	"sync"
@@ -305,12 +306,21 @@ func LoadConfig(configPath string) (*Server, error) {
 	}
 
 	if cfg.Database.SecretEncryptionKey == "" {
+		demoMode := strings.ToLower(strings.TrimSpace(os.Getenv("APIP_DEMO_MODE")))
+		if demoMode != "true" && demoMode != "1" {
+			return nil, fmt.Errorf("database.secret_encryption_key is not set. " +
+				"All replicas must share a stable key (PLATFORM_SECRET_ENCRYPTION_KEY or database.secret_encryption_key). " +
+				"Generate one with: openssl rand -hex 32. " +
+				"To allow an ephemeral key in a single-node dev environment, set APIP_DEMO_MODE=true")
+		}
 		key, err := generateRandomSecret()
 		if err != nil {
 			return nil, fmt.Errorf("failed to generate secret encryption key: %w", err)
 		}
 		cfg.Database.SecretEncryptionKey = key
-		slog.Warn("database.secret_encryption_key is not set — generated an ephemeral random key; all encrypted secrets will be unreadable on restart")
+		slog.Warn("APIP_DEMO_MODE: database.secret_encryption_key is not set — using an ephemeral random key. " +
+			"Encrypted secrets will be unreadable after restart and WILL NOT be shared across replicas. " +
+			"Set PLATFORM_SECRET_ENCRYPTION_KEY for any persistent or multi-replica deployment.")
 	}
 
 	return cfg, nil
--- a/platform-api/src/config/config.go
+++ b/platform-api/src/config/config.go
@@ -89,8 +89,7 @@ type Server struct {
 	Gateway          Gateway          `koanf:"gateway"`
 	EventHub         EventHub         `koanf:"event_hub"`
 
-	EnableScopeValidation   bool `koanf:"enable_scope_validation"`
-	OrgCreationRequiresAuth bool `koanf:"org_creation_requires_auth"`
+	EnableScopeValidation bool `koanf:"enable_scope_validation"`
 }
 
 // Auth groups all authentication-related configuration.
--- a/platform-api/src/config/config.go
+++ b/platform-api/src/config/config.go
@@ -178,8 +178,7 @@ type Database struct {
 
 	EncryptionKey                  string `koanf:"encryption_key"`
 	SubscriptionTokenEncryptionKey string `koanf:"subscription_token_encryption_key"`
-	SecretEncryptionKey            string `koanf:"secret_encryption_key"`
-	SecretVaultProvider            string `koanf:"secret_vault_provider"`
+	SecretEncryptionKey string `koanf:"secret_encryption_key"`
 }
 
 // DefaultDevPortal holds default DevPortal configuration for new organizations.
@@ -305,11 +304,14 @@ func LoadConfig(configPath string) (*Server, error) {
 		slog.Warn("auth.jwt.secret_key is not set — generated an ephemeral random key; all sessions will be invalidated on restart")
 	}
 
-	if cfg.Database.SecretEncryptionKey == "" {
+	// SecretEncryptionKey is optional when the shared DATABASE_ENCRYPTION_KEY is configured;
+	// server.go resolves the final key via: SecretEncryptionKey → EncryptionKey → JWT secret.
+	// Only fail (or warn in demo mode) when no key source is available at all.
+	if cfg.Database.SecretEncryptionKey == "" && cfg.Database.EncryptionKey == "" {
 		demoMode := strings.ToLower(strings.TrimSpace(os.Getenv("APIP_DEMO_MODE")))
 		if demoMode != "true" && demoMode != "1" {
-			return nil, fmt.Errorf("database.secret_encryption_key is not set. " +
-				"All replicas must share a stable key (PLATFORM_SECRET_ENCRYPTION_KEY or database.secret_encryption_key). " +
+			return nil, fmt.Errorf("no encryption key configured for secrets management. " +
+				"Set PLATFORM_SECRET_ENCRYPTION_KEY (secret-specific) or DATABASE_ENCRYPTION_KEY (shared). " +
 				"Generate one with: openssl rand -hex 32. " +
 				"To allow an ephemeral key in a single-node dev environment, set APIP_DEMO_MODE=true")
 		}
@@ -318,9 +320,9 @@ func LoadConfig(configPath string) (*Server, error) {
 			return nil, fmt.Errorf("failed to generate secret encryption key: %w", err)
 		}
 		cfg.Database.SecretEncryptionKey = key
-		slog.Warn("APIP_DEMO_MODE: database.secret_encryption_key is not set — using an ephemeral random key. " +
+		slog.Warn("APIP_DEMO_MODE: no encryption key configured for secrets — using an ephemeral random key. " +
 			"Encrypted secrets will be unreadable after restart and WILL NOT be shared across replicas. " +
-			"Set PLATFORM_SECRET_ENCRYPTION_KEY for any persistent or multi-replica deployment.")
+			"Set PLATFORM_SECRET_ENCRYPTION_KEY or DATABASE_ENCRYPTION_KEY for any persistent or multi-replica deployment.")
 	}
 
 	return cfg, nil
@@ -385,8 +387,6 @@ func envToKoanfKey(s string) string {
 		return "database.subscription_token_encryption_key"
 	case "platform_secret_encryption_key":
 		return "database.secret_encryption_key"
-	case "platform_secret_vault_provider":
-		return "database.secret_vault_provider"
 
 	// Auth
 	case "auth_skip_paths":
--- a/platform-api/src/config/config.go
+++ b/platform-api/src/config/config.go
@@ -89,7 +89,8 @@ type Server struct {
 	Gateway          Gateway          `koanf:"gateway"`
 	EventHub         EventHub         `koanf:"event_hub"`
 
-	EnableScopeValidation bool `koanf:"enable_scope_validation"`
+	EnableScopeValidation   bool `koanf:"enable_scope_validation"`
+	OrgCreationRequiresAuth bool `koanf:"org_creation_requires_auth"`
 }
 
 // Auth groups all authentication-related configuration.
--- a/platform-api/src/config/config.go
+++ b/platform-api/src/config/config.go
@@ -177,6 +177,8 @@ type Database struct {
 
 	EncryptionKey                  string `koanf:"encryption_key"`
 	SubscriptionTokenEncryptionKey string `koanf:"subscription_token_encryption_key"`
+	SecretEncryptionKey            string `koanf:"secret_encryption_key"`
+	SecretVaultProvider            string `koanf:"secret_vault_provider"`
 }
 
 // DefaultDevPortal holds default DevPortal configuration for new organizations.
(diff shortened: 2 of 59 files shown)


platform-api/src/internal/handler/secret.go
```diff
@@ -0,0 +1,257 @@
...
+		t, err := time.Parse(time.RFC3339, ua)
+		if err != nil {
+			c.JSON(http.StatusBadRequest, utils.NewErrorResponse(400, "Bad Request", "updatedAfter must be an RFC3339 timestamp"))
+			return
+		}
+		updatedAfter = &t
+	}
+
+	resp, err := h.secretService.List(orgID, limit, offset, updatedAfter)
+	if err != nil {
```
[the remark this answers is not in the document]
npamudika: We will do the projectId related implementations later. Hence this is not required now.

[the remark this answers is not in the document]
npamudika: No, I'm tracking it in my end.

platform-api/src/internal/dto/secret.go
```diff
@@ -0,0 +1,101 @@
...
+ */
+
+package dto
+
+import "time"
+
+// CreateSecretRequest is the request body for POST /api/v1/secrets.
+// Accepts multipart/form-data to support file-based secret values in future.
+type CreateSecretRequest struct {
+	Handle      string `form:"handle" binding:"required"`
```
AnuGayan, line 25: 🟠 **Major — OpenAPI contract does not match this DTO / the real client**

**Issue:** These request DTOs carry `form:` tags only, the real client (`portals/ai-workspace/src/apis/secretApis.ts`) posts `multipart/form-data`, and the integration tests post multipart too — but `openapi.yaml` documents `application/json` bodies with properties `name` (described as *the handle*), `displayName`, etc. The spec also never documents the required `handle` field, lists `type` values (`API_KEY`, `PRIVATE_KEY`) the service rejects, and documents `projectId`/`environment` that the handler ignores.

**Reason:** Any consumer generating a client from the published spec will fail: content-type mismatch, `name` won't bind to `DisplayName` (JSON binding ignores `form:` tags), and `handle` (required) will be missing → 400. The spec's `name`=handle vs the form's `handle`+`name`(=displayName) are swapped.

**Verify:** "Send `POST /api/v1/secrets` with `Content-Type: application/json` and body `{\"name\":\"k\",\"displayName\":\"K\",\"value\":\"v\",\"type\":\"API_KEY\"}` exactly as the OpenAPI example shows. Does it succeed, or 400 on missing `handle` / invalid type?"

**Suggested fix:** Make the spec authoritative and align the DTO — preferably switch to JSON with `json:` tags (matches the rest of the codebase, e.g. `dto/api-project.go`): `Handle string \`json:\"handle\" binding:\"required\"\``, etc. Then fix the spec field names/`type` enum and drop (or implement) `projectId`/`environment`.

platform-api/src/internal/service/secret_service.go
```diff
@@ -0,0 +1,245 @@
...
+	if err != nil {
+		return "", err
+	}
+	if secret.Status == model.SecretStatusDeprecated {
+		return "", errors.New("secret is deprecated")
+	}
+	return s.vault.Decrypt(context.Background(), secret.Ciphertext)
+}
+
+func hashSecret(plaintext string) string {
```
AnuGayan: 🟠 **Major (security) — unsalted SHA-256 of the plaintext is stored and exposed**

**Issue:** `hashSecret` is a bare `sha256(plaintext)`. This `hash` is persisted in `secrets.hash` and returned in `SecretSummary.Hash` (list/get) and `SecretSyncItem.Hash` (gateway sync).

**Reason:** For low-entropy secret values (passwords, short/structured API keys), an unsalted SHA-256 enables offline dictionary/brute-force recovery by anyone who can read a list response. For a feature whose purpose is to keep credentials out of DB/logs/API responses, publishing a plaintext-derived digest undercuts the goal.

**Verify:** "The `hash` field returned by `GET /api/v1/secrets` is `sha256(value)`. If an attacker reads that response, can they recover a weak secret value offline by hashing a wordlist?"

**Suggested fix:** Use a keyed digest for change detection instead — `hmac.New(sha256.New, changeDetectionKey)` over the plaintext. It still changes iff the value changes but is not brute-forceable without the server-side key.

platform-api/src/config/config.go
```diff
@@ -302,6 +304,15 @@ func LoadConfig(configPath string) (*Server, error) {
 		slog.Warn("auth.jwt.secret_key is not set — generated an ephemeral random key; all sessions will be invalidated on restart")
 	}
 
+	if cfg.Database.SecretEncryptionKey == "" {
```
AnuGayan: 🟠 **Major — ephemeral key fallback is unsafe under horizontal scaling / restart**

**Issue:** When `database.secret_encryption_key` is unset, a random key is generated per process and the server starts with only a `Warn`.

**Reason:** The control plane is horizontally scaled. Each replica generates a *different* key, so ciphertext written by replica A can't be decrypted by replica B → gateway secret sync and `GET /:id/value` fail non-deterministically. On a single instance, every restart silently renders all stored secrets undecryptable (data loss).

**Verify:** "Start two platform-api replicas with `secret_encryption_key` unset, create a secret on replica A, then fetch its value through replica B. Does decryption succeed?"

**Suggested fix:** Fail closed — refuse to start without an explicit key unless an explicit dev/insecure flag is set. At minimum, document that the key MUST be identical across all replicas and escalate the log to error.

platform-api/src/internal/handler/gateway_internal.go
```diff
@@ -805,6 +809,109 @@ func (h *GatewayInternalAPIHandler) CheckArtifactsExist(c *gin.Context) {
...
+			DisplayName: s.DisplayName,
+			Type:        s.Type,
+			Provider:    s.Provider,
+			Status:      s.Status,
+			Hash:        s.Hash,
+			CreatedAt:   s.CreatedAt,
+			UpdatedAt:   s.UpdatedAt,
+		}
+		if includeValues {
+			plaintext, err := h.secretService.Decrypt(orgID, s.Handle)
```
AnuGayan: 🟡 **Minor — one DEPRECATED secret 500s the whole bulk sync; also an N+1**

**Issue:** In the `includeValues=true` loop, `Decrypt` of a DEPRECATED secret returns an error → the entire request 500s. The test `TestGatewaySecretHandler_IncludeValues_DeprecatedSecretFails` is titled *"item has no 'value' field"* but actually asserts 500 — title/PR-doc and behavior disagree. Separately, each `Decrypt` re-runs `GetByHandle`, a DB round-trip, even though the loop already holds the `model.Secret` rows.

**Reason:** Retrying is deterministic-fail (the secret stays deprecated), so a gateway can wedge on startup sync. And the per-item refetch adds N queries the bulk endpoint was meant to avoid.

**Verify:** "If any secret referenced by a gateway's deployment is DEPRECATED, does `GET /api/internal/v1/secrets?includeValues=true` return 200 with that item's value omitted, or 500 for the whole batch?"

**Suggested fix:** Skip the value for non-ACTIVE secrets (omit `value`, keep the item) and continue; decrypt `s.Ciphertext` directly instead of calling `Decrypt`→`GetByHandle`. Then reconcile the test name + PR description.

platform-api/src/internal/handler/secret.go
```diff
@@ -0,0 +1,264 @@
...
+		return
+	}
+
+	c.Status(http.StatusNoContent)
+}
+
+// GetSecretValue is an internal endpoint for the GW controller only.
+// Returns the decrypted plaintext value of a secret by handle.
+// Must be protected by GW service account JWT at the middleware level.
+func (h *SecretHandler) GetSecretValue(c *gin.Context) {
```
AnuGayan: 🟡 **Minor — dead code that would leak plaintext if ever wired in**

**Issue:** `GetSecretValue` decrypts and returns a secret's plaintext, but it is **not** registered in `RegisterRoutes`. The gateway value path is served by `gateway_internal.go:GetGatewaySecretValue` (auth'd + deployment-scoped).

**Reason:** Its own comment says it "Must be protected … at the middleware level." If someone later mounts it under `/api/v1/secrets/:id/value`, any org user with a JWT could read decrypted secret values — a latent foot-gun in a security-sensitive file.

**Verify:** "Is `SecretHandler.GetSecretValue` referenced anywhere outside tests? If not, what stops it from being mounted under the JWT-protected `/api/v1` group later?"

**Suggested fix:** Remove the method; the internal handler already covers this with proper scoping.

platform-api/src/internal/service/secret_service.go
```diff
@@ -0,0 +1,245 @@
...
+
+func (s *SecretService) Get(orgID, handle string) (*dto.SecretSummary, error) {
+	secret, err := s.repo.GetByHandle(orgID, handle)
+	if err != nil {
+		return nil, err
+	}
+	return secretToSummary(secret), nil
+}
+
+func (s *SecretService) Update(orgID, handle, updatedBy string, req *dto.UpdateSecretRequest) (*dto.SecretResponse, error) {
```
AnuGayan, line 132: 🟡 **Minor — `Update` has no status guard**

**Issue:** `Update` re-encrypts and writes new ciphertext/hash without checking `status`, so a DEPRECATED secret can be rotated while staying DEPRECATED.

**Reason:** Leaves an odd "deprecated but freshly rotated" row whose lifecycle is ambiguous (it can't be referenced by new artifacts but holds a fresh value).

**Verify:** "Soft-delete a secret, then `PUT /api/v1/secrets/{handle}` with a new value. What status does the row have afterward, and is that intended?"

**Suggested fix:** Either reject updates to DEPRECATED secrets (`409`/`404`) or reactivate on rotation — whichever matches the intended lifecycle — and add a test.

platform-api/src/internal/repository/artifact_refs.go
```diff
@@ -0,0 +1,90 @@
...
+	"regexp"
+
+	"platform-api/src/internal/database"
+)
+
+// artifactLevelGatewayID is the sentinel gateway_id used for artifact-level (non-deployed) refs.
+// These rows represent the current artifact config and are used for delete protection.
+const artifactLevelGatewayID = ""
+
+var secretPlaceholderRe = regexp.MustCompile(`\{\{\s*secret\s+\\?"([^"\\]+)\\?"\s*\}\}`)
```
AnuGayan: 🔵 **Nit — duplicated placeholder regex across packages**

**Issue:** This `secretPlaceholderRe` is byte-for-byte identical to `secretPlaceholderRegex` in `service/secret_service.go:35`. One is used for ref-extraction, the other for validation.

**Reason:** If the two ever drift, a placeholder could be validated but not tracked (or vice versa), silently breaking delete-protection or gateway sync.

**Verify:** "Are the validation regex and the ref-extraction regex guaranteed to match the exact same set of placeholders? What enforces that?"

**Suggested fix:** Extract a single exported regex/helper to one shared package and use it from both call sites.

AnuGayan (commented):
Thorough review of the secrets feature. The data model (composite PKs, the artifact-level vs deployment-level ref split, indexes, FK cascades), the AES-GCM vault, and the test coverage are strong. I've left inline comments for the items worth addressing before merge. Each is formatted as **Issue / Reason / Verify / Suggested fix**.

Headline items: (1) the OpenAPI contract does not match the implementation for the public Create/Update endpoints, (2) an unsalted SHA-256 of the plaintext is stored and returned, and (3) the ephemeral-key fallback is unsafe under horizontal scaling. Overall: 🔧 request changes (security feature — the security/contract items matter), but nothing structurally wrong with the design.

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -481,3 +481,56 @@ CREATE TABLE IF NOT EXISTS events (
 );
 
 CREATE INDEX IF NOT EXISTS idx_events_gateway_id_processed_timestamp ON events(gateway_id, processed_timestamp);
+-- Secrets table for encrypted secret management
+CREATE TABLE IF NOT EXISTS secrets (
+    uuid            VARCHAR(40)  PRIMARY KEY,
+    organization_id VARCHAR(40)  NOT NULL,
```
Thushani-Jayasekera: shall we rename to `organization_uuid` maintain consistency with other tables.
Please apply the above in `artifact_secret_refs` table as well.

portals/ai-workspace/docker-compose.yaml
```diff
@@ -46,12 +47,18 @@ services:
     container_name: platform-api
     restart: unless-stopped
     command: ["-config", "/etc/platform-api/config-platform-api.toml"]
+    # entrypoint: ["sleep", "infinity"]
```
malinthaprasan: Can be removed?

platform-api/src/internal/database/schema.sql
```diff
@@ -479,3 +479,57 @@ CREATE TABLE IF NOT EXISTS events (
...
+    description     VARCHAR(1023),
+    ciphertext      BYTEA        NOT NULL,
+    hash            VARCHAR(80)  NOT NULL,
+    type            VARCHAR(20)  NOT NULL DEFAULT 'GENERIC', -- GENERIC | CERTIFICATE
+    provider        VARCHAR(20)  NOT NULL DEFAULT 'IN_BUILT',
+    status          VARCHAR(20)  NOT NULL DEFAULT 'ACTIVE',
+    created_at      TIMESTAMP    DEFAULT CURRENT_TIMESTAMP,
+    created_by      VARCHAR(255),
+    updated_at      TIMESTAMP    DEFAULT CURRENT_TIMESTAMP,
+    updated_by      VARCHAR(255),
```
malinthaprasan: In a latest discussion, we discussed to move the audit details into a seperate table. Please have a chat with @Thushani-Jayasekera

platform-api/src/internal/model/secret.go
```diff
@@ -0,0 +1,70 @@
...
+const (
+	SecretTypeCertificate = "CERTIFICATE"
+	SecretTypeGeneric     = "GENERIC"
+
+	SecretProviderInHouse        = "IN_BUILT"
+	SecretProviderAWSKMS         = "AWS_KMS"
+	SecretProviderHashiCorpVault = "HASHICORP_VAULT"
+
+	SecretStatusActive     = "ACTIVE"
+	SecretStatusDeprecated = "DEPRECATED"
```
malinthaprasan, line 31: When does a secret become DEPRECATED?

platform-api/src/resources/openapi.yaml
```diff
@@ -11094,6 +11315,181 @@ components:
...
+          type: string
+          example: wso2-openai-key
+        displayName:
+          type: string
+          example: WSO2 OpenAI API Key
+        description:
+          type: string
+        type:
+          type: string
+          enum: [API_KEY, CERTIFICATE, PRIVATE_KEY, GENERIC]
```
malinthaprasan: Only the types defined in the code needs to be reflected here.

platform-api/src/resources/openapi.yaml
```diff
@@ -11094,6 +11315,181 @@ components:
...
+        value:
+          type: string
+          description: Plaintext value — only present in create and rotate responses
+          example: sk-xxx
+        type:
+          type: string
+          enum: [API_KEY, CERTIFICATE, PRIVATE_KEY, GENERIC]
+        provider:
+          type: string
+          enum: [IN_BUILT, AWS_KMS, HASHICORP_VAULT]
```
malinthaprasan: Lets do not advertise AWS_KMS, HASHICORP_VAULT to the OpenAPI as we do not support them right now. We can keep the IN_BUILT only.

platform-api/src/resources/openapi.yaml
```diff
@@ -6538,6 +6538,227 @@ paths:
         '500':
           $ref: '#/components/responses/InternalServerError'
 
+  /secrets:
+    post:
+      summary: Create a secret
+      description: |
+        Create a new encrypted secret scoped to the organization. The plaintext value is
+        returned **once only** in this response and never again in any subsequent read.
```
malinthaprasan: `returned **once only**` - please check and update if its not valid.

Can you run a quick validation to the OpenAPI changes? It looks like it doesn't reflect some of the latest changes

platform-api/src/internal/model/secret.go
npamudika: When a secret is soft deleted, the status will be DEPRECATED

platform-api/src/resources/openapi.yaml
npamudika: Fixed

platform-api/src/resources/openapi.yaml
npamudika: Fixed

platform-api/src/resources/openapi.yaml
npamudika: Fixed.

platform-api/src/internal/database/schema.sql
npamudika: Checked on this and we are keeping audit details in the same table as well. So keeping this as it is.

npamudika:
@copilot resolve the merge conflicts in this pull request

platform-api/src/cmd/cmd
renuka-fernando, line 1: We don't need to commit this fille, this should be the platform API binary.

npamudika:
@coderabbitai Please approve

gateway/gateway-controller/pkg/utils/api_utils.go
```diff
@@ -1170,3 +1170,114 @@ func (s *APIUtilsService) CheckArtifactsExist(artifactIDs []string) ([]string, e
...
+	}
+
+	var listResp platformSecretsListResponse
+	if err := json.NewDecoder(resp.Body).Decode(&listResp); err != nil {
+		return nil, fmt.Errorf("failed to decode platform secrets response: %w", err)
+	}
+
+	s.logger.Info("Successfully fetched platform secrets",
+		slog.Int("count", len(listResp.List)),
+		slog.Bool("includeValues", includeValues),
```
Krishanx92: Check this values are necessary for logger as info, if not lets remove

platform-api/src/config/default_config.go
```diff
@@ -29,6 +29,7 @@ func defaultConfig() *Server {
 		OpenAPISpecPath:            "./resources/openapi.yaml",
 		LLMTemplateDefinitionsPath: "./resources/default-llm-provider-templates",
 		EnableScopeValidation:      true,
+		OrgCreationRequiresAuth:    false,
```
Krishanx92: WHy we need this?

npamudika: Seems this was added from a rebase, will remove it.

npamudika:
Fixes related to https://github.com/wso2/api-platform/issues/2149

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -496,3 +496,56 @@ CREATE TABLE IF NOT EXISTS events (
 );
 
 CREATE INDEX IF NOT EXISTS idx_events_gateway_id_processed_timestamp ON events(gateway_id, processed_timestamp);
+-- Secrets table for encrypted secret management
+CREATE TABLE IF NOT EXISTS secrets (
+    uuid              VARCHAR(40)  PRIMARY KEY,
+    organization_uuid VARCHAR(40)  NOT NULL,
+    handle            VARCHAR(100) NOT NULL,
```
Thushani-Jayasekera: @renuka-fernando should we change this to VARCHAR(40) according to the refactoring we did in my PR?

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -496,3 +496,56 @@ CREATE TABLE IF NOT EXISTS events (
...
+
+CREATE INDEX IF NOT EXISTS idx_secret_scopes_scope ON secret_scopes(scope, scope_value);
+
+-- Pre-computed secret handle references per deployed artifact per gateway.
+-- Populated at deploy time (status→DEPLOYED) and cleared at undeploy time.
+-- Eliminates the 6-table JOIN + application-level regex on every GW secret sync.
+CREATE TABLE IF NOT EXISTS artifact_secret_refs (
+    organization_uuid VARCHAR(40)  NOT NULL,
+    artifact_uuid     VARCHAR(40)  NOT NULL,
+    secret_handle     VARCHAR(100) NOT NULL,
```
Thushani-Jayasekera: [as above](https://github.com/wso2/api-platform/pull/2228/changes#r3479254195)

platform-api/src/internal/database/schema.postgres.sql
renuka-fernando: yes let's go with VARCHAR(40)

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -496,3 +496,56 @@ CREATE TABLE IF NOT EXISTS events (
...
 CREATE INDEX IF NOT EXISTS idx_events_gateway_id_processed_timestamp ON events(gateway_id, processed_timestamp);
+-- Secrets table for encrypted secret management
+CREATE TABLE IF NOT EXISTS secrets (
+    uuid              VARCHAR(40)  PRIMARY KEY,
+    organization_uuid VARCHAR(40)  NOT NULL,
+    handle            VARCHAR(100) NOT NULL,
+    name              VARCHAR(255) NOT NULL,
+    description       VARCHAR(1023),
+    ciphertext        BYTEA        NOT NULL,
+    hash              VARCHAR(80)  NOT NULL,
```
Thushani-Jayasekera: VARCHAR(255)

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -496,3 +496,56 @@ CREATE TABLE IF NOT EXISTS events (
...
+    type              VARCHAR(20)  NOT NULL DEFAULT 'GENERIC',
+    provider          VARCHAR(20)  NOT NULL DEFAULT 'IN_BUILT',
+    status            VARCHAR(20)  NOT NULL DEFAULT 'ACTIVE',
+    created_at        TIMESTAMP    DEFAULT CURRENT_TIMESTAMP,
+    created_by        VARCHAR(255),
+    updated_at        TIMESTAMP    DEFAULT CURRENT_TIMESTAMP,
+    updated_by        VARCHAR(255),
```
Thushani-Jayasekera, line 572: Add 
`data_version VARCHAR(20) NOT NULL DEFAULT '1.0',`

no need to handle this at code level

renuka-fernando (approved):

PR #2254, opened 2026-06-23
[Platform API] Refactor database schema and improve

Thushani-Jayasekera:
Issue: https://github.com/wso2/api-platform/issues/2279

https://claude.ai/code/artifact/a73f74e1-c55a-4804-89e4-46cf8f3c708f


platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -59,16 +64,10 @@ CREATE TABLE IF NOT EXISTS applications (
 -- Artifacts table
 CREATE TABLE IF NOT EXISTS artifacts (
     uuid VARCHAR(40) PRIMARY KEY,
-    handle VARCHAR(255) NOT NULL,
-    name VARCHAR(255) NOT NULL,
-    version VARCHAR(30) NOT NULL,
     kind VARCHAR(20) NOT NULL,
+    kind_version VARCHAR(20) NOT NULL,
```
Thushani-Jayasekera: remove

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -77,14 +76,21 @@ CREATE TABLE IF NOT EXISTS artifacts (
 -- REST APIs table
 CREATE TABLE IF NOT EXISTS rest_apis (
     uuid VARCHAR(40) PRIMARY KEY,
-    description VARCHAR(1023),
-    created_by VARCHAR(200),
+    handle VARCHAR(255) NOT NULL,
+    name VARCHAR(255) NOT NULL,
+    version VARCHAR(30) NOT NULL,
```
Thushani-Jayasekera: org_id,
unique(org_id, handle)

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -110,67 +118,70 @@ CREATE TABLE IF NOT EXISTS subscription_plans (
...
     subscription_token VARCHAR(512) NOT NULL,
     subscription_token_hash VARCHAR(64) NOT NULL,
     subscription_plan_uuid VARCHAR(40),
     organization_uuid VARCHAR(40) NOT NULL,
     status VARCHAR(20) NOT NULL DEFAULT 'ACTIVE',
+    created_by VARCHAR(200),
     created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
+    updated_by VARCHAR(200),
     updated_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
-    FOREIGN KEY (api_uuid) REFERENCES rest_apis(uuid) ON DELETE CASCADE,
```
Thushani-Jayasekera, line 128: artifact_uuid

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -110,67 +118,70 @@ CREATE TABLE IF NOT EXISTS subscription_plans (
...
 
 -- Gateway Custom Policies table (org-scoped custom policies synced from gateway manifests)
 CREATE TABLE IF NOT EXISTS gateway_custom_policies (
     uuid VARCHAR(40) PRIMARY KEY,
     organization_uuid VARCHAR(40) NOT NULL,
-    name VARCHAR(255) NOT NULL,
-    display_name VARCHAR(255),
-    version VARCHAR(15) NOT NULL,
+    handle VARCHAR(255) NOT NULL,
+    name VARCHAR(255),
```
Thushani-Jayasekera: no need to change as custom policies have a name/ display name

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -110,67 +118,70 @@ CREATE TABLE IF NOT EXISTS subscription_plans (
...
+    version VARCHAR(30) NOT NULL,
     description TEXT,
```
Thushani-Jayasekera: description VARCHAR(1023)

platform-api/src/internal/database/schema.postgres.sql
Thushani-Jayasekera: + all 4 audit fields

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -227,159 +241,135 @@ CREATE TABLE IF NOT EXISTS deployment_status (
     status_desired VARCHAR(20),
     performed_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
     status_reason VARCHAR(50),
+    created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
```
Thushani-Jayasekera: remove

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -227,159 +241,135 @@ CREATE TABLE IF NOT EXISTS deployment_status (
     status_desired VARCHAR(20),
     performed_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
```
Thushani-Jayasekera: add performed_by

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -227,159 +241,135 @@ CREATE TABLE IF NOT EXISTS deployment_status (
...
-    configuration TEXT NOT NULL,
+    configuration JSONB NOT NULL,
+    created_by VARCHAR(200),
     created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
+    updated_by VARCHAR(200),
     updated_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
+    organization_uuid VARCHAR(40) NOT NULL,
     FOREIGN KEY (organization_uuid) REFERENCES organizations(uuid) ON DELETE CASCADE,
-    UNIQUE(organization_uuid, handle)
+    UNIQUE(organization_uuid, handle, version)
```
Thushani-Jayasekera: remove version

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -227,159 +241,135 @@ CREATE TABLE IF NOT EXISTS deployment_status (
...
+    name VARCHAR(255) NOT NULL,
+    version VARCHAR(30) NOT NULL,
     project_uuid VARCHAR(40) NOT NULL,
     description VARCHAR(1023),
-    created_by VARCHAR(255),
     provider_uuid VARCHAR(40) NOT NULL,
-    openapi_spec TEXT,
-    status VARCHAR(20) NOT NULL DEFAULT 'CREATED',
+    openapi_spec JSONB,
+    status VARCHAR(20) NOT NULL DEFAULT 'pending',
```
Thushani-Jayasekera: remove as no logic is implemented around this

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -227,159 +241,135 @@ CREATE TABLE IF NOT EXISTS deployment_status (
...
     configuration JSONB NOT NULL,
+    created_by VARCHAR(200),
+    created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
+    updated_by VARCHAR(200),
+    updated_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
     FOREIGN KEY (uuid) REFERENCES artifacts(uuid) ON DELETE CASCADE,
     FOREIGN KEY (project_uuid) REFERENCES projects(uuid) ON DELETE CASCADE,
-    FOREIGN KEY (provider_uuid) REFERENCES llm_providers(uuid) ON DELETE RESTRICT
+    FOREIGN KEY (provider_uuid) REFERENCES llm_providers(uuid) ON DELETE RESTRICT,
+    CHECK (status IN ('pending', 'deployed', 'failed'))
```
Thushani-Jayasekera: remove

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -389,23 +379,27 @@ CREATE TABLE IF NOT EXISTS api_keys (
...
+    CHECK (status IN ('active', 'expired', 'revoked'))
 );
 
 -- Application API Key mappings table
 CREATE TABLE IF NOT EXISTS application_api_keys (
     application_uuid VARCHAR(40) NOT NULL,
     api_key_id VARCHAR(40) NOT NULL,
+    created_by VARCHAR(200),
     created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
+    updated_by VARCHAR(200),
```
Thushani-Jayasekera: remove?

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -416,7 +410,9 @@ CREATE TABLE IF NOT EXISTS application_api_keys (
 CREATE TABLE IF NOT EXISTS application_artifacts (
     application_uuid VARCHAR(40) NOT NULL,
     artifact_uuid VARCHAR(40) NOT NULL,
+    created_by VARCHAR(200),
     created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
+    updated_by VARCHAR(200),
     updated_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
```
Thushani-Jayasekera: remove?

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -475,9 +478,21 @@ CREATE TABLE IF NOT EXISTS events (
...
     PRIMARY KEY (gateway_id, event_id),
     FOREIGN KEY (gateway_id) REFERENCES gateway_states(gateway_id) ON DELETE CASCADE
 );
 
 CREATE INDEX IF NOT EXISTS idx_events_gateway_id_processed_timestamp ON events(gateway_id, processed_timestamp);
+CREATE INDEX IF NOT EXISTS idx_events_entity ON events(entity_type, entity_id);
+
+CREATE TABLE IF NOT EXISTS audit (
+   uuid VARCHAR(40) PRIMARY KEY,
+   action VARCHAR(50) NOT NULL CHECK(action IN ('CREATE', 'UPDATE', 'DELETE')),
```
Thushani-Jayasekera: remove check

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -475,9 +478,21 @@ CREATE TABLE IF NOT EXISTS events (
...
+   resource_uuid VARCHAR(40) NOT NULL,
+   resource_type VARCHAR(50),
+   organization_uuid VARCHAR(40) NOT NULL,
+   updated_by VARCHAR(200),
+   updated_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
```
Thushani-Jayasekera: performed_by
performed_at

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -226,160 +256,144 @@ CREATE TABLE IF NOT EXISTS deployment_status (
...
+    name VARCHAR(255) NOT NULL,
+    version VARCHAR(30) NOT NULL,
     description VARCHAR(1023),
-    created_by VARCHAR(255),
     template_uuid VARCHAR(40) NOT NULL,
-    openapi_spec TEXT,
+    openapi_spec JSONB,
     model_list TEXT,
-    status VARCHAR(20) NOT NULL DEFAULT 'CREATED',
     configuration JSONB NOT NULL,
```
Thushani-Jayasekera: as above

.agents/skills/db-schema-review/SKILL.md
```diff
@@ -0,0 +1,445 @@
...
+name: db-schema-review
+description: |
+  Design, change, and review relational database schemas. Use proactively whenever:
+  - Adding a new table
+  - Modifying an existing table (new column, type change, constraint, index)
+  - Reviewing schema changes before a PR
+  - Writing or evaluating a migration plan
+  - Asking "is this table well designed?" or "what indexes does this table need?"
+
+  Applies house conventions: UUID primary keys (VARCHAR(40)), handle/name/version identity triple for named resources, org-scoping, JSONB for JSON columns (Postgres), audit columns (created_by/at, updated_by/at), CHECK constraints on status fields, and standard indexing patterns.
```
Thushani-Jayasekera: "JSONB for JSON columns" --> 
If you're deciding between TEXT and JSONB:
- Use JSONB when the column stores JSON that you want to validate, query, filter, or index.
- Use TEXT when it's just an opaque blob/string and PostgreSQL never needs to understand its structure.
- Migrating later is possible, but converting large tables from TEXT → JSONB may require data cleanup and can lock/rewrite the table depending on PostgreSQL version and migration approach.

Thushani-Jayasekera: Concept | Postgres | MySQL | Oracle
-- | -- | -- | --
JSON | JSONB | JSON | CLOB + validation
Binary | BYTEA | BLOB | BLOB
Large text | TEXT | TEXT/LONGTEXT | CLOB
String | TEXT/VARCHAR | VARCHAR/TEXT | VARCHAR2

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -32,7 +34,9 @@ CREATE TABLE IF NOT EXISTS projects (
     name VARCHAR(255) NOT NULL,
     organization_uuid VARCHAR(40) NOT NULL,
     description VARCHAR(1023),
+    created_by VARCHAR(200),
```
Thushani-Jayasekera, line 40: username claim?

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -59,32 +64,33 @@ CREATE TABLE IF NOT EXISTS applications (
...
     project_uuid VARCHAR(40) NOT NULL,
-    lifecycle_status VARCHAR(20) DEFAULT 'CREATED',
-    transport VARCHAR(255), -- JSON array as TEXT
+    description VARCHAR(1023),
+    lifecycle_status VARCHAR(20) NOT NULL DEFAULT 'CREATED',
     configuration JSONB NOT NULL,
+    created_by VARCHAR(200),
+    created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
+    updated_by VARCHAR(200),
+    updated_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
```
Thushani-Jayasekera: + origin VARCHAR(20) NOT NULL DEFAULT 'control_plane', (new feature)

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -59,32 +64,33 @@ CREATE TABLE IF NOT EXISTS applications (
...
     uuid VARCHAR(40) PRIMARY KEY,
-    description VARCHAR(1023),
-    created_by VARCHAR(200),
+    organization_uuid VARCHAR(40) NOT NULL,
+    handle VARCHAR(255) NOT NULL,
+    name VARCHAR(255) NOT NULL,
+    version VARCHAR(30) NOT NULL,
     project_uuid VARCHAR(40) NOT NULL,
-    lifecycle_status VARCHAR(20) DEFAULT 'CREATED',
-    transport VARCHAR(255), -- JSON array as TEXT
```
Thushani-Jayasekera, line 84: no need to have it in configuration json.

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -59,32 +64,33 @@ CREATE TABLE IF NOT EXISTS applications (
 -- Artifacts table
 CREATE TABLE IF NOT EXISTS artifacts (
     uuid VARCHAR(40) PRIMARY KEY,
-    handle VARCHAR(255) NOT NULL,
-    name VARCHAR(255) NOT NULL,
-    version VARCHAR(30) NOT NULL,
     kind VARCHAR(20) NOT NULL,
```
Thushani-Jayasekera: rename to `type`

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -59,32 +64,33 @@ CREATE TABLE IF NOT EXISTS applications (
...
+    organization_uuid VARCHAR(40) NOT NULL,
+    handle VARCHAR(255) NOT NULL,
+    name VARCHAR(255) NOT NULL,
+    version VARCHAR(30) NOT NULL,
     project_uuid VARCHAR(40) NOT NULL,
-    lifecycle_status VARCHAR(20) DEFAULT 'CREATED',
-    transport VARCHAR(255), -- JSON array as TEXT
+    description VARCHAR(1023),
+    lifecycle_status VARCHAR(20) NOT NULL DEFAULT 'CREATED',
     configuration JSONB NOT NULL,
```
Thushani-Jayasekera: Add `data_version` to all the tables representing data models..., (VARCHAR type - maintain . versions , default v1.0)
-- platform api's configuration version...

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -226,160 +256,144 @@ CREATE TABLE IF NOT EXISTS deployment_status (
...
+    CHECK (status_desired IN ('DEPLOYED', 'UNDEPLOYED'))
 );
 
 -- LLM Provider Templates table
 CREATE TABLE IF NOT EXISTS llm_provider_templates (
     uuid VARCHAR(40) PRIMARY KEY,
-    organization_uuid VARCHAR(40) NOT NULL,
     handle VARCHAR(255) NOT NULL,
-    name VARCHAR(253) NOT NULL,
+    group_version_id VARCHAR(255) NOT NULL, -- RENAME to group_version_id
```
Thushani-Jayasekera: `group_id`

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -474,25 +502,44 @@ CREATE INDEX IF NOT EXISTS idx_application_api_keys_app_id ON application_api_ke
...
     originated_timestamp TIMESTAMPTZ NOT NULL,
-    entity_type TEXT NOT NULL,
-    action TEXT NOT NULL CHECK(action IN ('CREATE', 'UPDATE', 'DELETE')),
-    entity_id TEXT NOT NULL,
-    event_id TEXT NOT NULL,
+    entity_type VARCHAR(255) NOT NULL,
+    action VARCHAR(20) NOT NULL,
+    entity_id VARCHAR(255) NOT NULL,
+    event_id VARCHAR(64) NOT NULL,
     event_data TEXT NOT NULL,
```
Thushani-Jayasekera, line 534: @tharindu1st , to be consistent with the gateway controller db, i kept this as TEXT/ NVARCHAR(MAX).

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -21,8 +21,11 @@ CREATE TABLE IF NOT EXISTS organizations (
     handle VARCHAR(255) UNIQUE NOT NULL,
     name VARCHAR(255) NOT NULL,
     region VARCHAR(63) NOT NULL,
-    created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
-    updated_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP
+    data_version VARCHAR(20) NOT NULL DEFAULT 'v1.0',
```
renuka-fernando:
```suggestion
    data_version VARCHAR(20) NOT NULL DEFAULT '1.0',
```

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -21,8 +21,11 @@ CREATE TABLE IF NOT EXISTS organizations (
     handle VARCHAR(255) UNIQUE NOT NULL,
```
renuka-fernando:
```suggestion
    handle VARCHAR(40) UNIQUE NOT NULL,
```

renuka-fernando: Add this to the skill as well

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -32,8 +35,11 @@ CREATE TABLE IF NOT EXISTS projects (
...
     description VARCHAR(1023),
-    created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
-    updated_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
+    data_version VARCHAR(20) NOT NULL DEFAULT 'v1.0',
+    created_by VARCHAR(200),
+    created_at TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP,
+    updated_by VARCHAR(200),
+    updated_at TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP,
     FOREIGN KEY (organization_uuid) REFERENCES organizations(uuid) ON DELETE CASCADE,
     UNIQUE(name, organization_uuid)
```
renuka-fernando:
```suggestion
    UNIQUE(organization_uuid, handle)
```

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -32,8 +35,11 @@ CREATE TABLE IF NOT EXISTS projects (
     name VARCHAR(255) NOT NULL,
```
renuka-fernando, line 36: Don't we need a handle for project?

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -44,12 +50,14 @@ CREATE TABLE IF NOT EXISTS applications (
     handle VARCHAR(255) NOT NULL,
```
renuka-fernando:
```suggestion
    handle VARCHAR(40) NOT NULL,
```

platform-api/src/internal/database/schema.postgres.sql
renuka-fernando:
```suggestion
    UNIQUE(organization_uuid, handle)
```

platform-api/src/internal/database/schema.postgres.sql
Thushani-Jayasekera: cc: @renuka-fernando

platform-api/src/internal/database/schema.postgres.sql
Thushani-Jayasekera: This is available now.

platform-api/src/internal/database/schema.postgres.sql
Thushani-Jayasekera: Added

platform-api/src/internal/service/websub_api.go
```diff
@@ -156,6 +156,7 @@ func (s *WebSubAPIService) Create(orgUUID, createdBy string, req *api.WebSubAPI)
 		return nil, fmt.Errorf("failed to create WebSub API: %w", err)
 	}
 
+	_ = s.auditRepo.Record("CREATE", m.UUID, "websub_api", orgUUID, createdBy)
```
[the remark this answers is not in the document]
Thushani-Jayasekera: The finding is not valid. auditRepo is always constructed via repository.NewAuditRepo(db) on server startup (server.go:109) and passed as a required parameter — it cannot be nil. There are no nil guards on auditRepo.Record calls in any other service in the codebase (deployment, gateway, subscription, organization, etc.), so this is the established pattern. No fix needed.

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -18,148 +18,178 @@
...
     FOREIGN KEY (uuid) REFERENCES artifacts(uuid) ON DELETE CASCADE,
-    FOREIGN KEY (project_uuid) REFERENCES projects(uuid) ON DELETE CASCADE
+    FOREIGN KEY (organization_uuid) REFERENCES organizations(uuid) ON DELETE CASCADE,
+    FOREIGN KEY (project_uuid) REFERENCES projects(uuid) ON DELETE CASCADE,
+    UNIQUE(organization_uuid, handle)
 );
 
 -- Subscription plans table (organization-scoped rate/billing plans)
 CREATE TABLE IF NOT EXISTS subscription_plans (
     uuid VARCHAR(40) PRIMARY KEY,
```
renuka-fernando, line 102: Don't we need a handle?
cc @malinthaprasan, @tharindu1st

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -18,148 +18,178 @@
...
-    plan_name VARCHAR(40) NOT NULL,
+    plan_name VARCHAR(255) NOT NULL,
```
renuka-fernando: just name?

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -18,148 +18,178 @@
...
     status VARCHAR(20) NOT NULL DEFAULT 'ACTIVE',
-    created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
-    updated_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
+    data_version VARCHAR(20) NOT NULL DEFAULT '1.0',
+    created_by VARCHAR(200),
+    created_at TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP,
+    updated_by VARCHAR(200),
+    updated_at TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP,
     FOREIGN KEY (organization_uuid) REFERENCES organizations(uuid) ON DELETE CASCADE,
     UNIQUE(organization_uuid, plan_name),
```
renuka-fernando: UNIQUE(organization_uuid, handle),

platform-api/src/internal/database/schema.postgres.sql
Thushani-Jayasekera: also api_keys, does not have a field for name(display name).

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -18,148 +18,174 @@
...
-    FOREIGN KEY (api_uuid) REFERENCES rest_apis(uuid) ON DELETE CASCADE,
+    data_version VARCHAR(20) NOT NULL DEFAULT '1.0',
+    created_by VARCHAR(200),
+    created_at TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP,
+    updated_by VARCHAR(200),
+    updated_at TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP,
+    FOREIGN KEY (artifact_uuid) REFERENCES artifacts(uuid) ON DELETE CASCADE,
     FOREIGN KEY (organization_uuid) REFERENCES organizations(uuid) ON DELETE CASCADE,
     FOREIGN KEY (subscription_plan_uuid, organization_uuid)
       REFERENCES subscription_plans(uuid, organization_uuid) ON DELETE RESTRICT,
```
AnuGayan: 🔴 **Blocker — composite FK references a column set with no matching UNIQUE constraint**

**Problem:** `subscriptions` declares `FOREIGN KEY (subscription_plan_uuid, organization_uuid) REFERENCES subscription_plans(uuid, organization_uuid)`, but `subscription_plans` now only has `PRIMARY KEY(uuid)` and `UNIQUE(organization_uuid, handle)`. A composite FK requires the referenced columns to be covered by a PK/UNIQUE constraint — that constraint no longer exists.

**How it happened:** this refactor dropped `UNIQUE(uuid, organization_uuid)` from `subscription_plans`. The sibling `artifacts` table kept its equivalent (renamed to `UNIQUE(organization_uuid, uuid)`), which is why the `subscriptions → artifacts` composite FK still works — but the `subscription_plans` one was not preserved.

**Impact (verified):** PostgreSQL & SQL Server fail at `CREATE TABLE subscriptions` (*"there is no unique constraint matching given keys for referenced table subscription_plans"*) — the schema cannot bootstrap. On SQLite, inserts into `subscriptions` that reference a plan fail with `foreign key mismatch` (reproduced locally).

**How to fix:** re-add `UNIQUE(organization_uuid, uuid)` to `subscription_plans` in all four schema files (order-independent for the FK match), or drop this composite FK and keep only the single-column `subscription_plan_uuid` FK with app-level org-consistency checks.

platform-api/src/internal/database/schema.sql
```diff
@@ -18,144 +18,170 @@
...
     status VARCHAR(20) NOT NULL DEFAULT 'ACTIVE',
+    data_version VARCHAR(20) NOT NULL DEFAULT '1.0',
+    created_by VARCHAR(200),
     created_at DATETIME DEFAULT CURRENT_TIMESTAMP,
+    updated_by VARCHAR(200),
     updated_at DATETIME DEFAULT CURRENT_TIMESTAMP,
-    FOREIGN KEY (api_uuid) REFERENCES rest_apis(uuid) ON DELETE CASCADE,
     FOREIGN KEY (organization_uuid) REFERENCES organizations(uuid) ON DELETE CASCADE,
     FOREIGN KEY (subscription_plan_uuid, organization_uuid)
       REFERENCES subscription_plans(uuid, organization_uuid) ON DELETE RESTRICT,
```
AnuGayan: 🔴 **Blocker — composite FK references a column set with no matching UNIQUE constraint**

**Problem:** `subscriptions` declares `FOREIGN KEY (subscription_plan_uuid, organization_uuid) REFERENCES subscription_plans(uuid, organization_uuid)`, but `subscription_plans` now only has `PRIMARY KEY(uuid)` and `UNIQUE(organization_uuid, handle)`. A composite FK requires the referenced columns to be covered by a PK/UNIQUE constraint — that constraint no longer exists.

**How it happened:** this refactor dropped `UNIQUE(uuid, organization_uuid)` from `subscription_plans`. The sibling `artifacts` table kept its equivalent (renamed to `UNIQUE(organization_uuid, uuid)`), which is why the `subscriptions → artifacts` composite FK still works — but the `subscription_plans` one was not preserved.

**Impact (verified):** PostgreSQL & SQL Server fail at `CREATE TABLE subscriptions` (*"there is no unique constraint matching given keys for referenced table subscription_plans"*) — the schema cannot bootstrap. On SQLite, inserts into `subscriptions` that reference a plan fail with `foreign key mismatch` (reproduced locally).

**How to fix:** re-add `UNIQUE(organization_uuid, uuid)` to `subscription_plans` in all four schema files (order-independent for the FK match), or drop this composite FK and keep only the single-column `subscription_plan_uuid` FK with app-level org-consistency checks.

platform-api/src/internal/database/schema.sqlite.sql
```diff
@@ -18,147 +18,174 @@
...
+    data_version VARCHAR(20) NOT NULL DEFAULT '1.0',
+    created_by VARCHAR(200),
     created_at DATETIME DEFAULT CURRENT_TIMESTAMP,
+    updated_by VARCHAR(200),
     updated_at DATETIME DEFAULT CURRENT_TIMESTAMP,
-    FOREIGN KEY (api_uuid) REFERENCES rest_apis(uuid) ON DELETE CASCADE,
+    FOREIGN KEY (artifact_uuid) REFERENCES artifacts(uuid) ON DELETE CASCADE,
     FOREIGN KEY (organization_uuid) REFERENCES organizations(uuid) ON DELETE CASCADE,
     FOREIGN KEY (subscription_plan_uuid, organization_uuid)
       REFERENCES subscription_plans(uuid, organization_uuid) ON DELETE RESTRICT,
```
AnuGayan: 🔴 **Blocker — composite FK references a column set with no matching UNIQUE constraint**

**Problem:** `subscriptions` declares `FOREIGN KEY (subscription_plan_uuid, organization_uuid) REFERENCES subscription_plans(uuid, organization_uuid)`, but `subscription_plans` now only has `PRIMARY KEY(uuid)` and `UNIQUE(organization_uuid, handle)`. A composite FK requires the referenced columns to be covered by a PK/UNIQUE constraint — that constraint no longer exists.

**How it happened:** this refactor dropped `UNIQUE(uuid, organization_uuid)` from `subscription_plans`. The sibling `artifacts` table kept its equivalent (renamed to `UNIQUE(organization_uuid, uuid)`), which is why the `subscriptions → artifacts` composite FK still works — but the `subscription_plans` one was not preserved.

**Impact (verified):** PostgreSQL & SQL Server fail at `CREATE TABLE subscriptions` (*"there is no unique constraint matching given keys for referenced table subscription_plans"*) — the schema cannot bootstrap. On SQLite, inserts into `subscriptions` that reference a plan fail with `foreign key mismatch` (reproduced locally).

**How to fix:** re-add `UNIQUE(organization_uuid, uuid)` to `subscription_plans` in all four schema files (order-independent for the FK match), or drop this composite FK and keep only the single-column `subscription_plan_uuid` FK with app-level org-consistency checks.

platform-api/src/internal/database/schema.sqlserver.sql
```diff
@@ -15,145 +15,158 @@
...
+    -- removed via the artifact_uuid -> artifacts CASCADE edge (which itself cascades
+    -- from projects/organizations), so cleanup behavior is preserved.
     FOREIGN KEY (organization_uuid) REFERENCES organizations(uuid) ON DELETE NO ACTION,
     FOREIGN KEY (subscription_plan_uuid, organization_uuid)
-    REFERENCES subscription_plans(uuid, organization_uuid) ON DELETE NO ACTION,
-    FOREIGN KEY (api_uuid, organization_uuid)
-      REFERENCES artifacts(uuid, organization_uuid) ON DELETE NO ACTION,
-    UNIQUE(api_uuid, subscription_token_hash),
-    UNIQUE(api_uuid, subscriber_id, organization_uuid),
+        REFERENCES subscription_plans(uuid, organization_uuid) ON DELETE NO ACTION,
```
AnuGayan: 🔴 **Blocker — composite FK references a column set with no matching UNIQUE constraint**

**Problem:** `subscriptions` declares `FOREIGN KEY (subscription_plan_uuid, organization_uuid) REFERENCES subscription_plans(uuid, organization_uuid)`, but `subscription_plans` now only has `PRIMARY KEY(uuid)` and `UNIQUE(organization_uuid, handle)`. A composite FK requires the referenced columns to be covered by a PK/UNIQUE constraint — that constraint no longer exists.

**How it happened:** this refactor dropped `UNIQUE(uuid, organization_uuid)` from `subscription_plans`. The sibling `artifacts` table kept its equivalent (renamed to `UNIQUE(organization_uuid, uuid)`), which is why the `subscriptions → artifacts` composite FK still works — but the `subscription_plans` one was not preserved.

**Impact (verified):** PostgreSQL & SQL Server fail at `CREATE TABLE subscriptions` (*"there is no unique constraint matching given keys for referenced table subscription_plans"*) — the schema cannot bootstrap. On SQLite, inserts into `subscriptions` that reference a plan fail with `foreign key mismatch` (reproduced locally).

**How to fix:** re-add `UNIQUE(organization_uuid, uuid)` to `subscription_plans` in all four schema files (order-independent for the FK match), or drop this composite FK and keep only the single-column `subscription_plan_uuid` FK with app-level org-consistency checks.

platform-api/src/internal/repository/application.go
```diff
@@ -116,20 +124,27 @@ func (r *ApplicationRepo) GetAssociationTargetByUUID(targetUUID, orgID string) (
...
+		INNER JOIN (
+			SELECT uuid, handle, name, version, created_at, updated_at FROM rest_apis
+			UNION ALL SELECT uuid, handle, name, version, created_at, updated_at FROM websub_apis
+			UNION ALL SELECT uuid, handle, name, version, created_at, updated_at FROM webbroker_apis
+			UNION ALL SELECT uuid, handle, name, version, created_at, updated_at FROM llm_providers
+			UNION ALL SELECT uuid, handle, name, version, created_at, updated_at FROM llm_proxies
+			UNION ALL SELECT uuid, handle, name, version, created_at, updated_at FROM mcp_proxies
+		) src ON src.uuid = a.uuid
+		WHERE a.organization_uuid = ? AND (a.uuid = ? OR src.handle = ?)
+		LIMIT 1
```
AnuGayan: 🔴 **Blocker — raw `LIMIT 1` is not portable to SQL Server (+ lost deterministic ordering)**

**Problem:** this query uses a literal `LIMIT 1`. SQL Server has no `LIMIT`, so `GetAssociationTargetByIDOrHandle` fails on the SQL Server backend. The rewrite also dropped the previous `ORDER BY CASE WHEN uuid = ? THEN 0 ELSE 1 END`, so when a UUID and a different artifact's handle both match, the returned row is now nondeterministic.

**How it happened:** the portable `r.db.FetchFirstClause(1)` helper (and its required `ORDER BY`) was replaced with a hard-coded `LIMIT 1`.

**How to fix:** restore `ORDER BY CASE WHEN uuid = ? THEN 0 ELSE 1 END` and its bind arg, and append `r.db.FetchFirstClause(1)` instead of `LIMIT 1`.

AnuGayan (commented):
## Code review — DB schema refactor + DevPortal removal

Reviewed the full diff across all four DB backends and the Go repository/service/handler layers. The renames (`api_uuid`→`artifact_uuid`, `display_name`→`handle`) are applied consistently and the DevPortal removal is clean. I've left inline comments at each location; summary below.

### 🔴 Blockers
1. **`subscription_plans` lost the UNIQUE its composite FK needs** — breaks `CREATE TABLE subscriptions` on Postgres/SQL Server and subscription inserts on SQLite (verified locally). All four schema files.
2. **Raw `LIMIT 1` in `repository/application.go`** (×2) — fails on SQL Server, plus lost deterministic UUID-preference ordering.

### 🟠 Major
3. Nullable `created_by`/`updated_by` scanned into plain `string` → `converting NULL to string` (`subscription_repository.go` `GetByID`, `project.go`, `gateway.go`, `custom_policy.go`).
4. Binary columns (`configuration`/`openapi_spec`/`model_list`) written/read as Go `string` in `llm.go`/`mcp.go`/`websub_api.go`/`webbroker_api.go` — SQL Server `VARBINARY(MAX)` portability hazard + inconsistent with `gateway.go`/`deployment.go`.
5. `schema.sqlserver.sql` keeps `CHECK` constraints (subscriptions/gateways/gateway_tokens) the other three backends dropped → divergent validation.
6. `schema.sql` left `encrypted_secret` as `TEXT` while all other backends use a binary type.
7. `subscriptions` uniqueness re-keyed onto the nullable `application_id` → loses subscriber dedup and behaves differently on SQL Server vs others.

### 🟡 Minor / Nits
- `projects.handle` is `VARCHAR(255)` on mysql/sqlite/sqlserver but `VARCHAR(40)` on postgres (every other handle is 40).
- `gateway.go` `GetActiveTokenByHash` hardcodes `status = 'active'` instead of the constant.
- `handler/webbroker_api.go` keeps dead commented-out `PublishToDevPortal`/`UnpublishFromDevPortal` stubs (deleted outright in `websub_api.go`).
- `repository/api.go` `GetAPIAssociations`/`UpdateAPIAssociation` retain an unused `associationType` param.
- Stale `DefaultDevPortal` config remains in `config/config.go` + `config/default_config.go` after its consumer was removed.
- **`platform-api/src/api/generated.go` not regenerated** — `openapi.yaml` was stripped of DevPortal/publication, but the committed generated code still contains those types; the next `make generate` will diverge. Please run `make generate` and commit it.

### What's good
Careful cross-backend work (SQL Server multiple-cascade-path avoidance, idempotency guards, new supporting indexes), consistent renames end-to-end, correct `SMALLINT` boolean handling in `gateway.go`/`api.go`, and a clean DevPortal removal across server wiring, interfaces, constants, `openapi.yaml`, and `roles.yaml`.

platform-api/src/internal/repository/application.go
```diff
@@ -146,20 +161,27 @@ func (r *ApplicationRepo) GetAssociationTargetByIDOrHandle(targetIDOrHandle, org
...
+		INNER JOIN (
+			SELECT uuid, handle, name, version, created_at, updated_at FROM rest_apis
+			UNION ALL SELECT uuid, handle, name, version, created_at, updated_at FROM websub_apis
+			UNION ALL SELECT uuid, handle, name, version, created_at, updated_at FROM webbroker_apis
+			UNION ALL SELECT uuid, handle, name, version, created_at, updated_at FROM llm_providers
+			UNION ALL SELECT uuid, handle, name, version, created_at, updated_at FROM llm_proxies
+			UNION ALL SELECT uuid, handle, name, version, created_at, updated_at FROM mcp_proxies
+		) src ON src.uuid = a.uuid
+		WHERE a.organization_uuid = ? AND a.type = ? AND (a.uuid = ? OR src.handle = ?)
+		LIMIT 1
```
AnuGayan: 🔴 **Blocker — same `LIMIT 1` / lost-ordering issue as `GetAssociationTargetByIDOrHandle`**

**Problem:** `GetAssociationTargetByIDOrHandleAndKind` also uses literal `LIMIT 1` (fails on SQL Server) and dropped the UUID-preference `ORDER BY`.

**How it happened:** `FetchFirstClause(1)` + `ORDER BY` replaced by a hard-coded `LIMIT 1`.

**How to fix:** restore the deterministic `ORDER BY` and use `r.db.FetchFirstClause(1)`.

platform-api/src/internal/repository/subscription_repository.go
```diff
@@ -158,16 +159,17 @@ func isSubscriptionUniqueViolation(err error) bool {
...
 		FROM subscriptions
 		WHERE uuid = ? AND organization_uuid = ?
 	`
 	sub := &model.Subscription{}
 	var storedToken string
 	err := r.db.QueryRow(r.db.Rebind(query), subscriptionID, orgUUID).Scan(
-		&sub.UUID, &sub.APIUUID, &sub.SubscriberID, &sub.ApplicationID, &storedToken,
+		&sub.UUID, &sub.ArtifactUUID, &sub.SubscriberID, &sub.ApplicationID, &storedToken,
 		&sub.SubscriptionPlanID, &sub.OrganizationUUID, &sub.Status,
+		&sub.CreatedBy, &sub.UpdatedBy,
```
AnuGayan: 🟠 **Major — scanning nullable `created_by`/`updated_by` into a non-nullable `string` will panic on NULL**

**Problem:** `GetByID` scans `created_by`/`updated_by` directly into `&sub.CreatedBy, &sub.UpdatedBy` (both plain `string`). Those columns are nullable `VARCHAR(200)`, so any row with a NULL yields `sql: Scan error: converting NULL to string`.

**How it happened:** the new audit columns were added to the SELECT, but the scan targets weren't wrapped. Note this file's own `ListByFilters` already does it correctly with `sql.NullString` — `GetByID` is the inconsistent one.

**How to fix:** scan into `sql.NullString` locals and assign `.String`, matching `ListByFilters`.

platform-api/src/internal/repository/project.go
```diff
@@ -58,12 +58,13 @@ func (r *ProjectRepo) CreateProject(project *model.Project) error {
...
 	query := `
-		SELECT uuid, name, organization_uuid, description, created_at, updated_at
+		SELECT uuid, name, organization_uuid, description, created_by, created_at, updated_by, updated_at
 		FROM projects
 		WHERE uuid = ?
 	`
 	err := r.db.QueryRow(r.db.Rebind(query), projectId).Scan(
-		&project.ID, &project.Name, &project.OrganizationID, &project.Description, &project.CreatedAt, &project.UpdatedAt,
+		&project.ID, &project.Name, &project.OrganizationID, &project.Description,
+		&project.CreatedBy, &project.CreatedAt, &project.UpdatedBy, &project.UpdatedAt,
```
AnuGayan: 🟠 **Major — nullable `created_by`/`updated_by` scanned into plain `string` (NULL → scan error)**

**Problem:** this scan (and the equivalents in `GetProjectByNameAndOrgID` ~L88 and `GetProjectsByOrganizationID` ~L117) reads nullable `created_by`/`updated_by` directly into `&project.CreatedBy, &project.UpdatedBy` (plain `string`). A NULL value produces `sql: Scan error: converting NULL to string`.

**How it happened:** audit columns added to the projection without nullable scan handling. Only survives because fresh inserts write empty strings; seeded/legacy/other-path rows with NULL will break reads.

**How to fix:** scan into `sql.NullString` and assign `.String` (as `gateway.go`'s token reads already do).

platform-api/src/internal/repository/gateway.go
```diff
@@ -44,64 +45,85 @@ func (r *GatewayRepo) Create(gateway *model.Gateway) error {
...
+	gateway := &model.Gateway{}
+	var propertiesBytes []byte
+	var isCritical, isActive int
+	if err := row.Scan(
+		&gateway.ID, &gateway.OrganizationID, &gateway.Handle, &gateway.Name, &gateway.Description, &propertiesBytes, &gateway.Vhost,
+		&isCritical, &gateway.FunctionalityType, &gateway.Version, &isActive, &gateway.CreatedBy, &gateway.UpdatedBy, &gateway.CreatedAt, &gateway.UpdatedAt,
+	); err != nil {
+		return nil, err
+	}
+	gateway.IsCritical = isCritical != 0
```
AnuGayan, line 95: 🟠 **Major — gateway reads scan nullable `created_by`/`updated_by` into plain `string`**

**Problem:** `GetByUUID` (here) and `GetByOrganizationID`/`GetByHandleAndOrgID`/`List` (~L135/166/206) scan `&gateway.CreatedBy, &gateway.UpdatedBy` (plain `string`) from nullable columns → `converting NULL to string` on any NULL row.

**How it happened:** audit columns added to the SELECT without nullable handling — inconsistent with the gateway-token reads in this same file, which correctly use `sql.NullString`.

**How to fix:** use `sql.NullString` locals and assign `.String`, matching the token-read pattern below.

platform-api/src/internal/repository/custom_policy.go
```diff
@@ -48,22 +48,22 @@ func (r *CustomPolicyRepo) InsertCustomPolicy(policy *model.CustomPolicy) error
...
-		SELECT uuid, organization_uuid, name, display_name, version, description, policy_definition, created_at, updated_at
+		SELECT uuid, organization_uuid, name, display_name, version, description, policy_definition, created_by, updated_by, created_at, updated_at
 		FROM gateway_custom_policies
 		WHERE organization_uuid = ? AND name = ? AND version = ?
 	`
 	p := &model.CustomPolicy{}
 	err := r.db.QueryRow(r.db.Rebind(query), orgUUID, name, version).Scan(
 		&p.UUID, &p.OrganizationUUID, &p.Name, &p.DisplayName, &p.Version,
-		&p.Description, &p.PolicyDefinition, &p.CreatedAt, &p.UpdatedAt,
+		&p.Description, &p.PolicyDefinition, &p.CreatedBy, &p.UpdatedBy, &p.CreatedAt, &p.UpdatedAt,
```
AnuGayan: 🟠 **Major — nullable `created_by`/`updated_by` scanned into plain `string`**

**Problem:** this read (and the sibling list/get methods ~L96/133/163) scans `&p.CreatedBy, &p.UpdatedBy` (plain `string`) from nullable columns → `converting NULL to string` on NULL.

**How it happened:** new audit columns projected without nullable scan handling.

**How to fix:** scan into `sql.NullString` and assign `.String`.

platform-api/src/internal/repository/llm.go
```diff
@@ -329,15 +329,16 @@ func (r *LLMProviderRepo) Create(p *model.LLMProvider) error {
...
 
-	// Insert into llm_providers table
+	// Insert into llm_providers table (handle/name/version/timestamps now live here)
 	query := `
 		INSERT INTO llm_providers (
-			uuid, description, created_by, template_uuid, openapi_spec, model_list, status, configuration
+			uuid, handle, name, version, description, created_by, template_uuid, openapi_spec, model_list, configuration, created_at, updated_at, organization_uuid
 		)
-		VALUES (?, ?, ?, ?, ?, ?, ?, ?)`
+		VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)`
```
AnuGayan, line 343: 🟠 **Major — binary columns written as Go `string` (SQL Server portability + inconsistency)**

**Problem:** `p.OpenAPISpec` (a Go `string`) and `string(modelProvidersJSON)` are written into `openapi_spec`/`model_list`, which this PR changed to `BYTEA`/`BLOB`/`VARBINARY(MAX)`. lib/pq tolerates string↔bytea, but go-mssqldb's `VARBINARY(MAX)` does not reliably accept a `string` bind nor scan cleanly into one. `configuration` here (and the same pattern in `mcp.go`/`websub_api.go`/`webbroker_api.go`) has the same issue.

**How it happened:** the columns were migrated to binary and `gateway.go`/`deployment.go` were converted to `[]byte`, but the LLM/MCP/WebSub/WebBroker config/spec paths were left on `string`.

**How to fix:** write these payloads as `[]byte` and scan into `[]byte`, as done in `gateway.go`/`deployment.go`.

platform-api/src/internal/database/schema.sqlserver.sql
```diff
@@ -15,145 +15,158 @@
...
+    FOREIGN KEY (artifact_uuid, organization_uuid)
+        REFERENCES artifacts(uuid, organization_uuid) ON DELETE NO ACTION,
+    UNIQUE(artifact_uuid, subscription_token_hash),
+    UNIQUE(organization_uuid, artifact_uuid, application_id),
     CHECK (status IN ('ACTIVE', 'INACTIVE', 'REVOKED'))
```
AnuGayan: 🟠 **Major — SQL Server keeps CHECK constraints the other three backends dropped (divergent validation)**

**Problem:** this `CHECK (status IN ('ACTIVE','INACTIVE','REVOKED'))` on `subscriptions` — plus `CHECK (gateway_functionality_type IN (...))` on `gateways` (~L196) and `CHECK (status IN ('active','revoked'))` on `gateway_tokens` (~L258) — were **removed** in the postgres/mysql/sqlite schemas but **kept** here. So a status/type value the app accepts will insert on Postgres but be rejected on SQL Server (and vice-versa).

**How it happened:** the CHECK removals weren't mirrored into `schema.sqlserver.sql`.

**How to fix:** make this intentional and consistent — either keep these CHECKs on all four backends or drop them on all four. Don't leave SQL Server as the outlier.

platform-api/src/internal/database/schema.sql
```diff
@@ -222,174 +255,166 @@ CREATE TABLE IF NOT EXISTS deployment_status (
...
 
 -- WebSub API HMAC secrets table (for inbound webhook event verification)
 CREATE TABLE IF NOT EXISTS websub_api_hmac_secrets (
     uuid VARCHAR(40) PRIMARY KEY,
     artifact_uuid VARCHAR(40) NOT NULL,
-    name VARCHAR(63) NOT NULL,
-    display_name VARCHAR(255),
+    handle VARCHAR(40) NOT NULL,
+    name VARCHAR(255),
     encrypted_secret TEXT NOT NULL,
```
AnuGayan: 🟠 **Major — `encrypted_secret` left as `TEXT` here while every other backend uses a binary type**

**Problem:** postgres (`BYTEA`), sqlite (`BLOB`) and sqlserver (`VARBINARY(MAX)`) moved this column to binary, but the generic/MySQL `schema.sql` still has `encrypted_secret TEXT NOT NULL` — the same logical (encrypted, binary) column with a divergent type, and the Go layer writes encrypted bytes to it.

**How it happened:** the binary migration of this column was missed in `schema.sql` only.

**How to fix:** change to `encrypted_secret BLOB NOT NULL` to match the other three schemas.

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -18,148 +18,174 @@
...
-    FOREIGN KEY (api_uuid, organization_uuid)
+    FOREIGN KEY (artifact_uuid, organization_uuid)
       REFERENCES artifacts(uuid, organization_uuid) ON DELETE CASCADE,
-    UNIQUE(api_uuid, subscription_token_hash),
-    UNIQUE(api_uuid, subscriber_id, organization_uuid),
-    CHECK (status IN ('ACTIVE', 'INACTIVE', 'REVOKED'))
+    UNIQUE(artifact_uuid, subscription_token_hash),
+    UNIQUE(organization_uuid, artifact_uuid, application_id)
```
AnuGayan: 🟠 **Major — uniqueness now keyed on a nullable column; behaves differently per backend**

**Problem:** the unique key changed from `UNIQUE(api_uuid, subscriber_id, organization_uuid)` (`subscriber_id` NOT NULL) to `UNIQUE(organization_uuid, artifact_uuid, application_id)` where `application_id` is **nullable**. Two consequences: (1) the dedup guarantee that used to be on `subscriber_id` is gone; (2) NULL handling in UNIQUE differs by engine — SQL Server treats NULLs as equal (allows only **one** NULL `application_id` per artifact), while Postgres/SQLite/MySQL allow many. Same data, accepted on Postgres but rejected on SQL Server.

**How it happened:** the constraint was re-keyed around `application_id` as part of the artifact generalization.

**How to fix:** confirm this is intended; if multiple token-based (NULL-application) subscriptions must be allowed on all backends, rework this constraint (e.g. keep uniqueness on a NOT NULL column, or filter NULLs).

platform-api/src/internal/database/schema.sql
```diff
@@ -18,144 +18,170 @@
...
     created_at DATETIME DEFAULT CURRENT_TIMESTAMP,
+    updated_by VARCHAR(200),
     updated_at DATETIME DEFAULT CURRENT_TIMESTAMP
 );
 
 
 -- Projects table
 CREATE TABLE IF NOT EXISTS projects (
     uuid VARCHAR(40) PRIMARY KEY,
+    handle VARCHAR(255) NOT NULL,
```
AnuGayan: 🟡 **Minor — `projects.handle` length is inconsistent across backends**

**Problem:** `projects.handle` is `VARCHAR(255)` here (and in `schema.sqlite.sql` / `schema.sqlserver.sql`) but `VARCHAR(40)` in `schema.postgres.sql`. Every other `handle` column in the schema is `VARCHAR(40)`.

**How it happened:** the new `projects.handle` column wasn't given a consistent length across the four files.

**How to fix:** align all four to `VARCHAR(40)`.

platform-api/src/internal/repository/gateway.go
```diff
@@ -318,53 +319,75 @@ func (r *GatewayRepo) GetActiveTokensByGatewayUUID(gatewayId string) ([]*model.G
...
+	var revokedBy sql.NullString
 	query := `
-		SELECT uuid, gateway_uuid, token_hash, salt, status, created_at, revoked_at
+		SELECT uuid, gateway_uuid, token_hash, salt, status, created_by, created_at, revoked_by, revoked_at
 		FROM gateway_tokens
 		WHERE uuid = ?
 	`
 	err := r.db.QueryRow(r.db.Rebind(query), tokenId).Scan(
-		&token.ID, &token.GatewayID, &token.TokenHash, &token.Salt, &token.Status, &token.CreatedAt, &token.RevokedAt,
+		&token.ID, &token.GatewayID, &token.TokenHash, &token.Salt, &token.Status,
```
AnuGayan, line 363: 🔵 **Nit — `status = 'active'` hardcoded while sibling queries use the constant**

**Problem:** `GetActiveTokenByHash` hardcodes `status = 'active'`, whereas `GetActiveTokensByGatewayUUID`/`CountActiveTokens` were migrated to bind `constants.GatewayTokenStatusActive`. Same value, so not a runtime bug — just inconsistent.

**How it happened:** this query was missed during the constant migration.

**How to fix:** use `constants.GatewayTokenStatusActive` here too.

Thushani-Jayasekera:
TODOs:
-- MOVE subscription related quotas to a different table
-- We do not have the subscription plans available for an api stored... do we have to bring it???

platform-api/src/internal/database/schema.sqlserver.sql
Thushani-Jayasekera: we need to drop from sql server schema

platform-api/src/internal/database/schema.sqlite.sql
```diff
@@ -206,9 +238,11 @@ CREATE TABLE IF NOT EXISTS deployments (
     artifact_uuid VARCHAR(40) NOT NULL,
     organization_uuid VARCHAR(40) NOT NULL,
     gateway_uuid VARCHAR(40) NOT NULL,
-    base_deployment_id VARCHAR(40), -- Reference to the deployment used as base, NULL if based on "current"
-    content BLOB NOT NULL, -- Immutable deployment artifact (YAML string)
-    metadata TEXT, -- JSON object for flexible key-value metadata
+    base_deployment_id VARCHAR(40),
+    content BLOB NOT NULL,
+    metadata BLOB,
```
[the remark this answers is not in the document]
Thushani-Jayasekera: what is the source of learning?

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -209,8 +240,10 @@ CREATE TABLE IF NOT EXISTS deployments (
     gateway_uuid VARCHAR(40) NOT NULL,
     base_deployment_id VARCHAR(40),
     content BYTEA NOT NULL,
-    metadata TEXT, -- JSON object as TEXT
-    created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
+    metadata BYTEA,
```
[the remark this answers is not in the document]
Thushani-Jayasekera: what is the source of learning?

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -18,148 +18,174 @@
...
+    updated_at TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP,
     FOREIGN KEY (organization_uuid) REFERENCES organizations(uuid) ON DELETE CASCADE,
-    UNIQUE(organization_uuid, plan_name),
-    UNIQUE(uuid, organization_uuid),
-    CHECK (status IN ('ACTIVE', 'INACTIVE')),
-    CONSTRAINT chk_plan_throttle_pair CHECK (
-      (throttle_limit_count IS NULL AND throttle_limit_unit IS NULL) OR
-      (throttle_limit_count IS NOT NULL AND throttle_limit_unit IS NOT NULL)
-    )
+    UNIQUE(organization_uuid, handle)
```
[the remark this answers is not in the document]
Thushani-Jayasekera: docwe need this as uuid is the primary key?

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -18,148 +18,177 @@
...
     FOREIGN KEY (organization_uuid) REFERENCES organizations(uuid) ON DELETE CASCADE,
-    UNIQUE(organization_uuid, plan_name),
-    UNIQUE(uuid, organization_uuid),
-    CHECK (status IN ('ACTIVE', 'INACTIVE')),
-    CONSTRAINT chk_plan_throttle_pair CHECK (
-      (throttle_limit_count IS NULL AND throttle_limit_unit IS NULL) OR
-      (throttle_limit_count IS NOT NULL AND throttle_limit_unit IS NOT NULL)
-    )
+    UNIQUE(organization_uuid, handle),
+    UNIQUE(organization_uuid, uuid)
```
renuka-fernando:
```suggestion
```

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -18,148 +18,177 @@
...
 -- subscription_token: encrypted value (AES-256-GCM) for retrieval (legacy rows have hash)
 -- subscription_token_hash: SHA-256 hash for uniqueness and gateway sync
 CREATE TABLE IF NOT EXISTS subscriptions (
     uuid VARCHAR(40) PRIMARY KEY,
-    api_uuid VARCHAR(40) NOT NULL,
+    artifact_uuid VARCHAR(40) NOT NULL,
     subscriber_id VARCHAR(255) NOT NULL,
     application_id VARCHAR(255),
     subscription_token VARCHAR(512) NOT NULL,
     subscription_token_hash VARCHAR(64) NOT NULL,
```
renuka-fernando:
```suggestion
    subscription_token_hash VARCHAR(255) NOT NULL,
```

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -18,148 +18,177 @@
...
     subscription_plan_uuid VARCHAR(40),
     organization_uuid VARCHAR(40) NOT NULL,
```
renuka-fernando, line 132: Let's review this again
We have a rest api resource /org/{orgUUID}/subscription and a /subscriptions

cc: @Arshardh

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -18,148 +18,177 @@
...
 CREATE TABLE IF NOT EXISTS gateways (
     uuid VARCHAR(40) PRIMARY KEY,
     organization_uuid VARCHAR(40) NOT NULL,
+    handle VARCHAR(40) NOT NULL,
     name VARCHAR(255) NOT NULL,
-    version VARCHAR(64) NOT NULL DEFAULT '1.0',
-    display_name VARCHAR(255) NOT NULL,
     description VARCHAR(1023),
-    properties JSONB NOT NULL DEFAULT '{}'::jsonb,
+    version VARCHAR(30) NOT NULL DEFAULT 'v1.0',
```
renuka-fernando:
```suggestion
    version VARCHAR(30) NOT NULL DEFAULT '1.0',
```

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -18,148 +18,177 @@
...
+    UNIQUE(organization_uuid, handle)
+);
+
+-- Gateway Association Mappings table (links artifacts to gateways)
+CREATE TABLE IF NOT EXISTS gateway_association_mappings (
+    artifact_uuid VARCHAR(40) NOT NULL,
+    organization_uuid VARCHAR(40) NOT NULL,
+    gateway_uuid VARCHAR(40) NOT NULL,
+    created_at TIMESTAMPTZ DEFAULT NOW(),
+    updated_at TIMESTAMPTZ DEFAULT NOW(),
```
renuka-fernando, line 187: Let's add updated and created by

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -168,22 +197,25 @@ CREATE TABLE IF NOT EXISTS gateway_custom_policies (
     organization_uuid VARCHAR(40) NOT NULL,
     name VARCHAR(255) NOT NULL,
     display_name VARCHAR(255),
-    version VARCHAR(15) NOT NULL,
-    description TEXT,
-    policy_definition JSONB,
-    created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
-    updated_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
+    version VARCHAR(30) NOT NULL DEFAULT 'v1.0',
```
renuka-fernando:
```suggestion
    version VARCHAR(30) NOT NULL,
```

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -192,11 +224,13 @@ CREATE TABLE IF NOT EXISTS gateway_tokens (
...
-    revoked_at TIMESTAMP,
+    status VARCHAR(20) NOT NULL DEFAULT 'active',
+    data_version VARCHAR(20) NOT NULL DEFAULT '1.0',
+    created_by VARCHAR(200),
+    created_at TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP,
+    revoked_by VARCHAR(200),
+    revoked_at TIMESTAMPTZ,
     FOREIGN KEY (gateway_uuid) REFERENCES gateways(uuid) ON DELETE CASCADE,
-    CHECK (status IN ('active', 'revoked')),
     CHECK (revoked_at IS NULL OR status = 'revoked')
```
renuka-fernando:
```suggestion
```

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -18,148 +18,177 @@
...
+    created_by VARCHAR(200),
+    created_at TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP,
+    updated_by VARCHAR(200),
+    updated_at TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP,
+    FOREIGN KEY (organization_uuid) REFERENCES organizations(uuid) ON DELETE CASCADE,
+    UNIQUE(organization_uuid, handle)
+);
+
+-- Gateway Association Mappings table (links artifacts to gateways)
+CREATE TABLE IF NOT EXISTS gateway_association_mappings (
```
renuka-fernando:
```suggestion
CREATE TABLE IF NOT EXISTS artifact_gateway_mappings (
```

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -209,8 +243,10 @@ CREATE TABLE IF NOT EXISTS deployments (
     gateway_uuid VARCHAR(40) NOT NULL,
```
renuka-fernando, line 242: rename column deployment_id -> uuid

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -225,176 +261,171 @@ CREATE TABLE IF NOT EXISTS deployment_status (
     deployment_id VARCHAR(40) NOT NULL,
```
renuka-fernando:
```suggestion
    deployment_uuid VARCHAR(40) NOT NULL,
```

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -225,176 +261,171 @@ CREATE TABLE IF NOT EXISTS deployment_status (
...
-);
-
 -- LLM Provider Templates table
 CREATE TABLE IF NOT EXISTS llm_provider_templates (
     uuid VARCHAR(40) PRIMARY KEY,
     organization_uuid VARCHAR(40) NOT NULL,
-    handle VARCHAR(255) NOT NULL,
-    name VARCHAR(253) NOT NULL,
+    handle VARCHAR(40) NOT NULL,
+    group_id VARCHAR(255) NOT NULL,
```
renuka-fernando:
```suggestion
    group_id VARCHAR(40) NOT NULL,
```

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -192,209 +225,208 @@ CREATE TABLE IF NOT EXISTS gateway_tokens (
...
+    configuration BYTEA NOT NULL,
+    data_version VARCHAR(20) NOT NULL DEFAULT '1.0',
+    created_by VARCHAR(200),
+    created_at TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP,
+    updated_by VARCHAR(200),
+    updated_at TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP,
+    organization_uuid VARCHAR(40) NOT NULL,
     FOREIGN KEY (uuid) REFERENCES artifacts(uuid) ON DELETE CASCADE,
+    FOREIGN KEY (organization_uuid) REFERENCES organizations(uuid) ON DELETE CASCADE,
     FOREIGN KEY (project_uuid) REFERENCES projects(uuid) ON DELETE CASCADE,
```
renuka-fernando, line 340: Remove all the `ON DELETE RESTRICT`

platform-api/src/internal/database/schema.postgres.sql
renuka-fernando: Let's keep VARCHAR(255) for all hashes

platform-api/src/internal/database/schema.postgres.sql
renuka-fernando: Remove all CHECKs

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -431,68 +467,92 @@ CREATE TABLE IF NOT EXISTS application_api_keys (
...
+CREATE INDEX IF NOT EXISTS idx_artifact_gateway_mappings_artifact_uuid ON artifact_gateway_mappings(artifact_uuid);
+CREATE INDEX IF NOT EXISTS idx_artifact_gateway_mappings_gateway_uuid ON artifact_gateway_mappings(gateway_uuid);
+
+CREATE TABLE IF NOT EXISTS audit (
+   uuid VARCHAR(40) PRIMARY KEY,
+   action VARCHAR(50) NOT NULL,
+   resource_uuid VARCHAR(40) NOT NULL,
+   resource_type VARCHAR(50),
+   organization_uuid VARCHAR(40) NOT NULL,
+   performed_by VARCHAR(200),
```
renuka-fernando, line 550: new indexes for
- organization_uuid + resource_uuid

renuka-fernando (dismissed):

renuka-fernando (approved):

PR #2269, opened 2026-06-25
[Devportal] Implement ApiKey-App association functionality

### LINKED ISSUE #2256: [Feature][Devportal]: API key - Application Association capability

Area/DeveloperPortal (Developer Portal API, Developer Portal UI)

Aspect/Other (Anything else which does not match above categories)

### Suggested Feature

Add the capability to associate API keys with Applications in the devportal.
This is required for supporting app based analytics.

### Related Issues

_No response_

Piumal1999:
## Purpose
Fix #2256

## Goals
Implementing the functionality to associate an API Key with an Application.

## Approach
- Added a nullable `APP_ID` FK on `DP_API_KEY` → `DP_APPLICATION` (one app per key, ON DELETE SET NULL) — association is optional and purely for analytics attribution, no effect on key validity/authorization.
- Added new REST endpoints: PUT/DELETE `/api-keys/{apiKeyId}/application` and GET `/applications/{applicationId}/api-keys`; extended listApiKeys/generateApiKey with an optional appId. 
- Introduced a new webhook event, `apikey.application_updated` — per-key ({ key_id, application }), mirroring the existing `apikey.*` event family. Fires on associate/dissociate, and fans out one event per affected key on app rename/delete.
- UI: "Associate/change app" action + filter on the API Keys page; a new "API keys" section on the Application page to view/associate/remove keys for that app.
- Updated docs

## Samples

image

image


lasanthaS:
If there's no expiry date, shall we keep it as "Never"? Also, let's rename key actions to just "Actions"

lasanthaS (approved):

Issue #2109, opened 2026-06-05
[Feature]: API policy support for LLM providers


ashera96:
### Please select the area the issue is related to

Area/AIGateway (AI Gateway runtime/control plane)

### Please select the aspect the issue is related to

Aspect/AI (AI/LLM integration, MCP, AI readiness)

### Suggested Feature

Add support to attach policies at global level for LLM providers and proxies.

ashera96:
- Discussion related to the feature: https://github.com/wso2/api-platform/discussions/2125

ashera96:
- Doc PR: https://github.com/wso2/api-platform/pull/2358

PR #2289, opened 2026-06-26
[Gateway] Add data_version column to gateway-controller artifacts table

ashera96:
## Purpose

- Related to https://github.com/wso2/api-platform/issues/2109

Adds a `data_version` column to the gateway-controller `artifacts` table to record the **data-shape version** of each stored artifact (REST API, WebSub API, WebBroker API, LLM Provider, LLM Proxy, MCP Proxy).

The value is `<major>.<minor>`:
- **major** — parsed from the YAML `apiVersion` suffix (`gateway.api-platform.wso2.com/v1` → `1`).
- **minor** — a code-defined constant per kind, starting at `0` (so `v1` → `1.0`). Bump it when a kind's stored data shape changes in a backward-compatible way **without** bumping `apiVersion`. Stored rows then record exactly which data shape they were written with, enabling future data migrations.

This decouples the wire/API version (`apiVersion`, the major) from the stored-data-shape version (the minor).

## Changes

- **Schema (3 dialects):** `data_version` added to `artifacts` after `version`, matching each dialect's existing style — `TEXT NOT NULL DEFAULT '1.0'` for SQLite and Postgres, `NVARCHAR(20) NOT NULL DEFAULT '1.0'` for SQL Server. SQLite schema version bumped `3 → 4` (`PRAGMA user_version` and `currentSchemaVersion`).
- **Model:** `StoredConfig.DataVersion` field and a `GetApiVersion()` helper that prefers `SourceConfiguration` (the user-authored artifact) and falls back to `Configuration` (the LLM-derived RestAPI form).
- **Helper:** `ComputeDataVersion(kind, apiVersion)` plus an exhaustive `dataMinorVersions` map keyed by `ArtifactKind`.
- **Writes:** an `ensureDataVersion()` chokepoint in `SaveConfig`, `UpdateConfig` and `UpsertConfig`, so every artifact-addition path records a `data_version` and an explicitly-set value is never overwritten.
- **Reads:** `data_version` threaded through all artifact `SELECT`/scan paths (single-row reads and the joined `scanConfigRows` queries).

### DIFF
--- a/gateway/gateway-controller/pkg/storage/gateway-controller-db.sql
+++ b/gateway/gateway-controller/pkg/storage/gateway-controller-db.sql
@@ -1,5 +1,5 @@
 -- SQLite Schema for Gateway-Controller API Configurations
--- Version: 2
+-- Version: 4
 
 -- Base table for all artifact types (REST APIs, WebSub APIs, LLM Providers, LLM Proxies, MCP Proxies)
 CREATE TABLE IF NOT EXISTS artifacts (
@@ -7,6 +7,7 @@ CREATE TABLE IF NOT EXISTS artifacts (
     gateway_id TEXT NOT NULL,
     display_name TEXT NOT NULL,
     version TEXT NOT NULL,
+    data_version TEXT NOT NULL DEFAULT '1.0',
     kind TEXT NOT NULL,
     handle TEXT NOT NULL,
     desired_state TEXT NOT NULL CHECK(desired_state IN ('deployed', 'undeployed')),
@@ -307,4 +308,4 @@ CREATE TABLE IF NOT EXISTS webhook_secrets (
 
 CREATE INDEX IF NOT EXISTS idx_webhook_secrets_artifact ON webhook_secrets(gateway_id, artifact_uuid);
 
-PRAGMA user_version = 3;
+PRAGMA user_version = 4;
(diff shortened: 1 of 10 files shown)


Krishanx92 (approved):

gateway/gateway-controller/pkg/storage/gateway-controller-db.sql
```diff
@@ -1,12 +1,13 @@
...
--- Version: 2
+-- Version: 4
 
 -- Base table for all artifact types (REST APIs, WebSub APIs, LLM Providers, LLM Proxies, MCP Proxies)
 CREATE TABLE IF NOT EXISTS artifacts (
     uuid TEXT NOT NULL,
     gateway_id TEXT NOT NULL,
     display_name TEXT NOT NULL,
     version TEXT NOT NULL,
+    data_version TEXT NOT NULL DEFAULT '1.0',
```
renuka-fernando, line 10: Are we only doing this for artifact table?
cc: @malinthaprasan

renuka-fernando (approved):

gateway/gateway-controller/pkg/storage/gateway-controller-db.sql
malinthaprasan: Shall we just blindly do it to all the tables? :)
It doesn't do any harm right. No code changes needed as well. Just a default value with 1.0.
+ @tharindu1st

Issue #2279, opened 2026-06-25
[Task]: Refactor Platform API DB Schemas and Update code


Thushani-Jayasekera:
### Please select the area the issue is related to

Area/Other (Anything else which does not match above categories)

### Please select the aspect the issue is related to

Aspect/API (API backends, definitions, contracts, interfaces, OpenAPI)

### Description

$subject

DinithHerath:
Hi @Thushani-Jayasekera 

With the PR https://github.com/wso2/api-platform/pull/2254 the subscription plan creation breaks. Now ValidThrottleLimitUnits has uppercase length units[1] but in [2] we still require the lowercase units. Lets fix that.

[1] https://github.com/wso2/api-platform/blob/29c7a173aaa4de3d96312e0eb600625da884e931/platform-api/src/internal/service/subscription_plan_service.go#L81
[2] https://github.com/wso2/api-platform/blob/29c7a173aaa4de3d96312e0eb600625da884e931/platform-api/src/internal/handler/subscription_plan_handler.go#L127

Thanks
Dinith

PR #2305, opened 2026-06-26
Refactor subscription plans and limits structure in the database & rename tables

Thushani-Jayasekera:
- https://github.com/wso2/api-platform/issues/2279

**Summary**

This PR introduces a normalized subscription_plan_liplan can eventually define multiple throttling limits(burst + sustained request counts, bandwidth quotas, AI token quotas), and renames two join tables to the *_mappings convention. The REST API, gateway events, and gateway-controller conlimit per plan for now — the multi-limit storage is in place, but the surrounding code is intentionally left single-limit and marked with NOTE/TODO for a follow-up.

**Changes**

New subscription_plan_limits table (all four dialects — schema.sql, .sqlite, .postgres, .sqlserver)
- Columns: uuid, subscription_plan_uuid, organizatioOUNT | BANDWIDTH | TOTAL_TOKEN_COUNT | …), limit_count (nullable BIGINT), time_amount, time_unit, data_unit (nullable, only set for BANDWIDTH), stop_on_quota_reach, timestamps.
- The quota window is time_amount × time_unit, enablst+sustained pairs on one plan.
- FKs reference subscription_plans(uuid, organization_uuid) for tenant consistency; SQL Server uses NO ACTION on the org/composite
edges to avoid the multiple-cascade-paths restrictioia the plan→CASCADE edge.
- Added UNIQUE(organization_uuid, uuid) to subscription_plans as the composite-FK target (mirrors the artifacts pattern), plus indexes
on the new table.

Moved throttle fields out of subscription_plans
- Dropped the inline throttle_limit_count, throttle_limit_unit, and stop_on_quota_reach columns.
- The repository now reads via a LEFT JOIN to the si and Create/Update write the plan + exactly one limitrow in a transaction. A row is always written (a nil count persists as NULL) so stopOnQuotaReach round-trips faithfully even for plans
with no quota.
- model.SubscriptionPlan keeps ThrottleLimitCount / ThrottleLimitUnit / StopOnQuotaReach (no longer DB-mapped), documented with the
single-limit assumption.

Table renames (platform-api only)
- application_api_keys → application_api_key_mappings
- application_artifacts → application_artifact_mappi
- Updated table + index names, repository queries, and integration tests.

No impact on REST API or gateway behaviour

- Request/response DTOs and JSON contracts are unchanged (still built from the model fields).
- Gateway event payloads (SubscriptionPlan{Created,Ue-identical; the gateway-controller (separate DB/code) is untouched.
- Verified by the repository and integration test su

Follow-up (tracked via NOTE/TODO in code)

- Surface and propagate all limit rows through the Rcontroller (currently single REQUEST_COUNT only).

Migration note

These are CREATE-only schema files. Existing databasUE(organization_uuid, uuid), move data from the threeinline columns into subscription_plan_limits, then drop the columns. (Not included in this PR — happy to add it.)

Testing

- go build ./internal/... ✅
- go test ./internal/repository/... ✅
- go test -tags integration ./internal/integration/... ✅


platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -115,7 +113,38 @@ CREATE TABLE IF NOT EXISTS subscription_plans (
...
+    uuid VARCHAR(40) PRIMARY KEY,
+    subscription_plan_uuid VARCHAR(40) NOT NULL,
+    organization_uuid VARCHAR(40) NOT NULL,
+    limit_type VARCHAR(20) NOT NULL DEFAULT 'REQUEST_COUNT',
+    -- Nullable: a single row is always written per plan to carry stop_on_quota_reach,
+    -- even when the plan defines no quota count. A NULL limit_count means "no quota".
+    limit_count BIGINT,
+    time_amount INTEGER NOT NULL DEFAULT 1,
+    time_unit VARCHAR(20) NOT NULL,
+    data_unit VARCHAR(10),
```
Thushani-Jayasekera: when limit_type is bandwith

renuka-fernando: Let's add this as a comment in the sql script itself

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -115,7 +113,38 @@ CREATE TABLE IF NOT EXISTS subscription_plans (
...
+-- bandwidth quotas, AI token quotas). limit_type selects the dimension
+-- (REQUEST_COUNT, BANDWIDTH, TOTAL_TOKEN_COUNT, ...). The quota window is
+-- (time_amount x time_unit). data_unit (KB/MB/GB) is only set for BANDWIDTH.
+-- NOTE: the platform-api, REST APIs, gateway events and gateway-controller
+-- currently read/write only a SINGLE REQUEST_COUNT limit per plan. This needs
+-- to be improved to surface and propagate all limit rows.
+CREATE TABLE IF NOT EXISTS subscription_plan_limits (
+    uuid VARCHAR(40) PRIMARY KEY,
+    subscription_plan_uuid VARCHAR(40) NOT NULL,
+    organization_uuid VARCHAR(40) NOT NULL,
```
renuka-fernando: Do we need org uuid here?

For forign key we can just use subscription_plan_uuid and remove the unique constraint in subscription_plan table (with org uuid one)

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -115,7 +113,38 @@ CREATE TABLE IF NOT EXISTS subscription_plans (
...
+    time_amount INTEGER NOT NULL DEFAULT 1,
+    time_unit VARCHAR(20) NOT NULL,
+    data_unit VARCHAR(10),
+    stop_on_quota_reach SMALLINT NOT NULL DEFAULT 1,
+    created_at TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP,
+    updated_at TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP,
+    FOREIGN KEY (subscription_plan_uuid, organization_uuid)
+        REFERENCES subscription_plans(uuid, organization_uuid) ON DELETE CASCADE,
+    FOREIGN KEY (organization_uuid) REFERENCES organizations(uuid) ON DELETE CASCADE,
+    UNIQUE(subscription_plan_uuid, limit_type, time_unit)
```
renuka-fernando: This unique key wont' work now right?
This now won't work with different time amounts
ex:
1. 100 req per 30 min
2. 10 req per min

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -118,6 +116,32 @@ CREATE TABLE IF NOT EXISTS subscription_plans (
...
+    uuid VARCHAR(40) PRIMARY KEY,
+    subscription_plan_uuid VARCHAR(40) NOT NULL,
+    limit_type VARCHAR(20) NOT NULL DEFAULT 'REQUEST_COUNT',
+    -- Nullable: a single row is always written per plan to carry stop_on_quota_reach,
+    -- even when the plan defines no quota count. A NULL limit_count means "no quota".
+    limit_count BIGINT,
+    time_amount INTEGER NOT NULL DEFAULT 1,
+    time_unit VARCHAR(20) NOT NULL,
+    -- Data unit (KB/MB/GB) for the quota; only set when limit_type is BANDWIDTH.
+    data_unit VARCHAR(10),
```
renuka-fernando:
```suggestion
    limit_count_unit VARCHAR(10),
```

This is the unit for `limit_count` right?
And we can keep these columns next to each other, I meant the order of columns.

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -118,6 +116,32 @@ CREATE TABLE IF NOT EXISTS subscription_plans (
...
+    stop_on_quota_reach SMALLINT NOT NULL DEFAULT 1,
+    created_at TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP,
+    updated_at TIMESTAMPTZ DEFAULT CURRENT_TIMESTAMP,
```
renuka-fernando: Don't we need created_by and updated_by?

platform-api/src/internal/database/schema.postgres.sql
renuka-fernando: What about this order?
- limit_type
- time_unit
- time_amount
- limit_count
- limit_count_unit

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -118,6 +116,32 @@ CREATE TABLE IF NOT EXISTS subscription_plans (
...
+    FOREIGN KEY (subscription_plan_uuid) REFERENCES subscription_plans(uuid) ON DELETE CASCADE,
+    UNIQUE(subscription_plan_uuid, limit_type, time_amount, time_unit)
```
renuka-fernando, line 131: +1 for this UNIQUE constraint

ex:
- 1, REQUEST_COUNT, 30, min

platform-api/src/internal/database/schema.postgres.sql
Thushani-Jayasekera: yes

platform-api/src/internal/database/schema.postgres.sql
Thushani-Jayasekera: can we remove created_at, updated_at as well from this table. Because ideally it would be sufficient to store in subscription_plans table

renuka-fernando: +1

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -118,6 +116,30 @@ CREATE TABLE IF NOT EXISTS subscription_plans (
...
+-- to be improved to surface and propagate all limit rows.
+CREATE TABLE IF NOT EXISTS subscription_plan_limits (
+    uuid VARCHAR(40) PRIMARY KEY,
+    subscription_plan_uuid VARCHAR(40) NOT NULL,
+    limit_type VARCHAR(20) NOT NULL DEFAULT 'REQUEST_COUNT',
+    time_unit VARCHAR(20) NOT NULL,
+    time_amount INTEGER NOT NULL DEFAULT 1,
+    -- Nullable: a single row is always written per plan to carry stop_on_quota_reach,
+    -- even when the plan defines no quota count. A NULL limit_count means "no quota".
+    limit_count BIGINT,
```
renuka-fernando: Can there be plans without a limit?

platform-api/src/internal/database/schema.postgres.sql
```diff
@@ -119,6 +117,21 @@ CREATE TABLE IF NOT EXISTS subscription_plans (
...
+-- Subscription plan limits table (throttling limits for a plan).
+CREATE TABLE IF NOT EXISTS subscription_plan_limits (
+    uuid VARCHAR(40) PRIMARY KEY,
+    subscription_plan_uuid VARCHAR(40) NOT NULL,
+    limit_type VARCHAR(20) NOT NULL DEFAULT 'REQUEST_COUNT',
+    time_unit VARCHAR(20) NOT NULL,
+    time_amount INTEGER NOT NULL DEFAULT 1,
+    limit_count BIGINT NOT NULL,
+
+    limit_count_unit VARCHAR(10),
```
renuka-fernando:
```suggestion
    limit_count_unit VARCHAR(10),
```

platform-api/src/internal/database/schema.sql
```diff
@@ -120,6 +118,21 @@ CREATE TABLE IF NOT EXISTS subscription_plans (
...
+-- Subscription plan limits table (throttling limits for a plan).
+CREATE TABLE IF NOT EXISTS subscription_plan_limits (
+    uuid VARCHAR(40) PRIMARY KEY,
+    subscription_plan_uuid VARCHAR(40) NOT NULL,
+    limit_type VARCHAR(20) NOT NULL DEFAULT 'REQUEST_COUNT',
+    time_unit VARCHAR(20) NOT NULL,
+    time_amount INTEGER NOT NULL DEFAULT 1,
+    limit_count BIGINT NOT NULL,
+
+    limit_count_unit VARCHAR(10),
```
renuka-fernando:
```suggestion
    limit_count_unit VARCHAR(10),
```

platform-api/src/internal/database/schema.sqlite.sql
```diff
@@ -119,6 +117,21 @@ CREATE TABLE IF NOT EXISTS subscription_plans (
...
+-- Subscription plan limits table (throttling limits for a plan).
+CREATE TABLE IF NOT EXISTS subscription_plan_limits (
+    uuid VARCHAR(40) PRIMARY KEY,
+    subscription_plan_uuid VARCHAR(40) NOT NULL,
+    limit_type VARCHAR(20) NOT NULL DEFAULT 'REQUEST_COUNT',
+    time_unit VARCHAR(20) NOT NULL,
+    time_amount INTEGER NOT NULL DEFAULT 1,
+    limit_count BIGINT NOT NULL,
+
+    limit_count_unit VARCHAR(10),
```
renuka-fernando:
```suggestion
    limit_count_unit VARCHAR(10),
```

platform-api/src/internal/database/schema.sqlserver.sql
```diff
@@ -130,6 +128,24 @@ CREATE TABLE dbo.subscription_plans (
...
+IF OBJECT_ID(N'dbo.subscription_plan_limits', N'U') IS NULL
+CREATE TABLE dbo.subscription_plan_limits (
+    uuid VARCHAR(40) PRIMARY KEY,
+    subscription_plan_uuid VARCHAR(40) NOT NULL,
+    limit_type VARCHAR(20) NOT NULL DEFAULT 'REQUEST_COUNT',
+    time_unit VARCHAR(20) NOT NULL,
+    time_amount INT NOT NULL DEFAULT 1,
+    limit_count BIGINT NOT NULL,
+
+    limit_count_unit VARCHAR(10),
```
renuka-fernando:
```suggestion
    limit_count_unit VARCHAR(10),
```

renuka-fernando (approved):

PR #2288, opened 2026-06-25
[API Platform] Update OpenAPI documentation and refactor according to rest api guidelines

Thushani-Jayasekera:
$subject

Assessment Summary — Platform API (openapi-review branch)

Spec: platform-api/src/resources/openapi.yaml · 187 operations · OpenAPI 3.1.1

Generic api-design assessment (base skill)

```
┌───────────────────────────────┬───────┬──────────┬──────┬────────┬──────┐
│           Dimension           │ Score │ Critical │ High │ Medium │ Low  │
├───────────────────────────────┼───────┼──────────┼──────┼────────┼──────┤
│ Agent Readiness · Spectral    │ 83%   │ 0        │ 93   │ 1072   │ 1239 │
├───────────────────────────────┼───────┼──────────┼──────┼────────┼──────┤
│ Agent Readiness · AI Analysis │ 90%   │ 0        │ 3    │ 4      │ 2    │
├───────────────────────────────┼───────┼──────────┼──────┼────────┼──────┤
│ Security Readiness            │ 89%   │ 1        │ 324  │ 0      │ 0    │
├───────────────────────────────┼───────┼──────────┼──────┼────────┼──────┤
│ Design Readiness              │ 98%   │ 0        │ 0    │ 27     │ 2    │
└───────────────────────────────┴───────┴──────────┴──────┴────────┴──────┘
```

PR #2308, opened 2026-06-27
[API Platform] Update OpenAPI documentation and refactor according to rest api guidelines

Thushani-Jayasekera:
Relates to https://github.com/wso2/api-platform/pull/2288

- https://github.com/wso2/api-platform/issues/2312


platform-api/src/resources/openapi.yaml
```diff
@@ -8345,21 +8205,21 @@ components:
...
         subscriberId:
           type: string
           minLength: 1
           description: Unique subscriber identifier for the subscription (required)
           example: "user-123"
         applicationId:
           type: string
-          description: Application ID (from DevPortal/STS). Optional in token-based subscriptions.
-          example: "a1b2c3d4-e5f6-7890-abcd-ef1234567890"
+          description: Application handle. Optional in token-based subscriptions.
```
Thushani-Jayasekera, line 8217: breaking UIs

platform-api/src/resources/openapi.yaml
```diff
@@ -8345,21 +8205,21 @@ components:
       properties:
         apiId:
           type: string
-          description: API handle or UUID (REST API identifier)
-          example: "c9f2b6ae-1234-5678-9abc-def012345678"
+          description: REST API handle
```
renuka-fernando, line 8208: Is this the Rest API handle or the Artifact handle?
BTW, we may need to include Kind as well, if we are not going with UUID

platform-api/src/resources/openapi.yaml
```diff
@@ -586,7 +586,7 @@ paths:
         '500':
           $ref: '#/components/responses/InternalServerError'
 
-  /rest-apis/{apiId}:
+  /rest-apis/{apiHandle}:
```
renuka-fernando, line 589: Shall we go with just {id}


```suggestion
  /rest-apis/{id}:
```

malinthaprasan: Yeah, in general "id" is what we use to fetch a resource. handle or whatever the representation of the id is an internal detail.
This was done because of an entry in the guideline? We should fix the guideline if its the case.

malinthaprasan: Lets keep this until we come to a conclusion and change altogether

Thushani-Jayasekera:
closing this PR as this was handled through a different PR

PR #2314, opened 2026-06-28
test(platform-api): convert cross-DB integration tests to godog and broaden coverage

AnuGayan:
## Summary
Converts the platform-api cross-database integration tests from Go table-tests to a **godog (Cucumber)** suite, keeping the `integration` build tag and the existing `make it-*` targets, and extends coverage to several previously-untested repositories. Also fixes one latent SQL Server bug surfaced by the new secret tests.

## Cross-DB suite (`platform-api/src/internal/integration`)
- Replace `crud`/`cascade`/`lifecycle`/`harness` `*_test.go` with Gherkin `features/` + step definitions and an error-returning harness. Each scenario runs against a fresh DB and surfaces as a `go test` subtest via godog's `TestingT`.
- New scenarios covering the **LLM control plane** (provider template / provider / proxy), **MCP proxy**, **WebBroker API**, the **secret store**, and the **deployment store**. The LLM/MCP scenarios store a **dummy upstream API key** — no real LLM/MCP server or credential needed on any engine.
- The **deployment** scenarios drive `DeploymentRepo` through its real methods so the dialect-specific SQL runs on every engine: `CreateWithLimitEnforcement` (the `deployment_status` upsert + FETCH-first hard-limit pruning), `GetCurrentByGateway` (FETCH-first), `GetDeploymentsWithState` (the `ROW_NUMBER() OVER` ranking), `GetStatus`, `HasActiveDeployment`, `GetWithContent`, and the undeploy transition via `SetCurrentWithDetails` — the path whose SQL Server undeploy failure was fixed upstream.
- Carries over the recent upstream additions to the old table-tests: the `application_api_key_mappings` / `application_artifact_mappings` renames and the **subscription-plan-limit cascade**.
- Adds `godog v0.15.0`.

## Fix (`internal/repository/secret.go`)
The secret scenario surfaced a latent **SQL Server** bug where the `FindRefsAndSoftDelete` row-lock used the `LIMIT` keyword (rejected by SQL Server). It now uses the dialect-aware `FetchFirstClause` helper. (`SecretRepo.List` was independently fixed upstream with the same `PaginationClause` approach.)

## Test plan
- ✅ SQLite: `IT_DB=sqlite make it` — **23 scenarios / 176 steps pass**; existing `internal/repository` unit tests pass; `go vet -tags integration` and default build clean.
- ✅ `pr-check` / `pr-check-postgres` / `pr-check-sqlserver` (the cross-DB repo tests) green.
- ⏳ `make it-postgres` / `make it-sqlserver` need Docker + built images (verify the `secret.go` SQL Server fix locally).


AnuGayan:
Pushed fixes for the CodeRabbit comments and the failing e2e builds.

CI: the `e2e (postgres|sqlite|sqlserver)` jobs were failing in the **Build platform-api image** step (pre-existing on `main`): the Dockerfile needs a `httpkit` build-context, but the e2e workflow only passed `common`. Added `--build-context httpkit=../httpkit` to `.github/workflows/platform-api-gateway-e2e.yml`, mirroring `platform-api/Makefile`.

CodeRabbit comments:
- `steps_crud.go` `revokedTokenState` — now asserts the status captured from the scenario instead of the hardcoded constant (and the feature uses the real value `revoked`).
- `steps_llm.go` `createTemplateVersion` — derives the new template handle suffix from the requested `version` (`v2.0` -> `v2-0`) instead of a fixed string.
- `steps_llm.go` `createLLMProvider` — asserts `Create` populated the generated UUID, matching `createLLMTemplate`.
- `steps_pagination.go` `pageWebSubAPIs` — now pages with a general offset loop instead of assuming exactly two full pages.
- `steps_secret.go` `firstSecretRotated` — also asserts the rotated hash, not just the ciphertext.
- `steps_test.go` (e2e) `chatResponseIsOpenAIShaped` — decodes the JSON and asserts `object == "chat.completion"` plus a non-empty `choices` array instead of substring matching.

The WebBroker `Update` comment is already covered: the scenario's next `Then` step reads the row back and asserts the persisted status.

Re-verified on SQLite: 21 scenarios / 157 steps pass; `go vet -tags integration` and both module builds clean.

AnuGayan:
Update on the e2e jobs.

The `e2e` checks were failing on `main` before this PR because the workflow's platform-api image build was missing the `httpkit` build-context the Dockerfile requires, so the suite never actually ran. With that fixed, the bootstrap also needed a fix (login posts form-encoded credentials, matching the auth-login handler's `r.PostForm`). After both fixes:

- e2e (postgres): pass
- e2e (sqlite): pass
- e2e (sqlserver): fail

The remaining sqlserver failure is a **pre-existing, server-side SQL Server bug, unrelated to this PR**: the existing api-deployment scenario's undeploy step returns `500 Failed to undeploy deployment`. Deploy/serve works on SQL Server; only `DeploymentService.UndeployDeployment` 500s. It is not a test/harness issue and not introduced here.

Suggested follow-up (separate PR): fix the SQL Server incompatibility in the deployment undeploy path and add a `DeploymentRepo` cross-DB scenario to the new godog suite (the deployment repo create/list/status/restore/undeploy path is currently only exercised via the delete-cascade, so this class of bug isn't yet guarded). Happy to take that on next.

Note: a combined platform-api -> gateway LLM-proxy deployment e2e scenario was prototyped but descoped from this PR; the deploy-to-serve path returned 404 at the gateway and needs gateway-internals work better suited to its own change.

AnuGayan:
@copilot resolve the merge conflicts in this pull request

PR #2316, opened 2026-06-29
Update template ID/handle to groupId and update REST APIs

Tharanidk:
## Purpose
The handle should be changed to groupId with the change of some endpoints.

## Approach

- changed the handle to groupID

- Changed the REST APIs as follows.

<meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-090f3d4c-7fff-3b7a-a62e-7ba9e3e6f510"><div dir="ltr" style="margin-left:0pt;" align="left">
Method | Path | Description
-- | -- | --
POST | /llm-provider-templates | Create a custom template (starts at v1.0)
GET | /llm-provider-templates |  List all templates. Returns every version by default; 
GET |  /llm-provider-templates?query=latest:true | return only the latest version of each family( (used by the catalog listing)
GET | llm-provider-templates?query = groupId:openai | List all versions of the template group
GET| /llm-provider-templates?query=groupId:openai&version:v2.0 | Get a single version of the group. 
GET | /llm-provider-templates/{id} | Get one specific template version by its handle (not "latest")
PUT | /llm-provider-templates/{id} | Update the specific version identified by that handle
PATCH | /llm-provider-templates/{id} | Enable or disable a version
DELETE | /llm-provider-templates/{id} | Delete a version, blocked while in use
POST |/llm-provider-templates/copy?fromTemplateId=openai-v3-0&toTemplateId=openai-v4-0&toVersion=v4.0 | Create a new version in a family by cloning fromTemplateId (optional body overrides fields)(demotes the previous isLatest)

</div></b>


Tharanidk:
@CodeRabbit please review

Tharanidk:
@CodeRabbit please review

platform-api/src/resources/openapi.yaml
```diff
@@ -2229,17 +2229,33 @@ paths:
...
+        - name: query
+          in: query
+          required: false
+          description: >-
+            When set, create a new version within an existing template family
+            instead of a new family. URL-encoded DSL holding the family id,
+            e.g. query=groupId:openai (sent as groupId%3Aopenai).
+          schema:
+            type: string
+          example: groupId:openai
```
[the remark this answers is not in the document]
Tharanidk: This is invalid. There are no existing clients on the route /llm-provider-templates/{groupId}/versions

platform-api/src/resources/openapi.yaml
```diff
@@ -1139,213 +1091,211 @@ paths:
...
-          $ref: '#/components/responses/Forbidden'
-        '404':
-          $ref: '#/components/responses/NotFound'
         '409':
           $ref: '#/components/responses/Conflict'
         '500':
           $ref: '#/components/responses/InternalServerError'
 
-  /llm-provider-templates/{id}/versions:
     get:
```
renuka-fernando, line 1054: This should not be under `/llm-provider-templates/copy:`.
Should be moved to `/llm-provider-templates:`

renuka-fernando: No need to have a GET for /llm-provider-templates/copy:

Tharanidk: fixed with [e91fbb7](https://github.com/wso2/api-platform/pull/2316/commits/e91fbb7b6a0f76c6f8e2d0643af5db49e9d38f86)

platform-api/src/resources/openapi.yaml
```diff
@@ -1059,6 +1073,29 @@ paths:
...
+          description: >-
+            URL-encoded search DSL. `query=groupId:<id>` lists that family's
+            versions; adding `&version:<ver>` returns the single full template
+            for that version. Terms are `&`-separated and the whole value is
+            percent-encoded (e.g. groupId%3Aopenai%26version%3Av1.0). The version
+            may instead be supplied via the standalone `version` query param.
+          schema:
+            type: string
+          example: groupId:openai&version:v1.0
+        - name: version
```
renuka-fernando, line 1088: Does this query param is used? Version?

platform-api/src/resources/openapi.yaml
```diff
@@ -1074,91 +1111,98 @@ paths:
             type: integer
             minimum: 0
             default: 0
-        - name: versions
+        - name: latest
```
renuka-fernando, line 1114: Even this should come under query I think.

renuka-fernando (approved):
Let's do these changes in a separate PR

platform-api/src/resources/openapi.yaml
Tharanidk: fixed with f8a02f4d888a855076ef4f63d5ee5b24a3764caa

platform-api/src/resources/openapi.yaml
Tharanidk: fixed with f8a02f4d888a855076ef4f63d5ee5b24a3764caa

PR #2349, opened 2026-07-01
Add changes to integrate BFF layer to secret management feature

npamudika:
## Purpose
Add changes to integrate BFF layer to secret management requests after adding the changes in https://github.com/wso2/api-platform/pull/2311

Fix the on demand secret resolving issue at the gw controller

### DIFF
--- a/gateway/gateway-controller/pkg/constants/constants.go
+++ b/gateway/gateway-controller/pkg/constants/constants.go
@@ -18,6 +18,17 @@
 
 package constants
 
+import "regexp"
+
+// SecretPlaceholderRe matches {{ secret "handle" }} placeholders embedded in YAML
+// or JSON content. The quotes may be JSON-escaped (\") so both forms are matched.
+// The handle is in capture group 1.
+//
+// This is the single canonical definition — use it everywhere secret handles must
+// be extracted (on-demand sync, template pre-scan, etc.) to guarantee that
+// extraction and rendering can never silently diverge.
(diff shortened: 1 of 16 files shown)


gateway/gateway-controller/pkg/constants/constants.go
```diff
@@ -18,6 +18,17 @@
...
+import "regexp"
+
+// SecretPlaceholderRe matches {{ secret "handle" }} placeholders embedded in YAML
+// or JSON content. The quotes may be JSON-escaped (\") so both forms are matched.
+// The handle is in capture group 1.
+//
+// This is the single canonical definition — use it everywhere secret handles must
+// be extracted (on-demand sync, template pre-scan, etc.) to guarantee that
+// extraction and rendering can never silently diverge.
+var SecretPlaceholderRe = regexp.MustCompile(`\{\{\s*secret\s+\\?"([^"\\]+)\\?"\s*\}\}`)
```
renuka-fernando, line 30: Mm, I don't feel that hardcoding this is a good option. Let me check further.

npamudika: This will be used temporary until we add the event based notifications for the secrets related updates.

renuka-fernando: +1 doing this temporary

This {{ secret }} is rendered from go template engine we have to fix this later.

renuka-fernando (approved):

PR #2365, opened 2026-07-01
Rename default organization name to "default"

Piumal1999:
## Purpose
$subject

## Goals
To keep consistant default organization across products

## Approach
Updated the configs


platform-api/src/config/config.toml
```diff
@@ -67,8 +67,8 @@ enabled = false
 
 [auth.file_based.organization]
 id     = "99089a17-72e0-4dd8-a2f4-c8dfbb085295"
-name   = "AP Organization"
-handle = "ap-org"
+name   = "default"
+handle = "default"
```
Thushani-Jayasekera, line 71: @renuka-fernando shall we rename the configs also to 
```
handle = uuid/ default
display_name = default <rename>
region = "us"
```

POST /organizations
```
{
  "id": org handle claim / id config (this can be a unique identifier/ uuid - this is what we save for handle column)
  "displayName": org name claim /  default
  "region": "us"
}
```
extract org_id claim and set it to idp_ref_orgid 


When validating, org_id claim will be validated (the uuid)

https://github.com/wso2/api-platform/pull/2369

Thushani-Jayasekera: Current behaviour
```
┌─────────────────────────────────────┬────────┬──────────────────────────────────────────────────────────┐
│                Layer                │  Uses  │                           Why                            │
├─────────────────────────────────────┼────────┼──────────────────────────────────────────────────────────┤
│ JWT claim (organization)            │ UUID   │ Source of truth for validation; can't be spoofed via URL │
├─────────────────────────────────────┼────────┼──────────────────────────────────────────────────────────┤
│ Database (PK + all FKs)             │ UUID   │ Stable, immutable key; tables use organization_uuid      │
├─────────────────────────────────────┼────────┼──────────────────────────────────────────────────────────┤
│ URL path params ({organizationId})  │ Handle │ Human-readable slug                                      │
├─────────────────────────────────────┼────────┼──────────────────────────────────────────────────────────┤
│ API response (organizationId field) │ Handle │ User-friendly                                            │
└─────────────────────────────────────┴────────┴──────────────────────────────────────────────────────────┘
```

I am wondering whether the get /organizations/{id} call should have the org uuid for the lookup, instead of the handle, eventhough the uuid is not returned in any of the responses. wdyt?

renuka-fernando: For the UUID that is got from the JWT claim, shall we keep it in another column in the same table?
- UUID column - our own UUID
- IDP_REF_ORG_UUID - UUID from IDP

renuka-fernando: +1 for config handle and display name

renuka-fernando (approved):

Thushani-Jayasekera:
closing since this was handled by a seperate PR. @Piumal1999 pls reopen if necessary

PR #2486, opened 2026-07-06
[AI Gateway] Add multi-provider routing for LLM proxies

Aakashwije:
## Purpose
Before this change, using more than one LLM vendor from a single app meant standing up a separate proxy and endpoint per vendor, and formatting requests differently for each. There was no way to route one OpenAI-shaped endpoint to multiple upstreams, and translator policies had to be declared separately from the provider they targeted - duplicating the provider `id` by hand and keeping it in sync.

Relates to #2490 

## Goals
- Let a single LLM proxy fan out to multiple LLM providers behind **one endpoint**, with the target chosen per-request by a header.
- Move each vendor translator **inline onto the provider it belongs to** (`additionalProviders[].transformer`), so provider auth and its translator live together and the provider `id` is injected automatically.
- Keep the client contract unchanged: requests and responses stay OpenAI-shaped regardless of the selected vendor.
- Reject invalid transformer config at **deploy time**, not request time.

## Approach

A proxy declares a primary `provider` plus a list of `additionalProviders`, each with its own upstream auth and an inline `transformer`. The `openai-header-router` policy reads the `x-provider` header and writes the choice into `metadata["selected_provider"]`. The controller injects each provider's translator and upstream-auth as **conditional policies** gated on `selected_provider == <provider>`, so only the selected provider's translator and key are applied. Each additional provider is exposed as a named upstream for loopback routing.

```
Client (OpenAI-shaped request)
    │  Request + X-Provider
    ▼
LLM Proxy
    1. api-key-auth (consumer)
    2. openai-header-router  → metadata.selected_provider
    3. translator (openai-to-x)
    4. set-headers (loopback auth)
    │  Loopback
    ▼
Selected LLM Provider
    1. api-key-auth (validates the loopback key)
    2. upstream.auth (config) → injects vendor key
    │  HTTPS
    ▼
External LLM  (OpenAI · Anthropic · Azure OpenAI · Mistral · Gemini)
```

New config reads top-to-bottom as "provider + how to talk to it":

```yaml
additionalProviders:
  - id: anthropic-provider
    auth: { type: api-key, header: X-API-Key, value: ... }
    transformer:
      type: openai-to-anthropic
      version: v1
      params:
        model: claude-sonnet-4-5-20250929   # id injected automatically
```

Implemented across three modules:

- **gateway-controller** — `additionalProviders` + inline `transformer` in the management schema and validator; `transformProxy` exposes each additional provider as a named upstream and injects auth/translator policies gated on a `selected_provider` execution condition.
- **platform-api** — `additionalProviders` across openapi, model, dto and service, carried into the deployment artifact sent to the gateway.
- **kubernetes-operator** — `additionalProviders` on the `LlmProxy` CRD (types, deepcopy, regenerated CRD manifests) and controller payload.

## User stories
- As an **app developer**, I can call one endpoint and switch LLM vendors with a single header, without changing my request format or my code.
- As a **platform admin**, I can attach multiple providers to one proxy, declare each vendor's translator right next to its auth, and have invalid config rejected when I deploy it.
- As a **security-conscious operator**, my app holds zero vendor credentials — vendor keys live only in provider config and are injected at the gateway.

## Documentation

N/A for this PR. The example YAMLs under `gateway/examples/` (`openai-multi-provider-proxy.yaml` and per-provider configs) serve as usage documentation. Product-doc updates, if any, will accompany the follow-up translator-policies PR.

## Automation tests
**Unit tests**

- New tests in `gateway-controller/pkg/utils` cover conditional upstream-auth injection and conditional transformer injection — correct `selected_provider` execution condition, injected `id` param, and per-provider isolation.
- Full unit suites pass across gateway-controller, platform-api, and kubernetes-operator; no regressions in existing tests.

**Integration tests**

- Verified end-to-end on a local gateway stack: 5 providers (OpenAI, Anthropic, Azure OpenAI, Mistral, Gemini) each route correctly by `x-provider` header and return their own vendor model.
- Consumer auth returns `401` on wrong/missing key; deploy-time validation returns `400` for a transformer missing `type`/`version`.
- Concurrency burst (100 requests across 5 providers, 10 concurrent) showed **zero cross-provider routing**.
- The godog `llm-proxies.feature` scenarios land with the follow-up translator-policies PR (they depend on those policy binaries).

## Security checks

- Ran FindSecurityBugs plugin and verified report? **N/A** - FindSecurityBugs is a Java/SpotBugs tool; this change is Go. `go vet` was run clean.
- Confirmed that this PR doesn't commit any keys, passwords, tokens, usernames, or other secrets? **yes** - staged diff scanned; only `REPLACE_WITH_*` placeholders and non-secret loopback identifiers. Real keys kept in a gitignored local path.

## Samples
- `gateway/examples/openai-multi-provider-proxy.yaml` - one proxy fanning out to multiple providers with inline transformers.
- Per-provider configs: `openai-provider.yaml`, `anthropic-provider.yaml`, `azure-openai-provider.yaml`, `mistral-provider.yaml`, `gemini-provider.yaml`.

## Related PRs

Follow-up (separate PR): translator policy implementations (`openai-to-anthropic` / `azure-openai` / `mistral` / `gemini`, `openai-header-router`), their `build.yaml` / `build-manifest.yaml` wiring, and the `llm-proxies.feature` integration test.

## Test environment
- Go 1.26.2 (matches `gateway-controller/go.mod`)
- Docker Desktop — gateway-runtime, gateway-builder, gateway-controller images (tag `1.2.0-M2-SNAPSHOT`)
- macOS (darwin)
- SQLite (controller local store)
866034cc-5297-4283-8222-f24b0649f220


gateway/gateway-controller/api/management-openapi.yaml
```diff
@@ -5837,6 +5837,63 @@ components:
...
+          example: anthropic-provider
+        as:
+          type: string
+          description: >
+            Logical upstream name used by policies to select this provider. Must
+            be unique within the proxy. Defaults to `id` when omitted.
+          pattern: '^[a-zA-Z0-9\-_]+$'
+          minLength: 1
+          maxLength: 100
+          example: anthropic-upstream
```
renuka-fernando, line 5862: @Arshardh can you remember why we need this `as`?

gateway/gateway-controller/api/management-openapi.yaml
```diff
@@ -5837,6 +5837,63 @@ components:
...
+        LlmProvider over the internal loopback route.
+      properties:
+        id:
+          type: string
+          description: Unique id of a deployed llm provider
+          example: anthropic-provider
+        as:
+          type: string
+          description: >
+            Logical upstream name used by policies to select this provider. Must
```
renuka-fernando:
```suggestion
            Logical LLM Provider name used by policies to select this provider. Must
```

gateway/gateway-controller/api/management-openapi.yaml
Arshardh: We discussed to have an optional alias just in case the provider names will not be readable enough to configure in a routing policy

gateway/examples/anthropic-openai-proxy.yaml
```diff
@@ -0,0 +1,42 @@
+# --------------------------------------------------------------------
```
renuka-fernando, line 11: Let's add license headers, otherwise this looks license headers

gateway/examples/anthropic-provider.yaml
```diff
@@ -0,0 +1,62 @@
...
+    kind: LlmProvider
+    name: anthropic-provider
+  displayName: Anthropic provider loopback key
+  apiKey:
+    value: anthropic_provider_loopback_key_1234567890abcdef
+    # anthropic_provider_loopback_key_1234567890abcdef
+  maskedApiKey: anthropic_provider_loopback_key_1234********cdef
+  expiresIn:
+    duration: 30
+    unit: days
```
renuka-fernando: Let's remove this API key YAML

gateway/examples/openai-multi-provider-proxy.yaml
```diff
@@ -0,0 +1,132 @@
...
+  version: v1.0
+  context: /openai-multi
+  provider:
+    id: openai-provider
+    auth:
+      type: api-key
+      header: X-API-Key
+      value: openai_provider_loopback_key_1234567890abcdef
+  additionalProviders:
+    - id: anthropic-provider
```
renuka-fernando, line 44: If `as` is used, let's add a sample for `as` as well.

platform-api/resources/openapi.yaml
```diff
@@ -7964,6 +7973,28 @@ components:
...
+          example: anthropic-provider
+        as:
+          type: string
+          description: >
+            Logical upstream name used by policies to select this provider. Must
+            be unique within the proxy. Defaults to `id` when omitted.
+          pattern: '^[a-zA-Z0-9\-_]+$'
+          minLength: 1
+          maxLength: 100
+          example: anthropic-upstream
```
renuka-fernando, line 7996: Let's add transformers same as the GW

gateway/gateway-controller/pkg/config/llm_validator.go
```diff
@@ -590,13 +590,109 @@ func (v *LLMValidator) validateProxyData(spec *api.LLMProxyConfigData) []Validat
 			Message: "spec.provider.id must consist of lowercase alphanumeric characters, hyphens, or dots, and must start and end with an alphanumeric character",
 		})
 	}
+	if spec.Provider.Auth != nil {
```
renuka-fernando, line 593: @Arshardh

Do we need to make provider auth required?

Arshardh: nope, we don't need to

PR #2489, opened 2026-07-07
[Platform API] Add CORS configuration to server settings and update .gitignore

Thushani-Jayasekera:
- Introduced CORS struct in the server configuration to manage allowed origins for credentialed requests.
- Updated validation logic to ensure non-wildcard origins are specified when not in demo mode.
- Enhanced CORS middleware to conditionally allow credentials based on the configured origins.
- Added entries to .gitignore to exclude sensitive key material and local directories from version control.


malinthaprasan (approved):

malinthaprasan:
Should by default disable CORS. inbuilt apps doesn't need this as they have BFFs.

Issue #2341, opened 2026-07-01
[Task]:  Restrict changing the handle for PUT opertion


renuka-fernando:
### Please select the area the issue is related to

Area/Management (Management API or Management Portal UI)

### Please select the aspect the issue is related to

Aspect/API (API backends, definitions, contracts, interfaces, OpenAPI)

### Description

Ex: PUT /rest-apis/{restApiId} should not update the handle of the Rest API

PR #2496, opened 2026-07-07
Refactor immutability checks across various handlers and services to use utility functions for better consistency

Thushani-Jayasekera:
## Purpose
- https://github.com/wso2/api-platform/issues/2341

### DIFF
--- a/platform-api/internal/utils/handle.go
+++ b/platform-api/internal/utils/handle.go
@@ -46,10 +46,10 @@ var (
 // the resource's path handle. A nil or empty body value is treated as "not provided"
 // and allowed; only a non-empty value that differs from the path handle is rejected.
 func ValidateHandleImmutable(pathHandle string, bodyHandle *string) error {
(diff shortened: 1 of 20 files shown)


platform-api/internal/utils/handle.go
```diff
@@ -42,6 +42,25 @@ var (
 	multipleHyphensRegex = regexp.MustCompile(`-+`)
 )
 
+// ValidateHandleImmutable checks a PUT request body's optional handle/id field against
+// the resource's path handle. A nil or empty body value is treated as "not provided"
+// and allowed; only a non-empty value that differs from the path handle is rejected.
+func ValidateHandleImmutable(pathHandle string, bodyHandle *string) error {
```
renuka-fernando, line 48: Shall we use the second func from this func. From this func we can do the null check and call the second func

renuka-fernando (approved):

Issue #2475, opened 2026-07-06
[Bug]: API Key policy removal not synced after re-enabling deployment sync


Thushani-Jayasekera:
### Please select the area the issue is related to

Area/Management (Management API or Management Portal UI)

### Please select the aspect the issue is related to

Aspect/API (API backends, definitions, contracts, interfaces, OpenAPI)

### Description

$subject

1. deploy an LLM Proxy with API Key auth policy with `deployment_sync_enabled` to true.
2. docker-compose down gateway and set `deployment_sync_enabled` to false and restart gateway
3. Remove API Key auth policy and deploy API. There should not be any change reflected in UI side 
4. docker-compose down gateway and set `deployment_sync_enabled` to true and restart gateway
5. Now we expect the security tab to display as no auth

<img width="1512" height="687" alt="Image" src="https://github.com/user-attachments/assets/c5a05adb-7e42-4a3a-8eeb-1dbced34b246" />

However, it still shows as API Key auth is enabled.
Based on the rest API call response, the policy is properly removed from CP side. However, security section has not been updated properly.
```
    "security": {
        "apiKey": {
            "enabled": true,
            "in": "header",
            "key": "X-API-Key"
        },
        "enabled": true
    },
```

### Steps to Reproduce

as above

### Severity Level of the Issue

Severity/Blocker (Core functionality is broken and consumer is blocked. Need immediate attention)

thivindu:
Fixed via - https://github.com/wso2/api-platform/pull/2505

PR #2505, opened 2026-07-07
Reset cp_sync_status on update of artifacts from gateway

thivindu:
This pull request updates the config upsert logic to ensure that `cp_sync_status` and `cp_sync_info` are always reset on update, not just on insert. This fixes an issue where these fields could become stale, especially for gateway-origin artifacts, and adds regression tests to prevent future breakage.

Fixes -
- https://github.com/wso2/api-platform/issues/2475

**Bug fix: Upsert logic for sync status fields**

* Updated the comment in `sql_store.go` to clarify that `cp_sync_status` and `cp_sync_info` are reset on update, while `cp_artifact_id` and `created_at` are only set on insert.
* Modified the SQL upsert operation in `sql_store.go` to explicitly set `cp_sync_status` and `cp_sync_info` on updates, ensuring their values are refreshed appropriately.

**Testing improvements**

* Added regression tests in `sqlite_test.go` to verify that:
  - Updating a gateway-origin config resets `cp_sync_status` to pending but preserves `cp_artifact_id`.
  - Updating a control-plane-origin config keeps `cp_sync_status` as NULL.

### DIFF
--- a/gateway/gateway-controller/pkg/storage/sql_store.go
+++ b/gateway/gateway-controller/pkg/storage/sql_store.go
@@ -648,7 +648,8 @@ func (s *sqlStore) UpsertConfig(cfg *models.StoredConfig) (bool, error) {
 
 	// Portable insert-or-update guarded by deployed_at: only overwrite when the
 	// incoming deployed_at is strictly newer (or the stored value is NULL).
-	// cp_sync_* and created_at are written on INSERT only — preserved on update.
+	// cp_artifact_id and created_at are written on INSERT only — preserved on update
+	// cp_sync_status and cp_sync_info ARE reset on update
 	didWrite, err := s.upsert(tx, upsertSpec{
 		table: "artifacts",
 		columns: []string{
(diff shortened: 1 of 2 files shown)


Thushani-Jayasekera (approved):

gateway/gateway-controller/pkg/storage/sql_store.go
```diff
@@ -667,10 +668,12 @@ func (s *sqlStore) UpsertConfig(cfg *models.StoredConfig) (bool, error) {
 			"display_name = ?", "version = ?", "data_version = ?", "kind = ?", "handle = ?",
 			"desired_state = ?", "deployment_id = ?", "origin = ?",
 			"updated_at = ?", "deployed_at = ?",
+			"cp_sync_status = ?", "cp_sync_info = ?",
```
renuka-fernando, line 671: So in the existing code, any updates to an Artifact are not reflected in the CP in general?
I mean, as defined in this issue https://github.com/wso2/api-platform/issues/2475 the GW has to restart and switch on/off sync. Or is just a PUT /rest-apis also not working?

thivindu: Updates to artifacts using PUT calls when the sync was off were not reflected in AI Workspace when sync was turned on and the gateway connects. Updates to an artifact is reflected in the CP in real time when sync is on. The issue only arises when the update is done when sync is off.

PR #2521, opened 2026-07-07
Retry update on creation failure in dp to cp import

thivindu:
### Summary

When two or more gateways push an artifact with the same `metadata.name` (handle) to the control plane at the same time, the importer could fail one of the pushes with a unique-constraint error and, on the gateway's retry, leave a duplicate deployment. This PR makes the importer race-safe by catching the unique violation, re-resolving the artifact by handle, and retrying the push as an update.
### Root cause

`ArtifactImportService.importValidated` resolves an existing artifact with `GetByHandle` and then lets the per-kind importer take its Create branch when nothing was found. The read and the write are not in one critical section, so two concurrent pushes of the same (org, handle) both observe nil and both attempt an insert. The child table's `UNIQUE(organization_uuid, handle)` index lets only one commit; the loser gets a unique-constraint violation that surfaced as a failed import. It affects all artifact kinds (REST, LLM provider, LLM proxy, MCP proxy, LLM provider template) since it lives in the shared import path.

### Changes

- `repository/errors.go` (new): IsUniqueViolation(err) — a shared, dialect-aware detector for unique/primary-key violations across the three databases the CP supports (SQLite, PostgreSQL, SQL Server). Matches through %w wrapping. The existing unexported isUniqueViolation now delegates to it (and gains SQL Server coverage).
- `service/artifact_import.go`: extracted the resolve → decide → import step into `resolveAndImport(...)` and wrapped it in a bounded retry loop (`importConflictMaxRetries = 3`) in `importValidated`. On a `IsUniqueViolation` error, it re-resolves by handle (the winner's row now exists) and retries — which takes the importer's Update branch and re-runs the last-in-wins decision (`DecideMetadataWrite`).

The retried (losing) push re-enters `DecideMetadataWrite` with the now-committed watermark:
a newer deployedAt → WriteFullMetadata (it wins the working copy)
a stale one → `SkipWorkingCopy` (the winner's copy is untouched)

Only the spurious insert failure is removed. Each push still records its own immutable deployment row.

Why catch-and-retry (not ON CONFLICT / advisory locks)

The CP targets three SQL dialects, so ON CONFLICT/MERGE and Postgres advisory locks aren't portable. Catch-and-retry is portable and already idiomatic in this codebase (subscription_repository does the same). Verified that every per-kind importer/repo propagates the raw DB error via %w (no sentinel translation) and that each child table carries the UNIQUE(organization_uuid, handle) constraint, so the violation is reliably detectable for all kinds.

Testing

- repository.TestIsUniqueViolation — SQLite / Postgres (23505) / SQL Server (2627, 2601) strings, %w-wrapped, nil, and unrelated errors.
- service.TestArtifactImport_RetriesAsUpdateOnHandleConflict — a forced unique violation is recovered via retry, re-resolves to the same CP UUID (no duplicate artifact), and the newer push wins.
- service.TestArtifactImport_RetryPreservesLastInWinsForStalePush — a stale loser leaves the winner's working copy unchanged after retry.
- service.TestImportersReturnUniqueViolationOnDuplicateHandle — against the real SQLite driver, confirms a duplicate handle produces an IsUniqueViolation-detectable error for all five kinds (REST, LLM provider, LLM proxy, MCP proxy, template).

### DIFF
--- a/platform-api/internal/repository/errors.go
+++ b/platform-api/internal/repository/errors.go
@@ -1,0 +1,50 @@
+/*
+ *  Copyright (c) 2026, WSO2 LLC. (http://www.wso2.org) All Rights Reserved.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ */
+
+package repository
+
+import "strings"
+
+// IsUniqueViolation reports whether err is a database unique-constraint (or
+// primary-key) violation. It is dialect-aware across the three databases the
+// control plane supports — SQLite, PostgreSQL and SQL Server — matching on the
+// stable substrings/error codes each driver surfaces.
+//
+// Detection is by err.Error() substring, so a violation wrapped with
+// fmt.Errorf("...: %w", err) is still detected without unwrapping.
+func IsUniqueViolation(err error) bool {
+	if err == nil {
+		return false
+	}
+	s := err.Error()
+	switch {
+	// SQLite: "UNIQUE constraint failed: <table>.<column>"
+	case strings.Contains(s, "UNIQUE constraint failed"):
+		return true
+	// PostgreSQL (pgx): SQLSTATE 23505, "duplicate key value violates unique constraint".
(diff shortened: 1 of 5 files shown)


platform-api/internal/repository/errors.go
```diff
@@ -0,0 +1,50 @@
...
+import "strings"
+
+// IsUniqueViolation reports whether err is a database unique-constraint (or
+// primary-key) violation. It is dialect-aware across the three databases the
+// control plane supports — SQLite, PostgreSQL and SQL Server — matching on the
+// stable substrings/error codes each driver surfaces.
+//
+// Detection is by err.Error() substring, so a violation wrapped with
+// fmt.Errorf("...: %w", err) is still detected without unwrapping.
+func IsUniqueViolation(err error) bool {
```
renuka-fernando, line 29: Mm, so we have to update this func when we introduce a new DB type, neh

thivindu: Yes, since we are throwing the raw error from the repository layer and we need to catch this specific error.

renuka-fernando (approved):

Issue #2534, opened 2026-07-08
/mcp is appended to all MCP proxying calls


Arshardh:
### Please select the area the issue is related to

Gateway

### Please select the aspect the issue is related to

Aspect/API (API backends, definitions, contracts, interfaces, OpenAPI)

### Description

When calling an MCP proxy's backend, there is a /mcp part appended always. there can be servers that does not support that part.

### Steps to Reproduce

Create MCP Proxy, add MCP backend that does not support /mcp, do a call to the proxy, and it should fail

### Severity Level of the Issue

Severity/Major (Important functionality is broken. Should be prioritized. Doesn't need immediate attention)

PR #2565, opened 2026-07-09
Remove appending /mcp when calling MCP server backends

Arshardh:


gateway/configs/config.toml
```diff
@@ -51,3 +51,10 @@ pricing_file = "/etc/policy-engine/llm-pricing/model_prices.json"
 enabled = false
 artifacts_dir = "/etc/api-platform-gateway/immutable_gateway/artifacts"
 
+[mcp]
```
RakhithaRR: We don't need this here right? The default value is already included in config.go

Arshardh: Removed

RakhithaRR (approved):

Krishanx92:
Fix https://github.com/wso2/api-platform/issues/2534

Issue #2568, opened 2026-07-09
[Task]: [Platform API] Add config to disable TLS for platform API


Thushani-Jayasekera:
### Please select the area the issue is related to

Platform API

### Please select the aspect the issue is related to

Aspect/Configuration (Config files, settings, env vars, defaults)

### Description

$subject

Issue #2564, opened 2026-07-09
[Task]:  Refactor AI Workspace started log to show portal URL


Thushani-Jayasekera:
### Please select the area the issue is related to

AI Workspace

### Please select the aspect the issue is related to

Aspect/UI (Frontend layouts, components, styling), Aspect/Logging (Log formats, instrumentation, improvements)

### Description

$subject

PR #2571, opened 2026-07-09
[Platform API[ Enhance TLS configuration for Platform API

Thushani-Jayasekera:
- https://github.com/wso2/api-platform/issues/2568
- https://github.com/wso2/api-platform/issues/2564

This document enumerates the possible TLS configurations across the two
components in play and the network hops between them:

- **Platform API** — the Go backend (HTTPS listener on `9243` by default).
- **AI Workspace BFF** — serves the SPA and reverse-proxies the browser to the
  Platform API. It is a TLS *server* to the browser and a TLS *client* to the
  Platform API.

The guiding principle: **each TLS boundary is configured independently, and by
the mechanism that fits it.** A listener you own is a boolean toggle; a client
connection takes its on/off from the target URL scheme, not a separate flag.

---

## 1. Platform API alone

The Platform API exposes **two independent listeners**. Enable either or both.

| Listener | Default | Port | Env overrides |
|---|---|---|---|
| HTTPS | **on** | `9243` | `HTTPS_ENABLED`, `HTTPS_PORT`, `HTTPS_CERT_DIR` (legacy: `TLS_ENABLED`, `PORT`, `TLS_CERT_DIR`) |
| Plain HTTP | **off** | `9080` | `HTTP_ENABLED`, `HTTP_PORT` |

**Certificate behavior (HTTPS listener):**

- `HTTPS_CERT_DIR` must contain `cert.pem` and `key.pem` when
  `APIP_DEMO_MODE=false`.
- In demo mode (`APIP_DEMO_MODE=true`, the default) a self-signed pair is
  generated into that directory automatically.

**Possible states:**

| # | HTTP | HTTPS | Meaning | Safe to expose directly? |
|---|---|---|---|---|
| A | off | on | Default. TLS-only backend. | Yes (self-signed warns; mount a real cert in prod) |
| B | on | off | Plain HTTP only. | **No** — only behind a TLS-terminating upstream or on a trusted internal network |
| C | on | on | Both, each on its own port (e.g. migrate clients HTTP→HTTPS without downtime). | HTTPS port yes; HTTP port only as in (B) |
| D | off | off | No listener — misconfiguration. | — |

> Enable the plain-HTTP listener only when a trusted upstream (ingress,
> service-mesh sidecar) terminates TLS, or for internal traffic. Never expose it
> directly to untrusted networks.

---

## 2. AI Workspace → Platform API (direct)

The BFF connects straight to the Platform API. There are **two boundaries**:

```
Browser ──(hop 1: inbound)──▶ BFF ──(hop 2: outbound)──▶ Platform API
```

### Hop 1 — Browser → BFF (inbound, BFF is the TLS server)

Controlled by a boolean, `BFF_TLS_ENABLED` (default `true`).

| `BFF_TLS_ENABLED` | Cert source | Result |
|---|---|---|
| `true` (default) | Mounted `BFF_TLS_CERT_FILE` + `BFF_TLS_KEY_FILE` | BFF serves HTTPS with your cert |
| `true` | No cert mounted, **demo mode** | BFF serves HTTPS with an auto-generated self-signed cert (browser warns) |
| `true` | No cert mounted, **prod mode** | **Startup fails** — a real cert is required |
| `true` | Exactly one of cert/key present | **Startup fails** — incomplete mount is not a silent downgrade |
| `false` | — | BFF serves **plain HTTP**. Only valid when something upstream terminates TLS (see topology 3) |

### Hop 2 — BFF → Platform API (outbound, BFF is the TLS client)

The **`PLATFORM_API_URL` scheme is the single source of truth** — there is no
separate on/off flag that could contradict it.

| `PLATFORM_API_URL` | TLS on this hop? | Trust configuration |
|---|---|---|
| `http://platform-api:9080` | No | none (setting a CA file or skip-verify here is rejected at startup) |
| `https://platform-api:9243` | Yes | see the trust ladder below |

**Trust ladder for an `https://` upstream** (pick the highest rung that applies):

| Rung | Config | Verification | Use |
|---|---|---|---|
| 1. Public CA | *(nothing)* | on | Upstream cert signed by a public CA |
| 2. Private CA | `PLATFORM_API_CA_FILE=/path/ca.pem` | **on** | **Recommended for the default self-signed Platform API.** PEM bundle appended to the system roots |
| 3. Skip verify | `PLATFORM_API_TLS_SKIP_VERIFY=true` | **off** | Dev/demo only. **Rejected when `APIP_DEMO_MODE=false`** — use rung 2 instead |

**Matching the two hops to the Platform API from topology 1:**

| BFF `PLATFORM_API_URL` | Requires Platform API listener |
|---|---|
| `https://…:9243` | HTTPS listener on (state A or C) |
| `http://…:9080` | HTTP listener on (state B or C) |

---

## 3. AI Workspace → Platform API through another hop

An intermediary (ingress controller, load balancer, service-mesh sidecar, or an
API gateway) sits on one or both boundaries and terminates TLS on behalf of the
component behind it. The knobs don't change — you're just moving *where* TLS is
terminated.

### 3a. Intermediary in front of the BFF (inbound offload)

```
Browser ──HTTPS──▶ Ingress ──HTTP──▶ BFF ──?──▶ Platform API
                  (terminates TLS)
```

- Set `BFF_TLS_ENABLED=false` — the BFF serves plain HTTP; the ingress presents
  the public cert to the browser.
- Outside demo mode this logs a warning (plain HTTP listener); keep the BFF off
  untrusted networks so only the ingress can reach it.
- The BFF→Platform API hop is unaffected — configure it exactly as in topology 2,
  hop 2.

### 3b. Intermediary between the BFF and the Platform API (outbound hop)

```
BFF ──▶ (proxy / gateway / mesh sidecar) ──▶ Platform API
```

The BFF only ever sees the **next hop** — point `PLATFORM_API_URL` at the
intermediary, and its scheme decides TLS for the BFF→intermediary leg. What the
intermediary does onward to the Platform API is the intermediary's concern.

| `PLATFORM_API_URL` (→ intermediary) | BFF→intermediary leg | Trust config |
|---|---|---|
| `https://gateway:443` | TLS | Trust the **intermediary's** cert via rung 1/2/3 above (not the Platform API's) |
| `http://sidecar:8080` | Plain HTTP | none — valid only when the sidecar/loopback is inside a trusted boundary (e.g. mesh mTLS on `localhost`) |

> With a service mesh doing mTLS, the app-level connection is typically plain
> `http://` to `localhost`, and the sidecar wraps it in mTLS transparently. In
> that case use `http://` and leave `PLATFORM_API_CA_FILE` /
> `PLATFORM_API_TLS_SKIP_VERIFY` unset.

### 3c. Intermediaries on both boundaries

Combine 3a and 3b — the two boundaries remain fully independent. A common
production shape:

```
Browser ──HTTPS──▶ Ingress ──HTTP──▶ BFF ──HTTPS(CA-pinned)──▶ Platform API
```
`BFF_TLS_ENABLED=false`, `PLATFORM_API_URL=https://platform-api:9243`,
`PLATFORM_API_CA_FILE=/etc/ai-workspace/platform-ca.pem`.

---

## Quick reference — env vars by boundary

| Boundary | Component role | Key vars |
|---|---|---|
| Platform API listeners | Server | `HTTP_ENABLED`/`HTTP_PORT`, `HTTPS_ENABLED`/`HTTPS_PORT`/`HTTPS_CERT_DIR` |
| Browser → BFF | Server | `BFF_TLS_ENABLED`, `BFF_TLS_CERT_FILE`, `BFF_TLS_KEY_FILE`, `BFF_TLS_SELF_SIGNED` |
| BFF → Platform API | Client | `PLATFORM_API_URL` (scheme = on/off), `PLATFORM_API_CA_FILE`, `PLATFORM_API_TLS_SKIP_VERIFY` |

Demo mode (`APIP_DEMO_MODE`) is the master switch that decides whether insecure
conveniences (self-signed fallback, skip-verify) are permitted; setting it to
`false` forces every boundary onto real certificates.


Thushani-Jayasekera:
@coderabbit review

portals/ai-workspace/bff/internal/config/config.go
```diff
@@ -71,6 +71,10 @@ type Config struct {
 
 // TLSConfig controls how the BFF terminates TLS.
 type TLSConfig struct {
+	// Enabled terminates TLS on the BFF listener. Defaults to true. Set to false
```
malinthaprasan: What do you mean by Enabled "terminates" TLS?

malinthaprasan: Or does it have to be just Enabled TLS?

malinthaprasan:
There are two ways we can handle TLS enable/disable here:

1. The way you have done it currently - same port, with a config to enable/disable TLS
2. Use two separate listeners for TLS vs plain, and control them independently. They would run on two different ports.

Having two separate ports has some benefits:

- We can use the plain port internally, and open only the TLS port for external communication.
- Someone doing a migration from TLS -> non-TLS (or vice versa) can do it seamlessly, because we can keep both ports open at once temporarily.

And, usually we don't serve non-TLS on a port like 9243  - but it's a minor thing and not a standard, but when we use XX43 ports it usually means TLS is enabled.

Just think about these and decide the best approach.

platform-api/internal/server/server.go
```diff
@@ -761,50 +759,92 @@ func (s *Server) Start(port string, certDir string) error {
...
-			return fmt.Errorf("failed to create cert directory: %v", err)
+			return nil, fmt.Errorf("failed to create cert directory: %v", err)
 		}
 		generatedCert, err := generateSelfSignedCert(certPath, keyPath, s.logger)
 		if err != nil {
 			s.logger.Error("Failed to generate self-signed certificate", "error", err)
-			return fmt.Errorf("failed to generate self-signed certificate: %v", err)
+			return nil, fmt.Errorf("failed to generate self-signed certificate: %v", err)
 		}
 		cert = generatedCert
```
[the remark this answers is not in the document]
Thushani-Jayasekera: This is not required.

platform-api/internal/server/server.go
```diff
@@ -761,50 +759,92 @@ func (s *Server) Start(port string, certDir string) error {
...
-		httputil.WriteJSON(w, http.StatusOK, map[string]any{"status": "ok"})
-	})
-
-	// CreateOrganization TLS configuration
-	tlsConfig := &tls.Config{
+	return &tls.Config{
 		Certificates: []tls.Certificate{cert},
 		MinVersion:   tls.VersionTLS12,
-	}
+	}, nil
```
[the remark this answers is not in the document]
Thushani-Jayasekera: Not required.
Go 1.26's default curve preferences are already [X25519MLKEM768, X25519, P256, P384, P521] (crypto/tls/defaults.go:28), so the hybrid post-quantum key share is offered first even with no CurvePreferences set at all)

platform-api/internal/server/server.go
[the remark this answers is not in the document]
renuka-fernando: We will remove these demo-mode-related, auto-generating credentials and certs in a separate PR. So not required for this now.

platform-api/config/config.go
```diff
@@ -570,9 +615,32 @@ func envToKoanfKey(s string) string {
...
+	// HTTPS / TLS listener.
+	// Legacy TLS_ENABLED / TLS_CERT_DIR and the single-listener PORT are still
+	// honored and map onto the HTTPS listener, which historically served on 9243.
+	case "https_enabled", "tls_enabled":
+		return "https.enabled"
+	case "https_port", "port":
+		return "https.port"
+	case "https_cert_dir", "tls_cert_dir":
+		return "https.cert_dir"
+
```
renuka-fernando, line 633: Maybe we won't need these env var. We can remove them with a separate PR.

platform-api/config/config.go
```diff
@@ -570,9 +615,32 @@ func envToKoanfKey(s string) string {
...
+	// Listener timeouts (apply to both listeners). Values are durations, e.g.
+	// "10s", "2m". 0 disables the timeout.
+	case "timeouts_read_header":
+		return "timeouts.read_header"
+	case "timeouts_read":
+		return "timeouts.read"
+	case "timeouts_write":
+		return "timeouts.write"
+	case "timeouts_idle":
+		return "timeouts.idle"
```
renuka-fernando, line 643: Same for these env vars

platform-api/config/config.go
Thushani-Jayasekera: ack

platform-api/config/config.go
Thushani-Jayasekera: ack

portals/ai-workspace/bff/internal/config/config.go
```diff
@@ -41,9 +42,17 @@ type Config struct {
 	// TLS for the BFF listener
 	TLS TLSConfig
 
-	// Upstream Platform API
-	PlatformAPIURL        string // base URL, e.g. https://platform-api:9243
-	PlatformTLSSkipVerify bool   // accept the Platform API self-signed cert
+	// Upstream Platform API. The http/https scheme of PlatformAPIURL is the single
+	// source of truth for whether the outbound hop uses TLS — there is deliberately
+	// no separate boolean, since that could contradict the URL.
+	PlatformAPIURL string // base URL, e.g. https://platform-api:9243
```
renuka-fernando, line 48: Maybe we can have a separate section for Platform API and under that we can have URL, CA Cert file, skip verify etc.

portals/ai-workspace/bff/internal/config/config.go
```diff
@@ -208,13 +228,15 @@ func Load() (*Config, error) {
...
-			SelfSigned: selfSigned,
+			TerminateTLS: tlsEnabled,
+			SelfSigned:   selfSigned,
 			// Convention matches the legacy entrypoint.sh mount path. buildTLS
 			// falls back to a self-signed cert when these files are absent.
 			CertFile: getenv("BFF_TLS_CERT_FILE", "/etc/ai-workspace/tls/tls.crt"),
 			KeyFile:  getenv("BFF_TLS_KEY_FILE", "/etc/ai-workspace/tls/tls.key"),
 		},
 		PlatformAPIURL:        strings.TrimRight(getenv("PLATFORM_API_URL", ""), "/"),
+		PlatformCAFile:        getenv("PLATFORM_API_CA_FILE", ""),
```
renuka-fernando, line 239: No need to have this as an ENV var

portals/ai-workspace/bff/internal/config/config.go
```diff
@@ -261,6 +283,26 @@ func Load() (*Config, error) {
...
+	// missing/typo'd scheme must fail at startup rather than surface as an opaque
+	// dial error on the first proxied request.
+	u, err := url.Parse(cfg.PlatformAPIURL)
+	if err != nil || (u.Scheme != "http" && u.Scheme != "https") || u.Host == "" {
+		return nil, fmt.Errorf("PLATFORM_API_URL must be an absolute http:// or https:// URL, got %q", cfg.PlatformAPIURL)
+	}
+	// Trust knobs only apply to an https upstream; flag them on a plain-http URL so a
+	// mistaken belief that TLS is in effect is caught early.
+	if u.Scheme == "http" {
+		if cfg.PlatformCAFile != "" || cfg.PlatformTLSSkipVerify {
```
renuka-fernando, line 296: Doesn't this PlatformCAFile is default to OS level CA cert files?

renuka-fernando (approved):

Thushani-Jayasekera:
refactored some in : https://github.com/wso2/api-platform/pull/2602

Issue #2592, opened 2026-07-11
[Task]:  [AI Workspace] `{{ env }}` / `{{ file }}` interpolation in gateway `config.toml`


Thushani-Jayasekera:
### Please select the area the issue is related to

AI Workspace

### Please select the aspect the issue is related to

Aspect/Configuration (Config files, settings, env vars, defaults)

### Description

Related discussion: https://github.com/wso2/api-platform/discussions/1983

PR #2642, opened 2026-07-13
AI Workspace - Configuration Convention & the `/me` Endpoint

Thushani-Jayasekera:
# AI Workspace BFF — Configuration Convention & the `/me` Endpoint

This note captures two related pieces of the AI Workspace BFF: the restructured
configuration model, and the `GET /me` permissions hop the BFF makes to the
Platform API. They are documented together because the `/me` call exists to
resolve a gap that the configuration model deliberately refuses to paper over.

- https://github.com/wso2/api-platform/issues/2592
---

### 1.1 One source of truth: `config.toml`

The BFF loads **all** configuration from a single `config.toml`
(`internal/config/settings.go`, `config.go`). There is no implicit environment
overlay on top of the file — a key takes its value from the environment or a
mounted secret **only** when its own interpolation token says so:

```toml
log_level    = '{{ env "APIP_AIW_LOG_LEVEL" "info" }}'
client_secret = '{{ file "/secrets/ai-workspace/oidc_client_secret" }}'
```

Consequences of this design:

- **Every source a value can come from is visible in the file itself**, rather
  than implied by a naming rule or a precedence ladder. If a key has no token, it
  simply cannot be set from the environment.
- **One mechanism covers both ordinary settings and secrets.** The OIDC client
  secret is resolved the same way as `log_level` — via a `{{ env }}` / `{{ file }}`
  token — so there is no separate secret-loading path.
- **Interpolation fails closed.** An unset `{{ env }}` variable with no default,
  or an unreadable/disallowed `{{ file }}` path, aborts startup rather than
  silently yielding an empty credential.

Tokens are resolved by the shared `common/configinterpolate` library. `{{ file }}`
reads are constrained to an allowlist (`/etc/ai-workspace`, `/secrets/ai-workspace`
by default, overridable via `APIP_CONFIG_FILE_SOURCE_ALLOWLIST`).

### 1.2 Naming convention: `APIP_AIW_`

The AI Workspace namespaces its environment variables with the `APIP_AIW_`
prefix — mirroring the Platform API's `APIP_CP_` and the Developer Portal's
`APIP_DP_`. The convention is: a key's variable is **its dotted config path,
uppercased, dots→underscores, prefixed**.

```
[oidc] client_id   →   APIP_AIW_OIDC_CLIENT_ID
oidc.scope         →   APIP_AIW_OIDC_SCOPE
```

Crucially this is a *convention, not a binding* — the prefix documents the
expected variable name, but the environment only reaches config through the
explicit `{{ env "NAME" }}` token. A token may name any variable.

### 1.3 Flattened, typed access

The decoded TOML tree is flattened to dotted keys (`[platform_api] url` →
`platform_api.url`). Values are read through small typed accessors that decide
what a bad value should do:

- `get` / `getbool` / `getdur` — a **malformed** boolean or duration **fails
  startup** rather than silently reverting to the default (a typo shouldn't
  quietly change behaviour).
- `getint64` — used for defensive size ceilings (e.g. `MaxMeResponseBytes`); an
  unset, malformed, or non-positive value **degrades to the safe default**
  instead of blocking startup, because a size bound is a guard rail, not a
  correctness knob.

`auth_mode` is validated explicitly: anything other than `basic` or `oidc` fails
startup, so a typo can't silently degrade to basic auth.

### 1.4 Browser-safe runtime config (allowlist, not filter)

The SPA reads a subset of config from `window.__RUNTIME_CONFIG__`. The set of
keys that reach the browser is an **allowlist** (`browserSafeKeys` in
`runtime_config.go`), not a filter:

- Only keys on the list ever reach the page. A new server-side key — a secret,
  an upstream URL, a cookie setting — **cannot leak into the browser merely by
  being added to `config.toml`.**
- `[oidc] client_secret` / `client_id` / `authority` are deliberately absent: the
  BFF performs the entire OIDC handshake, so the SPA needs no client identity.
- The SPA's API base URLs are **forced onto the same-origin proxy prefix**, so
  the browser only ever talks to the BFF, never the Platform API directly.

The runtime key spelling (`runtimeKey`) matches the `{{ env }}` token spelling, so
a value has **one name** across `config.toml`, the environment, Vite's
`import.meta.env`, and `window.__RUNTIME_CONFIG__`.

### 1.5 Claim-mapping mirror

`[oidc.claim_mappings]` deliberately mirrors the Platform API's
`[auth.idp.claim_mappings]` **key for key**. Both sides describe the *same* IDP
token, so they must agree — naming them identically makes drift obvious. The same
config entry drives both the BFF's session mapping and the SPA's identity display,
keeping the two layers in sync from one source.

---

### 2.1 What it is

`internal/server/permissions.go` calls the Platform API endpoint:

```
GET /api/portal/v0.9/me
```

It returns the signed-in user's **effective identity and permissions** — the
roles they hold and the scopes those roles grant:

```go
type meResponse struct {
    Roles  []string `json:"roles"`
    Scopes []string `json:"scopes"`
}
```

The BFF issues this call on the user's behalf (forwarding their JWT) and folds the
result into the server-side session.

### 2.2 Why it matters — the problem it solves

The SPA gates **every** UI control on `hasPermission(scope)`. So the session must
carry a scope list. But there is a mode where the token does not provide one:

> **Role validation mode.** The IDP issues a token with **roles and no scope
> claim.** Decoding the scope claim yields nothing — so a fully privileged user
> would be handed an **empty app**, every privileged control hidden, because the
> SPA has no scopes to check against.

The Platform API already knows how to expand roles into scopes — it does this via
`roles.yaml` for its *own* authorization decisions. `GET /me` returns **that same
expansion**. This is the important part:

- The UI is driven by **the exact scope list the API actually enforces**, not by a
  second copy of the role→scope mapping shipped to the browser.
- There is **no duplicated authorization logic**. If the mapping changes on the
  Platform API, the workspace UI follows automatically — nothing to re-ship.

### 2.3 How it behaves (`enrichPermissions`)

```
if the token already carries scopes  → no upstream hop, use them as-is
else                                 → GET /me, adopt returned scopes
```

- **A scope-bearing token needs no `/me` call.** The enrichment only runs to fill
  a gap the token left.
- **Failure degrades safely.** If the `/me` lookup fails, the user is left with
  **no scopes** — the app hides privileged controls rather than offering actions
  the API would then reject. It logs a warning and continues; it does not block
  login.
- **Role label fallback.** In role mode there's no `platform_role` claim, so the
  displayed role label falls back to the first IDP role from `/me`.

### 2.4 Defensive bound: `MaxMeResponseBytes`

The `/me` payload is a small identity record, so the response read is capped
(`platform_api.max_me_response_bytes`, default **1 MiB**) via `io.LimitReader`.
Anything larger is treated as an upstream fault, not something to buffer into
memory. Per §1.3, a non-positive or malformed configured limit falls back to the
safe default rather than failing startup.

---

## 3. How the two connect

| Concern | Config model says | `/me` does |
|---|---|---|
| Where does authoritative permission data live? | Not duplicated to the browser (allowlist keeps role→scope mapping server-side) | Fetches the API's own expansion at request time |
| What if the token lacks scopes? | The BFF won't fabricate them from a shipped mapping | Resolves them from the enforcing service |
| What if resolution fails? | Fail-closed philosophy | Degrades to *no* scopes (hide, don't offer) |
| Defensive limits | `getint64` degrades to safe default | Caps the response read at `MaxMeResponseBytes` |

The through-line: **a single authoritative source, resolved at the layer that
enforces it, never re-implemented in the browser.** The configuration model keeps
authorization data off the client; `/me` supplies it on demand from the service
that owns it.


portals/ai-workspace/docker-compose.yaml
```diff
@@ -107,23 +126,29 @@ services:
...
+      # These resolve the matching '{{ env "APIP_AIW_..." }}' tokens in
+      # configs/config.toml. Only deployment-specific values live here — everything
+      # else belongs in that file.
+      #
+      # The BFF talks to the Platform API over the internal compose network, whose
+      # certificate is self-signed.
+      - APIP_AIW_PLATFORM_API_URL=https://platform-api:9243
+      - APIP_AIW_PLATFORM_API_TLS_SKIP_VERIFY=true
+      - APIP_DEMO_MODE=${APIP_DEMO_MODE:-true}
+      # ── OIDC — to delegate login to any OIDC provider, set auth_mode = "oidc" and
```
Thushani-Jayasekera, line 138: all these configs from the dockerfile will be removed from the next effort and .env file mount will be used

renuka-fernando (approved):

Krishanx92 (approved):

PR #2649, opened 2026-07-13
[PlatformAPI] Refactor gateway version translation and cover all API kinds

### LINKED ISSUE #2492: [Task]: Improve platform to gateway version related artifact transformation implementation

Area/Management (Management API or Management Portal UI)

### Description

Improve the implementation in platform-api's newly introduced `deploymenttransform` package to transform artifact to gateway compatible version.
- Instead of `policies.go` add logic against kind instead
- Introduce packaging for each gateway version
- For each version specific kind transformation, take as input the to and from version and write transformations ensuring scalability

ashera96:
## Purpose

- Resolves: https://github.com/wso2/api-platform/issues/2492

Reworks the platform-to-gateway artifact translation layer so it is organized by kind and by version (scalable), and extends it to **every** deployable kind.

Previously, only LLM providers and proxies were version-translated; deploying a REST API, MCP proxy, or WebSub/WebBroker API to a gateway older than `1.2.0` shipped `apiVersion: .../v1` and the gateway rejected it with *"Unsupported API version (must be '.../v1alpha1')"*. This folds in the behavior of the earlier PR #2547 within the new structure.

---

### `internal/deploymenttransform` → `internal/gatewaytranslator` (go-standard package layout)

* **`translator.go`** — Entry point `Translate(kind, sourceDataVersion, targetDataVersion, artifact)`: normalizes the artifact up to the gateway-latest shape, then down-converts only when the target gateway is older.
* **`version.go`** — `Version` (gateway semver); `GTE` renamed to the more readable `AtLeast` + `Below`.
* **`dataversion.go`** — `PlatformDataVersion` / `GatewayDataVersion`; the single gateway-wide apiVersion boundary `MinGatewayV1Version = "1.2.0"`; `ComputeDataVersion` ported from `gateway-controller`.
* **`normalizer/`** — Up-converts a stored artifact to gateway-latest (LLM: flat `policies` → `globalPolicies`/`operationPolicies`, idempotent; identity for all other kinds).
* **`versiontranslator/`** — A pure router by target data version → **`versiontranslator/v1alpha1/`**, a self-contained producer.
    * The `v1alpha1` apiVersion swap is applied to **every** kind by default via a `deploymentArtifact` interface.
    * A per-kind shape handler is registered only where a kind's data shape actually changes (LLM).
    * Any unlisted/future kind still down-converts its apiVersion — acting as the regression guard for this class of bug.

### Data Version Stamping
The `data_version` column existed on the schema but no Go code wrote it.
* Now stamped on Create/Update for the six kind-bearing entities via `ComputeDataVersion` (LLM provider/proxy → `1.1`, REST/MCP/WebSub/WebBroker → `1.0`; pre-set values preserved).
* Read back into the models and used as `sourceDataVersion` at deploy time.

### Call-site Wiring
`Translate` is invoked in all six CP "deploy from current" paths:
* LLM provider/proxy (`llm_deployment.go`)
* REST/WebSub (`deployment.go`)
* MCP (`mcp_deployment.go`)
* WebSub/WebBroker (`plugins/eventgateway/service/*_deployment.go`)

---

## Testing

* **Unit**: `internal/gatewaytranslator/...` — Matrix over all kinds × gateway `1.2.0` (v1) / `1.1.0` (v1alpha1) × source `1.0` / `1.1`, plus `ParseVersion`/`AtLeast`/`Below`, `ComputeDataVersion`, normalizer parity with the existing service-layer migration, and an interface-implementation guard.
* **Integration** (`//go:build integration`, sqlite): Deploy REST / MCP / LLM provider (current 1.1 split-source and legacy 1.0 flat-source) to gateways seeded at `1.1.0` and `1.2.0`; assert the stored artifact's `apiVersion`, policy shape, and `data_version` stamping.

---

## Known limitations / follow-ups

* **Redeploy from an existing deployment (`Base != "current"`)**: Reuses the stored artifact content and is not re-translated — so promoting a snapshot built for one gateway to a different-versioned gateway can ship the wrong `apiVersion`. This limitation pre-dates this PR (the old layer only handled the "current" branch); flagged for a follow-up.
* `data_version` stamping is scoped to the six kind-bearing entities that feed the translator; the remaining non-artifact tables (and `llm_provider_templates`) keep the SQL `DEFAULT '1.0'`.
* `ComputeDataVersion`'s per-kind minor map now diverges from `gateway-controller`'s (LLM `1` vs `0`); the two are independent per-store markers, but the intent should be reconciled.


ashera96:
@coderabbitai please review and approve

platform-api/internal/service/llm_deployment.go
```diff
@@ -199,10 +199,12 @@ func (s *LLMProviderDeploymentService) DeployLLMProvider(providerID string, req
 		if err != nil {
 			return nil, fmt.Errorf("failed to generate LLM provider deployment YAML: %w", err)
 		}
-		target := deploymenttransform.ParseVersion(gateway.Version)
-		if err := deploymenttransform.Default().Transform(
+		sourceDataVersion := gatewaytranslator.PlatformDataVersion(provider.DataVersion)
+		targetDataVersion := gatewaytranslator.GatewayDataVersionForGateway(gateway.Version)
```
renuka-fernando, line 203: Rename `targetDataVersion` -> targetGatewayVersion

platform-api/internal/gatewaytranslator/translator.go
```diff
@@ -0,0 +1,60 @@
...
+//   - kind is the artifact kind (e.g. constants.LLMProvider).
+//   - sourceDataVersion is the platform data version the artifact was
+//     generated from.
+//   - targetDataVersion is the gateway data version the target gateway
+//     accepts (see TargetGatewayDataVersion).
+func Translate(kind string, sourceDataVersion PlatformDataVersion, targetDataVersion GatewayDataVersion, artifact any) error {
+	if err := normalizer.Normalize(kind, string(sourceDataVersion), artifact); err != nil {
+		return fmt.Errorf("gatewaytranslator: normalize %q: %w", kind, err)
+	}
+	if targetDataVersion == GatewayDataVersionV1 {
```
renuka-fernando, line 53: Rename `GatewayDataVersionV1` -> `CurrentLatestGatewayVersion`

Should be something like this

if targetDataVersion >= CurrentLatestGatewayVersion {

platform-api/internal/gatewaytranslator/versiontranslator/versiontranslator.go
```diff
@@ -0,0 +1,59 @@
...
+// KnownKinds returns the artifact kinds this translator is aware of.
+func KnownKinds() []string {
+	out := make([]string, len(knownKinds))
+	copy(out, knownKinds)
+	return out
+}
+
+// DownConvert produces, in place, the shape targetDataVersion expects for
+// artifact (already in gateway-latest shape).
+func DownConvert(kind string, targetDataVersion string, artifact any) error {
```
renuka-fernando, line 52:
```suggestion
func DownConvert(kind string, targetGatewayVersion string, artifact any) error {
```

platform-api/internal/gatewaytranslator/versiontranslator/v1alpha1/llm.go
```diff
@@ -25,74 +25,53 @@ import (
...
-	policyNameLLMCost              = "llm-cost"
+	policyNameLLMCost               = "llm-cost"
 	policyNameLLMCostBasedRateLimit = "llm-cost-based-ratelimit"
 )
 
 func init() {
-	minVer := ParseVersion(MinSplitPoliciesVersion)
-	isOldGateway := func(target Version) bool { return !target.GTE(minVer) }
+	shapeHandlers[constants.LLMProvider] = downConvertLLMProvider
+	shapeHandlers[constants.LLMProxy] = downConvertLLMProxy
```
renuka-fernando, line 37: Let's create separate llm_provider.gol, llm_proxy.go

renuka-fernando (approved):

PR #2702, opened 2026-07-15
Fix MCP proxy upstream URL manipulation

### LINKED ISSUE #2673: [Bug]: AI Workspace MCP Proxy not working as expected

AI Workspace

### Description

1. MCP proxy invocation not working as expected, fails to connect with a 404.
2. AI workspace must change: drop the .replace(/\/mcp$/, '') and send the validated URL as-is. This is the same URL that fetch-server-info just validated, so it's guaranteed to be the working endpoint.
3. platform-api refetch: remove the ensureMCPEndpointURL append (service/mcp.go:405) — once the workspace stores full URLs it's a no-op for /mcp-suffixed servers but still breaks custom-path servers (e.g. /api/v1/mcp-server would get /mcp appended on refetch).
4.  Existing data: proxies already created via the workspace have stripped URLs in the DB. Since the workspace only ever stripped a trailing /mcp, a one-time migration re-appending /mcp to stored MCP upstream URLs is safe and restores correctness. Relying on the gateway's global legacy flag instead would keep old proxies working but break newly created ones (the flag isn't per-API), so migration is the cleaner path.

### Steps to Reproduce

1. Create MCP Proxy via the provided sample URL
2. Deploy and try to connect to the deployed MCP Server, you should then be able to notice the connection issue resulting in a 404.

Severity/Blocker (Core functionality is broken and consumer is blocked. Need immediate attention)

_No response_

ashera96:
## Purpose

- Resolves: https://github.com/wso2/api-platform/issues/2673
 
* **Follow-up to #2565**: Aligns with the change that modified the gateway default to forward MCP `/mcp` routes to exactly the configured upstream path instead of auto-appending `/mcp`. The stored upstream is now expected to be the full MCP endpoint URL.
* **The Problem**: Both `ai-workspace` and `platform-api` still manipulated the URL against the old contract. As a result, the stored/deployed upstream no longer matched the real MCP endpoint, causing a `404` error on connect (including via the create wizard's *"Try with Sample URL"*).
* **The Fix**: Preserve the provided backend URL end-to-end and never manipulate it with a `/mcp` substring.
    * **`ai-workspace`**: Stop stripping a trailing `/mcp` from the validated endpoint URL before storing it as the upstream — send the exact URL `fetch-server-info` already validated.
    * **`platform-api`**: Stop re-appending `/mcp` to the stored upstream URL on refetch (`FetchServerInfo`). Use it verbatim, consistent with Create/Update and the gateway deployment YAML builder, which already use it as-is. Removed the now-dead `ensureMCPEndpointURL` helper.

> **Note**: Migration of existing proxies whose stored URLs were already stripped of `/mcp` is out of scope for this PR and will be handled separately.

### DIFF
--- a/platform-api/internal/service/mcp.go
+++ b/platform-api/internal/service/mcp.go
@@ -24,8 +24,6 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
-	"net/url"
-	"strings"
 	"time"
 
 	"github.com/wso2/api-platform/platform-api/api"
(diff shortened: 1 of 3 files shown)


ashera96:
@coderabbitai please review

Krishanx92 (approved):

renuka-fernando (approved):

Issue #2621, opened 2026-07-13
[Bug]: Claude Code responses buffer when analytics is enabled


SavinduDimal:
### Please select the area the issue is related to

Gateway

### Please select the aspect the issue is related to

Aspect/API (API backends, definitions, contracts, interfaces, OpenAPI), Aspect/AI (AI/LLM integration, MCP, AI readiness)

### Description

When claude code traffic is routed through WSO2 AI Gateway, responses get buffered in the gateway when analytics is enabled (even without the LLM Cost policy). When analytics is disabled this works without any issues.

### Steps to Reproduce

Configure claude code traffic through an AI Gateway as mentioned in doc [1] and try to use it.

[1] - https://wso2.com/api-platform/docs/guides/ai-and-mcp/ai-coding-assistants/claude-code-configuration-with-ai-gateway/

### Severity Level of the Issue

Severity/Major (Important functionality is broken. Should be prioritized. Doesn't need immediate attention)

SavinduDimal:
Fixed through : https://github.com/wso2/api-platform/pull/2878

PR #2703, opened 2026-07-15
Implement stateful per-chunk streaming compression

SavinduDimal:
## Purpose

- $Subject
- Related Issues: https://github.com/wso2/api-platform/issues/2621

### Issue 1: Analytics + Streaming

- Commit https://github.com/wso2/api-platform/pull/2703/changes/70326d66e90093a96ef5d177476fad47b1fce622 fixes the initial streaming not working while analytics is enabled by introducing changes to the chunk recompressor. Previously we re-compressed each streaming chunk by calling recompressBody, which spins up a fresh writer and Close()s it every chunk. so each chunk became its own independent, self-contained gzip/brotli member. `streamCompressor` keeps one writer alive for the life of the stream, flushing (`Z_SYNC_FLUSH`) after each chunk and closing only at EndOfStream, producing a single continuous member the client decodes in full.

### Issue 2: Timeout error
- Then found a new issue when we work with large responses, response flow got stalled and faced timeout errors. 
- The old streamDecompressor used an io.Pipe + decoder goroutine + a bounded output channel (cap 64). On long/large responses the channel filled, the decoder blocked on the send, stopped reading the pipe, and the next FeedChunk blocked forever on pw.Write — a back-pressure deadlock. This is the mid-stream stall (stream would forward ~part of the body, then go silent until the client timed out). It only showed up on large tasks, which is why short prompts worked and the original 2-chunk tests missed it.

- A minimal patch could stop the deadlock, but the io.Pipe design has a deeper flaw: pw.Write returning tells you bytes were copied to the reader, not decoded, so FeedChunk had no reliable signal for "all output for this chunk is ready" — it relied on a racy runtime.Gosched() + best-effort drain. Getting a correct signal on top of the pipe means reimplementing exactly what the refactor introduces, so the pipe becomes dead weight.

- The refactor replaces io.Pipe + channel + errChan + Gosched with a single sync.Cond + a feederReader that records when the decoder has consumed all fed input and parked. FeedChunk now returns exactly when every decodable byte is available — deadlock-free and deterministic — and ends up with fewer moving parts than before.

### DIFF
--- a/gateway/gateway-runtime/policy-engine/internal/kernel/decompression.go
+++ b/gateway/gateway-runtime/policy-engine/internal/kernel/decompression.go
@@ -23,7 +23,7 @@ import (
 	"compress/gzip"
 	"fmt"
 	"io"
-	"runtime"
+	"sync"
 
 	"github.com/andybalholm/brotli"
 )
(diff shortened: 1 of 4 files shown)


renuka-fernando:
Related PR: https://github.com/wso2/api-platform/pull/2878

SavinduDimal:
Closing this PR since the issues are not reproducible after PR: https://github.com/wso2/api-platform/pull/2878

PR #2778, opened 2026-07-21
refactor(ai-workspace): harden SPA static path containment check

renuka-fernando:
## Purpose

A static-analysis scan flagged the `os.Stat` call in `portals/ai-workspace/bff/internal/server/static.go` as a High severity path/directory traversal sink. Investigation confirmed this is a false positive — `filepath.Clean` on the root-absolute `r.URL.Path` collapses all escaping `..` before the join, so `os.Stat` can never resolve outside `staticDir`. However, the existing containment guard relied on a bare `strings.HasPrefix` check, which is a weak pattern flagged by the project's `file-access.md` rule and left the safety of the handler as an emergent property rather than an explicit assertion. The `spaHandler` also had no unit test coverage at all.

## Goals

Make the SPA static handler's traversal containment explicit and correct, satisfy the `file-access.md` separator-suffixed prefix rule, silence the scanner finding as a documented false positive with hardened code, and lock in the behaviour with the handler's first unit tests.

## Approach

- Replace the bare `strings.HasPrefix(full, filepath.Clean(staticDir))` guard with a separator-suffixed containment assertion so a sibling directory sharing the root's name prefix (e.g. `/var/www` vs `/var/www-secret`) cannot pass.
- Explicitly allow the exact-root case (`full == root`) so a request for `/` still falls through to serving `index.html`.
- Add a comment documenting the security intent so containment is no longer an implicit side effect of `Clean` on a rooted path.
- Add `static_test.go` — the first unit test for `spaHandler` — covering asset serving, index no-store fallback, and traversal containment.

## User stories

N/A

## Documentation

N/A — internal hardening of an existing request handler with no behavioural or API change.

## Automation tests

 - Unit tests
   > New `static_test.go` (137 lines), the first coverage for `spaHandler`: `TestSPAHandlerServesHashedAssetCacheable` (asset served 200, cacheable), `TestSPAHandlerServesIndexNoStore` (`/` and SPA client routes fall back to `index.html` with `no-store`), and `TestSPAHandlerPathTraversalContained` (8 raw and URL-encoded `..` payloads assert a file outside `staticDir` is never served — the regression test for this finding). All passing.
 - Integration tests
   > N/A

## Security checks

 - Ran FindSecurityBugs plugin and verified report? no (Go component; the flagged finding was reviewed manually, confirmed a false positive, then hardened and covered by a regression test)

## Samples

N/A

## Related PRs

N/A

## Test environment

Go 1.26 (ai-workspace BFF component); `go test ./internal/server/` passes and LSP diagnostics report no type or compile errors.

## Remarks

The original code was **not** an exploitable traversal — this is defense-in-depth hardening that makes the containment guarantee explicit, plus regression tests. The scanner finding on the `os.Stat` line should be marked a false positive.

### DIFF
--- a/portals/ai-workspace/bff/internal/server/static.go
+++ b/portals/ai-workspace/bff/internal/server/static.go
@@ -36,8 +36,12 @@ func spaHandler(staticDir string) http.Handler {
 		clean := filepath.Clean(r.URL.Path)
 		full := filepath.Join(staticDir, clean)
 
-		// Prevent path traversal outside staticDir.
-		if !strings.HasPrefix(full, filepath.Clean(staticDir)) {
+		// Prevent path traversal outside staticDir. Assert containment against
+		// the root with a trailing separator so a sibling dir whose name merely
+		// shares the prefix (e.g. root "/var/www" vs "/var/www-secret") cannot
+		// pass a bare HasPrefix check.
+		root := filepath.Clean(staticDir)
+		if full != root && !strings.HasPrefix(full, root+string(filepath.Separator)) {
 			http.NotFound(w, r)
 			return
 		}
--- a/portals/ai-workspace/bff/internal/server/static_test.go
+++ b/portals/ai-workspace/bff/internal/server/static_test.go
@@ -1,0 +1,137 @@
+/*
+ * Copyright (c) 2026, WSO2 LLC. (https://www.wso2.com).
+ *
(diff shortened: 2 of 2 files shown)


renuka-fernando:
@coderabbitai review

renuka-fernando:
@coderabbitai review

Thushani-Jayasekera (approved):

PR #2781, opened 2026-07-22
Prevent deletion of custom policies used by LLM providers

### LINKED ISSUE #2777: [Bug]: Custom Policies Delete issue

Platform API

### Description

A custom policy can be deleted even when it is currently used by an LLM provider.

The custom policy is first synchronized through the gateway manifest and then added to an LLM provider. After the provider is deployed and successfully invoked, the custom policy can still be deleted. This behavior is incorrect because deleting a policy that is actively referenced by a deployed LLM provider may leave the provider configuration in an invalid or inconsistent state.

### Steps to Reproduce

1. Synchronize a custom policy using the gateway manifest.
2. Create an LLM provider.
3. Add the synchronized custom policy to the LLM provider.
4. Deploy the LLM provider.
5. Invoke the deployed provider and confirm that it works successfully.
6. Navigate to the custom policies list.
7. Attempt to delete the custom policy used by the LLM provider.

Severity/Blocker (Core functionality is broken and consumer is blocked. Need immediate attention)

_No response_

lahiruudayakumara:
## Purpose

Custom policies referenced by LLM providers can currently be deleted, leaving deployed provider configurations invalid or inconsistent.

Resolves #2777.

## Goals

- Prevent deletion of custom policies that are actively used by LLM providers.
- Keep custom-policy usage records synchronized when providers are created or updated.
- Remove stale usage records when policies are removed from a provider.

## Approach

- Track custom-policy references from LLM provider global, operation, and LLM-specific policies.
- Resolve policy references using organization, case-insensitive name, and major version.
- Reuse the existing custom-policy usage registry and deletion guard.
- Rely on the existing artifact foreign-key cascade to remove usage records when an LLM provider is deleted.
- Add unit coverage for usage insertion, retention, and removal.

## User stories

As an API platform administrator, I cannot delete a custom policy while an LLM provider references it, preventing invalid deployed configurations.

## Documentation

N/A. This fixes existing lifecycle behavior and does not introduce or change a public API.

## Automation tests

- Unit tests
  - Added coverage for global, operation, and LLM-specific policy references.
  - Covered case-insensitive policy-name resolution.
  - Covered major-version matching.
  - Covered insertion of new usages, retention of existing usages, and removal of stale usages.
- Integration tests
  - Not added; the change reuses the existing custom-policy usage registry and deletion guard.

Validation performed:

```text
go build ./cmd/main.go
make test

### DIFF
--- a/platform-api/internal/repository/custom_policy.go
+++ b/platform-api/internal/repository/custom_policy.go
@@ -232,6 +232,26 @@ func (r *CustomPolicyRepo) DeleteCustomPolicyUsage(policyUUID, apiUUID string) e
 	return err
 }
 
+// replaceCustomPolicyUsagesTx atomically replaces every custom-policy usage for
+// an artifact. It is used from artifact persistence transactions so the artifact
+// configuration and its deletion guards cannot diverge.
+func replaceCustomPolicyUsagesTx(tx *sql.Tx, db *database.DB, artifactUUID string, policyUUIDs []string) error {
+	if _, err := tx.Exec(db.Rebind(`DELETE FROM gateway_custom_policy_usages WHERE artifact_uuid = ?`), artifactUUID); err != nil {
+		return err
+	}
+	seen := make(map[string]struct{}, len(policyUUIDs))
+	for _, policyUUID := range policyUUIDs {
+		if _, exists := seen[policyUUID]; exists {
+			continue
+		}
+		seen[policyUUID] = struct{}{}
+		if _, err := tx.Exec(db.Rebind(`INSERT INTO gateway_custom_policy_usages (policy_uuid, artifact_uuid) VALUES (?, ?)`), policyUUID, artifactUUID); err != nil {
+			return err
+		}
+	}
+	return nil
+}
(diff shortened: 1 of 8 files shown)


Thushani-Jayasekera (approved):
LGTM

Thushani-Jayasekera:
A couple of follow-up improvements we can address [separately](https://github.com/wso2/api-platform/issues/2802):
- We should remove the conditional around reconcilePolicyUsages and always maintain policy usage for LLM providers, consistent with how we handle APIs. Not storing these usages was the original bug, so keeping the reconciliation unconditional ensures consistent behavior.
- The current insert loop executes one tx.Exec per policy UUID, resulting in N database round-trips within the transaction. Since the policy UUIDs are already deduplicated upfront, this could be optimized into a single multi-row INSERT to reduce transaction overhead.

These are both non-blocking improvements, and they're already present in the refreshCustomPolicyUsages method. We can address them in a follow-up, so I'm okay to merge this as-is.

Thanks for the contribution!

PR #2814, opened 2026-07-23
Adding external plugin support

malinthaprasan:
## Purpose

`platform-api` can only be extended by editing it. There is no supported way for a separate module (e.g. `api-cloud`) to run the platform server and add its own REST APIs, scopes, and middleware without forking or patching platform-api sources.

https://github.com/wso2/api-platform/issues/2831

## Goals

Make `platform-api` an importable library with a stable public extension contract, so a wrapper module can run and extend the full server via import only — no changes to platform-api.

## Approach

Two-tier plugin model:

- **`pdk/`** (external tier) — the public contract: `pdk.Plugin`, `pdk.Deps` (capabilities as interfaces over public `api`/`config` types), `pdk.Middleware` with two fixed chain positions, and `pdk/middleware` for reading the authenticated org from the request. No internal types leak.
- **`internal/plugin`** (internal tier) — unchanged full access for in-tree plugins; keeps `AuthSkipPathProvider`, which `pdk` deliberately has no equivalent of.
- **`platform/`** — the façade a wrapper imports to run the server: `platform.New(WithConfig…, WithPlugin(…))` + `Run()`.

Server wiring:

- `StartPlatformAPIServer(cfg, logger, internalPlugins, externalPlugins)` — external plugins are wrapped so both tiers run through one init/shutdown loop; only the `Deps` each receives differs.
- Removed the global plugin registry (`plugin.Register`/`All`). Built-ins now come from `internal/builtins.Plugins()`, build-tag selected, passed explicitly by `cmd/main.go` (keeps `go build ./cmd/main.go` working).

## User stories

As a wrapper-module developer, I can run the platform-api server and add my own authenticated, scope-enforced REST APIs and middleware without modifying platform-api.

## Documentation

N/A for product docs — this is an internal extension surface. It is documented in the package docs on `pdk` and `platform`.

## Automation tests

- Unit tests
  > `internal/server/plugins_test.go`, `internal/server/external_plugin_test.go` — scope merge into the shared registry; startup aborts on empty / malformed / scope-less specs; skip-path validation; external plugins receive `pdk.Deps` (never `plugin.Deps`) and cannot contribute skip paths; middleware position wiring; shutdown-on-failed-startup.
- Integration tests
  > N/A — no runtime behaviour change for existing routes; existing server tests cover the unchanged chain.

## Security checks

- Ran FindSecurityBugs plugin and verified report? N/A (Go)

## Samples

N/A

## Related PRs

N/A

## Test environment

Go 1.26, macOS / Linux — `go build ./...`, `go vet`, `go test ./internal/server/... ./platform/...`


malinthaprasan:
@coderabbitai please review

malinthaprasan:
@coderabbitai - these external plugins are server wide, and not user defined. So there's no risk of cross org access etc. Can you revisit all the comments you made?

platform-api/pdk/plugin.go
```diff
@@ -0,0 +1,65 @@
...
+	Shutdown(ctx context.Context) error
+}
+
+// AuthSkipPathProvider is an optional interface a Plugin may also implement to
+// declare public (unauthenticated) path prefixes. The server appends them to the
+// auth skip-path list before the auth middleware is built. Keep prefixes narrow
+// and specific — this is an auth-bypass surface (GO-AUTH-004).
+type AuthSkipPathProvider interface {
+	AuthSkipPaths() []string
+}
```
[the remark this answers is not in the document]
malinthaprasan: https://github.com/wso2/api-platform/pull/2814#issuecomment-5057726011

platform-api/pdk/middleware.go
```diff
@@ -0,0 +1,62 @@
...
+// enforcer).
+type ChainPosition int
+
+const (
+	// BeforePlatformChain is outermost: it runs before CORS and auth, on every
+	// request, with NO authenticated identity in the context yet. Good for request
+	// IDs, tracing, panic recovery, IP allow-lists. It cannot mark a request as
+	// authenticated — the platform's auth always runs after it and is the only
+	// thing that sets identity, so GO-AUTH-005 still holds.
+	BeforePlatformChain ChainPosition = iota
```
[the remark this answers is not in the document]
malinthaprasan: https://github.com/wso2/api-platform/pull/2814#issuecomment-5057726011

platform-api/pdk/deps.go
```diff
@@ -0,0 +1,64 @@
...
+
+	// GetGateway returns a single gateway by id within an organization (Read).
+	GetGateway(gatewayID, orgID string) (*api.GatewayResponse, error)
+
+	// UpdateGateway updates a gateway within an organization (Update).
+	UpdateGateway(gatewayID, orgID, updatedBy string, req *api.GatewayResponse) (*api.GatewayResponse, error)
+
+	// DeleteGateway removes a gateway within an organization (Delete).
+	DeleteGateway(gatewayID, orgID, deletedBy string) error
+}
```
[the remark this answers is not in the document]
malinthaprasan: https://github.com/wso2/api-platform/pull/2814#issuecomment-5057726011

platform-api/pdk/deps.go
```diff
@@ -0,0 +1,64 @@
...
+// adapter code. The assignment itself is the compile-time contract check: if a
+// signature drifts, the server stops building.
+type Deps struct {
+	Gateways Gateways
+	// add more capability groups as external plugins need them
+	// (APIs, Subscriptions, Applications, Projects, Organizations, LLM, MCP, …)
+
+	Config *config.Server
+	Logger *slog.Logger
+}
```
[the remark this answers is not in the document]
malinthaprasan: External plugins are trusted so this is not a problem

platform-api/pdk/plugin.go
```diff
@@ -0,0 +1,65 @@
...
+	Init(deps *Deps) error
+
+	// RegisterRoutes mounts the plugin's HTTP routes on the shared mux. Only
+	// called after Init has succeeded.
+	RegisterRoutes(mux *http.ServeMux)
+
+	// OpenAPISpec returns the plugin's OpenAPI 3.x YAML bytes, merged into the
+	// platform scope registry to enforce per-route scopes. Return nil when the
+	// plugin needs no scope enforcement.
+	OpenAPISpec() []byte
```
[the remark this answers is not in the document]
malinthaprasan: Fixed [Mandate a non-empty OpenAPI spec from every plugin](https://github.com/wso2/api-platform/pull/2814/commits/fcd83b73884fc7a970423434c978608f6fbc5cf5)

platform-api/internal/server/server.go
```diff
@@ -414,7 +427,35 @@ func StartPlatformAPIServer(cfg *config.Server, slogger *slog.Logger) (*Server,
...
+
+	// Plugin-contributed middleware, collected in the loop below and spliced into
+	// the chain when it is built: preChain outermost (before CORS/auth), postChain
+	// innermost (after scope enforcement, before the mux).
+	var preChain, postChain []pdk.Middleware
+
+	for _, p := range plugins {
 		if err := p.Init(pluginDeps); err != nil {
 			return nil, fmt.Errorf("plugin %q failed to initialize: %w", p.Name(), err)
 		}
```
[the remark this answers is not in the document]
malinthaprasan: Fixed [Shut down initialized plugins when startup aborts](https://github.com/wso2/api-platform/pull/2814/commits/8ad84bb0ba49e3647dec2e52698fde83187502dc)

platform-api/internal/server/plugins.go
```diff
@@ -0,0 +1,186 @@
...
+		pluginRegistry, regErr := middleware.LoadScopeRegistryFromBytes(spec)
+		if regErr != nil {
+			return fail("plugin %q OpenAPI spec failed to load into the scope registry: %w", p.Name(), regErr)
+		}
+		if pluginRegistry.Len() == 0 {
+			return fail("plugin %q OpenAPI spec declares no scope requirements; every route a plugin registers must declare its required scopes", p.Name())
+		}
+		scopeRegistry.Merge(pluginRegistry)
+
+		p.RegisterRoutes(mux)
```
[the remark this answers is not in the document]
malinthaprasan: I'd prefer to keeping this simple. A plugin may register some routes that are not advertised via the OpenAPI (If the OpenAPI is used in a gateway, these routes won't be exposed as well)

platform-api/internal/server/plugins.go
```diff
@@ -0,0 +1,186 @@
...
+	case path == "":
+		return fmt.Errorf("path is empty, which matches every request")
+	case path == "/":
+		return fmt.Errorf("path is the root prefix, which matches every request")
+	case !strings.HasPrefix(path, "/"):
+		return fmt.Errorf("path must start with %q", "/")
+	case strings.Contains(path, ".."):
+		return fmt.Errorf("path must not contain %q", "..")
+	}
+	return nil
```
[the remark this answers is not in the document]
malinthaprasan: This is removed. [Remove the auth skip-path hook from the external plugin tier](https://github.com/wso2/api-platform/pull/2814/commits/272ec3ead58926a53700a5ba1322b0ff57e232bb)

platform-api/internal/server/plugins.go
```diff
@@ -0,0 +1,186 @@
...
+			return fail("plugin %q returned an empty OpenAPI spec; a spec declaring each route's scopes is required", p.Name())
+		}
+		pluginRegistry, regErr := middleware.LoadScopeRegistryFromBytes(spec)
+		if regErr != nil {
+			return fail("plugin %q OpenAPI spec failed to load into the scope registry: %w", p.Name(), regErr)
+		}
+		if pluginRegistry.Len() == 0 {
+			return fail("plugin %q OpenAPI spec declares no scope requirements; every route a plugin registers must declare its required scopes", p.Name())
+		}
+		scopeRegistry.Merge(pluginRegistry)
```
AnuGayan, line 109: 🟠 **Major — a plugin can override / weaken scopes on core routes.** `ScopeRegistry.Merge` overwrites on key conflict, and the core registry is populated (server.go:367) before this merge — so a plugin whose spec declares a key that collides with a first-party route (e.g. `GET:/api/v1/gateways`) silently **replaces** the core scope requirement with its own, weaker one. There's no conflict detection. Since external / wrapper plugins are the target of this PR, this is a privilege-escalation surface on core routes.

Suggest a strict merge that aborts startup when a plugin key already exists in the core registry:
```go
if _, exists := r.scopes[k]; exists {
    return fmt.Errorf("scope registry: plugin redeclares existing route %q", k)
}
```

platform-api/internal/server/server.go
```diff
@@ -414,22 +427,45 @@ func StartPlatformAPIServer(cfg *config.Server, slogger *slog.Logger) (*Server,
...
+	// here works because they satisfy the pdk interfaces by shape, and the
+	// assignment itself is the compile-time contract check: if a service method
+	// signature drifts from the pdk interface, this stops building.
+	pdkDeps := &pdk.Deps{
+		Gateways: gatewayService,
+		Config:   cfg,
+		Logger:   slogger,
+	}
+
+	wiring, err := initPlugins(slogger, mux, scopeRegistry, pluginDeps, pdkDeps, internalPlugins, externalPlugins)
```
AnuGayan, line 436: 🟠 **Major — plugin scopes are invisible to `roles.yaml` validation in IDP role mode.** `loadRoleScopeMap` / `ValidateRoleScopeMap` runs at **server.go:375**, *before* this call merges plugin scopes into the registry. `ValidateRoleScopeMap` rejects any role referencing a scope not in the registry, so a wrapper that maps a role to one of its own plugin scopes fails startup with `references unknown scope`. That breaks the feature's own goal in `validationMode: role` — it only works in `scope` mode, where scopes come straight from the token.

Fix: merge the plugin scope registries **before** `loadRoleScopeMap`, or re-run role-map validation after `initPlugins` against the combined registry. (Note the ordering interaction with the strict-merge fix on the Merge call.)

platform-api/pdk/plugin.go
```diff
@@ -0,0 +1,62 @@
...
+	// through the platform's authentication and scope chain.
+	RegisterRoutes(mux *http.ServeMux)
+
+	// OpenAPISpec returns the plugin's OpenAPI 3.x YAML bytes, merged into the
+	// platform scope registry to enforce per-route scopes. It is mandatory:
+	// returning empty bytes, bytes the registry loader rejects, or a spec that
+	// declares no scope requirement on any operation all abort startup. Every
+	// route the plugin mounts must appear in this document with its required
+	// scopes, so no extension route can be served without an explicit scope
+	// requirement (GO-AUTH-007).
```
AnuGayan: 🟠 **Major — this docstring overstates the GO-AUTH-007 guarantee.** It states *every route the plugin registers must appear in this document with its required scopes, so no extension route can be served without an explicit scope requirement.* The actual guard (plugins.go:111, `pluginRegistry.Len() == 0`) only proves **≥1** scoped operation exists — it does not verify correspondence between routes passed to `RegisterRoutes(mux)` and spec entries. A plugin registering 5 routes but scoping only 1 passes startup; the other 4 fall through `ScopeEnforcer` **unguarded** (authorization.go: not-found → pass-through).

Since `ServeMux` registrations can't be enumerated to cross-check, please either soften this to say route/spec correspondence is the plugin author's responsibility (not enforced), or add an explicit enforcement mechanism. Same wording appears in `internal/plugin/plugin.go`.

platform-api/internal/server/plugins.go
```diff
@@ -0,0 +1,186 @@
...
+					continue
+				}
+				switch m.Position {
+				case pdk.BeforePlatformChain:
+					w.preChain = append(w.preChain, m.Wrap)
+				case pdk.AfterPlatformChain:
+					w.postChain = append(w.postChain, m.Wrap)
+				default:
+					return fail("plugin %q declared middleware at unknown chain position %d", p.Name(), m.Position)
+				}
```
AnuGayan, line 147: 🟡 **Minor — this middleware-wiring branch is untested.** No test in `plugins_test.go` / `external_plugin_test.go` exercises `BeforePlatformChain` / `AfterPlatformChain` placement, the `Wrap == nil` skip, or this `default:` abort on an unknown position. The PR description lists "middleware position wiring" as covered — add a table test asserting a `MiddlewareProvider` plugin populates `preChain`/`postChain` correctly and that an out-of-range position aborts startup.

platform-api/internal/server/plugins.go
```diff
@@ -0,0 +1,186 @@
...
+		spec := p.OpenAPISpec()
+		if len(spec) == 0 {
+			return fail("plugin %q returned an empty OpenAPI spec; a spec declaring each route's scopes is required", p.Name())
+		}
+		pluginRegistry, regErr := middleware.LoadScopeRegistryFromBytes(spec)
+		if regErr != nil {
+			return fail("plugin %q OpenAPI spec failed to load into the scope registry: %w", p.Name(), regErr)
+		}
+		if pluginRegistry.Len() == 0 {
+			return fail("plugin %q OpenAPI spec declares no scope requirements; every route a plugin registers must declare its required scopes", p.Name())
```
AnuGayan: 🟡 **Minor — error text asserts more than the check verifies.** This check only confirms the *spec* has zero scoped operations, not per-route coverage (same root as the `pdk.OpenAPISpec` docstring). Consider rewording to e.g. `the plugin's OpenAPI spec declares no scoped operations` to avoid implying per-route enforcement.

platform-api/internal/server/server.go
```diff
@@ -464,9 +499,18 @@ func StartPlatformAPIServer(cfg *config.Server, slogger *slog.Logger) (*Server,
...
 	// Build the middleware chain that wraps the mux.
-	// Order: CORS → auth → scope enforcer → mux
+	// Order: [plugin preChain] → CORS → auth → org resolver → scope enforcer →
+	//        [plugin postChain] → mux
 	var chain []func(http.Handler) http.Handler
 
+	// Plugin "before" middleware — outermost, before CORS/auth. No authenticated
+	// identity is in the context here. pdk.Middleware shares
+	// its underlying type with the chain element type, so each is assignable.
+	for _, mw := range preChain {
```
AnuGayan, line 511: 🔵 Nit — `for _, mw := range preChain { chain = append(chain, mw) }` can be `chain = append(chain, preChain...)` (same for `postChain` below). `pdk.Middleware`'s underlying type is identical to the chain element type, and the PR already relies on that assignability. Purely cosmetic.

platform-api/internal/plugin/plugin.go
```diff
@@ -96,6 +101,24 @@ type Deps struct {
...
+// public (unauthenticated) path prefixes. The server validates each returned path
+// and appends it to cfg.Auth.SkipPaths before the auth middleware is built. Keep
+// prefixes narrow and specific — this is an auth-bypass surface (GO-AUTH-004).
+// Matching is a prefix match, so an over-broad prefix aborts startup: a path must
+// be non-empty, start with "/", not be "/" alone, and contain no "..".
+type AuthSkipPathProvider interface {
+	AuthSkipPaths() []string
+}
+
+// MiddlewareProvider is an optional interface a Plugin may implement to contribute
```
AnuGayan, line 114: 🔵 Nit — `MiddlewareProvider` is duplicated here and in `pdk`. It's justified (lets the server's single `p.(plugin.MiddlewareProvider)` assertion cover both tiers via the `externalPlugin` forwarder), but a one-line comment pointing at that forwarder in `external_plugin.go` would save the next reader the round-trip.

AnuGayan (changes requested):
## Automated review — external plugin support

Clean two-tier plugin architecture with well-considered fail-closed startup and capability-scoped `Deps`. The security *intent* is sound; my main concerns are two spots where the enforced behavior is weaker than the code documents, plus one ordering bug that breaks the feature in IDP role-validation mode. Details inline.

**Verdict: 🔧 Request changes** — Findings 1–3 (inline) before merge; 4–7 are minor/nits.

**What's good**
- Empty / malformed / scope-less specs correctly abort startup, with tests for all three.
- `validateAuthSkipPath` rejects `""`, `/`, no-leading-slash and `..`; external plugins are structurally denied the skip-path hook (nicely tested).
- `externalPlugin` hands over `pdk.Deps`, never `plugin.Deps` — asserted at both unit and startup-path level.
- Reverse-order shutdown of initialized plugins on failure + the `startupOK` guard against double-shutdown.
- Dropping the global `Register`/`All` registry in favor of explicit passing is the right call for testability.

platform-api/internal/server/plugins.go
malinthaprasan: There are requirements to override existing routes as well. So we might need to change the scopes of the routes as well and that is a valid requirement.

platform-api/internal/server/server.go
malinthaprasan: This is incorrect. To do this, the types has to be exactly the same. Not only the 'underlying literal type'. There exact types of preChain and chain are not the same here.

platform-api/internal/plugin/plugin.go
malinthaprasan: Fixed [Point the internal MiddlewareProvider doc at the external plugin forw…](https://github.com/wso2/api-platform/pull/2814/commits/44a30d8690c78bf408f4ec60236c55ffa4847ebb)

platform-api/internal/server/plugins.go
malinthaprasan: No longer required with https://github.com/wso2/api-platform/pull/2814#discussion_r3646002951

platform-api/internal/server/plugins.go
malinthaprasan: Fixed [Cover plugin middleware chain position wiring with tests](https://github.com/wso2/api-platform/pull/2814/commits/5da51f7f999adda4c6e5034a96e1d991aa2ccfa7)

platform-api/pdk/plugin.go
malinthaprasan: This check is removed. No need to enforce scoped routes only. 
[Stop requiring plugin specs to declare at least one scope](https://github.com/wso2/api-platform/pull/2814/commits/3c0d4b52cf03d19a8d98133049c7ca10a8e335bb)

platform-api/internal/server/server.go
malinthaprasan: Fixed [Validate roles.yaml after plugin scopes are merged into the registry](https://github.com/wso2/api-platform/pull/2814/commits/f15d567a133bbe975123b96d5c5f7f50a5796b33)

AnuGayan (approved):

PR #2879, opened 2026-07-27
Fix issue of emitting two analytics event per LLM proxy call

### LINKED ISSUE #2441: [Improvement]: Duplicate Analytics Events when Invoking an LLM Proxy

Area/AIGateway (AI Gateway runtime/control plane)

Aspect/Monitoring (Metrics, observability, health checks, alerts)

### Suggested Improvement

When invoking an LLM proxy, two different analytics events are published from the LLM proxy and the LLM provider. 

In the current scenario;
- LLM provider is secured with API key
- LLM proxy is secured with a JWT policy
When invoking the LLM proxy two analytics events can be seen in the analytics dashboards
- LLM provider related event is recorded without any user details
- LLM proxy related event is recorded with userID extracted from the token claims

Need to discuss whether we need to fix this from the gateway side or whether this should be managed from the analytics dashboard.

### Related Issues

https://github.com/wso2/api-platform/issues/2394

Tharanidk:
## Problem

Invoking an LLM proxy produced **two** analytics events for a single client call, because the proxy forwards to its provider over an internal loopback on the same Envoy listener. This caused:

- **Double-counted requests** in Moesif — e.g. "Total Requests" showed 12 for ~6 calls,
  while the per-API "AI API Details" panel showed the correct count.
- **Split data** — the proxy (ingress) event carried the user identity but no token usage;
  the provider (loopback) event carried token/cost/model but showed the user as `anonymous`.

## Fix 

1. Token usage/cost/model is already extracted on the proxy route (the LLM response returns through it) and present in `event.Properties`, but the publisher only emitted it for `LlmProvider`. The guard now includes `LlmProxy`, so the proxy event is a complete AI record.

2. Before publishing, an `LlmProvider` event whose downstream remote address is loopback(`127.0.0.1`/`::1`) is dropped, since it is the internal hop of a proxy call, not a real client request. Directly invoked providers (real client address) are unaffected and still publish.

## Moesif Dashboard

When I send 3 requests from the proxy and 4 requests from the provider, the analytics dashboard is showing as below.
image
image
image

Resolved: https://github.com/wso2/api-platform/issues/2441

### DIFF
--- a/gateway/gateway-runtime/policy-engine/internal/analytics/analytics.go
+++ b/gateway/gateway-runtime/policy-engine/internal/analytics/analytics.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"log/slog"
 	"maps"
+	"net"
 	"strconv"
 	"time"
 
@@ -144,10 +145,47 @@ func (c *Analytics) Process(event *v3.HTTPAccessLogEntry) {
 	}
 
 	analyticEvent := c.prepareAnalyticEvent(event)
+
+	// Suppress internal loopback provider events to avoid double-counting in analytics.
+	if c.isInternalLoopbackProviderEvent(analyticEvent) {
+		correlationID := ""
+		if analyticEvent.MetaInfo != nil {
+			correlationID = analyticEvent.MetaInfo.CorrelationID
+		}
+		slog.Debug("Suppressing internal loopback provider analytics event",
+			"apiType", analyticEvent.API.APIType,
+			"correlationId", correlationID,
+		)
+		return
+	}
+
 	for _, publisher := range c.publishers {
 		publisher.Publish(analyticEvent)
 	}
 
+}
+
+// isInternalLoopbackProviderEvent checks if the event is an internal loopback provider event.
+
+func (c *Analytics) isInternalLoopbackProviderEvent(event *dto.Event) bool {
+	if event == nil || event.API == nil {
+		return false
+	}
+	if event.API.APIType != "LlmProvider" {
+		return false
+	}
+	return isLoopbackAddress(event.UserIP)
+}
+
+// isLoopbackAddress reports whether ip is an IPv4/IPv6 loopback address.
+func isLoopbackAddress(ip string) bool {
+	if ip == "127.0.0.1" || ip == "::1" {
+		return true
+	}
+	if parsed := net.ParseIP(ip); parsed != nil {
+		return parsed.IsLoopback()
+	}
+	return false
 }
 
 // isInvalid checks if the log entry is invalid.
--- a/gateway/gateway-runtime/policy-engine/internal/analytics/analytics.go
+++ b/gateway/gateway-runtime/policy-engine/internal/analytics/analytics.go
@@ -352,6 +352,8 @@ func (c *Analytics) prepareAnalyticEvent(logEntry *v3.HTTPAccessLogEntry) *dto.E
 	if userIP == "" {
 		userIP = Unknown
 	}
+
+	directRemoteIP := logEntry.GetCommonProperties().GetDownstreamDirectRemoteAddress().GetSocketAddress().GetAddress()
 	if userAgent == "" {
 		userAgent = Unknown
 	}
@@ -378,10 +380,11 @@ func (c *Analytics) prepareAnalyticEvent(logEntry *v3.HTTPAccessLogEntry) *dto.E
 	// can drop it (one counted event per client call) while the traffic log keeps it. The
 	// primary signal is the x-wso2-internal-loopback marker the proxy stamps on its loopback
 	// forward — a direct provider call never carries it, so it is never suppressed regardless
-	// of network topology. The loopback downstream address is a secondary spoof guard.
+	// of network topology. The loopback guard uses directRemoteIP (the physical TCP peer),
+	// NOT userIP, so a forged "X-Forwarded-For: 127.0.0.1" cannot fake an internal source.
 	if extendedAPI.APIType == "LlmProvider" &&
 		keyValuePairsFromMetadata[InternalLoopbackMetadataKey] == "true" &&
-		isLoopbackAddress(userIP) {
+		isLoopbackAddress(directRemoteIP) {
 		event.Properties[PropInternalLoopbackProvider] = true
 	}
 
--- a/gateway/gateway-runtime/policy-engine/internal/analytics/analytics.go
+++ b/gateway/gateway-runtime/policy-engine/internal/analytics/analytics.go
@@ -81,6 +81,16 @@ const (
 
 	// UserIDMetadataKey represents the user ID metadata key for analytics.
 	UserIDMetadataKey string = "x-wso2-user-id"
+
+	// InternalLoopbackMetadataKey is the analytics-metadata key carrying the marker that
+	// the proxy stamps on its internal loopback forward to the provider (surfaced by the
+	// analytics system policy from the x-wso2-internal-loopback request header).
+	InternalLoopbackMetadataKey string = "x-wso2-internal-loopback"
+
+	// PropInternalLoopbackProvider is the event property set when a provider event is the
+	// internal loopback hop of a proxy call. Read by the Moesif publisher to skip it (avoids
+	// double-counting), while other publishers (traffic log) still emit it.
+	PropInternalLoopbackProvider string = "isInternalLoopbackProvider"
 )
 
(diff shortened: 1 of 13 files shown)


gateway/gateway-runtime/policy-engine/internal/analytics/analytics.go
```diff
@@ -144,12 +145,49 @@ func (c *Analytics) Process(event *v3.HTTPAccessLogEntry) {
...
+		return false
+	}
+	if event.API.APIType != "LlmProvider" {
+		return false
+	}
+	return isLoopbackAddress(event.UserIP)
+}
+
+// isLoopbackAddress reports whether ip is an IPv4/IPv6 loopback address.
+func isLoopbackAddress(ip string) bool {
```
O-sura, line 186: This way we assume that only internally generated requests will arrive over loopback which isn't always true i guess.

Ex:  in deployments using a local proxy/sidecar (e.g. Istio), kubectl port-forward, or any reverse proxy forwarding over 127.0.0.1, a legitimate client request can also have a loopback TCP peer. 

In those cases also the request is the client's first and only hop through the gateway, userIP will still be 127.0.0.1 and the analytics event will be suppressed even though there is no duplicate to remove.

The challenge is that isLoopbackAddress() tells us how the last TCP connection reached the process, not why the request exists. A gateway self-call and a client request forwarded via a local proxy are indistinguishable from this signal alone so better to verify this first

gateway/gateway-runtime/policy-engine/internal/analytics/analytics.go
```diff
@@ -144,12 +145,49 @@ func (c *Analytics) Process(event *v3.HTTPAccessLogEntry) {
...
+	if c.isInternalLoopbackProviderEvent(analyticEvent) {
+		correlationID := ""
+		if analyticEvent.MetaInfo != nil {
+			correlationID = analyticEvent.MetaInfo.CorrelationID
+		}
+		slog.Debug("Suppressing internal loopback provider analytics event",
+			"apiType", analyticEvent.API.APIType,
+			"correlationId", correlationID,
+		)
+		return
```
O-sura, line 176: The early `return` drops the event for the traffic-log publisher too.  The provider hop vanishes from the traffic log, and that's the only record measuring the real vendor round-trip. The surviving `LlmProxy` event's `BackendLatency` covers the whole chain including the loopback re-entry, so "was the LLM slow, or were we?" becomes unanswerable for proxied calls.

Verify whether this behavior is acceptable or not

O-sura (changes requested):

gateway/gateway-runtime/policy-engine/internal/analytics/analytics.go
Tharanidk: Will resolve this in a seperate issue

gateway/gateway-runtime/policy-engine/internal/analytics/analytics.go
Tharanidk: fixed

gateway/system-policies/analytics/analytics.go
```diff
@@ -172,6 +173,11 @@ func (a *AnalyticsPolicy) OnRequestHeaders(_ context.Context, reqCtx *policy.Req
 		if appNames := reqCtx.Headers.Get("x-wso2-application-name"); len(appNames) > 0 {
 			analyticsMetadata[ApplicationNameMetadataKey] = appNames[0]
 		}
+		// Marker stamped by the proxy on its internal loopback forward to the provider.
+		// Normalized to "true" so the policy-engine can drop the duplicate provider hop.
+		if loopback := reqCtx.Headers.Get(InternalLoopbackMetadataKey); len(loopback) > 0 && loopback[0] != "" {
+			analyticsMetadata[InternalLoopbackMetadataKey] = "true"
+		}
```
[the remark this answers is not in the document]
Tharanidk: not valud

O-sura:
The latest changes fixed the issue highlighted here in https://github.com/wso2/api-platform/pull/2879#discussion_r3662519836 but it introduces another issue. 

The marker header isn't stripped at ingress, and `userIP` reads `DownstreamRemoteAddress`, which Envoy derives from `X-Forwarded-For` when present. So any caller can send two headers and have their own analytics event discarded while still getting a 200.
 
 ```bash
# baseline → 1 analytics event
curl -k -sS -o /dev/null -w 'HTTP=%{http_code}\n' -X POST \
  https://localhost:8443/openai/latest/chat/completions \
  -H 'Content-Type: application/json' \
  -d '{"model":"gpt-4o-mini","messages":[{"role":"user","content":"Hi"}]}'

# same call + two headers → HTTP 200, ZERO analytics events
curl -k -sS -o /dev/null -w 'HTTP=%{http_code}\n' -X POST \
  https://localhost:8443/openai/latest/chat/completions \
  -H 'Content-Type: application/json' \
  -H 'x-wso2-internal-loopback: 1' \
  -H 'X-Forwarded-For: 127.0.0.1' \
  -d '{"model":"gpt-4o-mini","messages":[{"role":"user","content":"Hi"}]}'
```

Possible fix would be to use DownstreamDirectRemoteAddress field from the access log instead of the DownstreamRemoteAddress when reading the IP

File: gateway/gateway-runtime/policy-engine/internal/analytics/analytics.go [hunk 4]
```diff
@@ -343,6 +376,18 @@ func (c *Analytics) prepareAnalyticEvent(logEntry *v3.HTTPAccessLogEntry) *dto.E
...
+	// primary signal is the x-wso2-internal-loopback marker the proxy stamps on its loopback
+	// forward — a direct provider call never carries it, so it is never suppressed regardless
+	// of network topology. The loopback guard uses directRemoteIP (the physical TCP peer),
+	// NOT userIP, so a forged "X-Forwarded-For: 127.0.0.1" cannot fake an internal source.
+	if extendedAPI.APIType == "LlmProvider" &&
+		keyValuePairsFromMetadata[InternalLoopbackMetadataKey] == "true" &&
+		isLoopbackAddress(directRemoteIP) {
+		event.Properties[PropInternalLoopbackProvider] = true
+	}
+
```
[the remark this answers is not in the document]
Tharanidk: Normal external clients can't do this. so, this is not a valid comment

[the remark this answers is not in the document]
O-sura: If it is a local process or sidecar, then that's a deployment decision(The gateway owner knows such sidecar exists or if it's a process, we add such process intentionally) so that shouldn't be an issue. For port-forwarding also since we are depending on both IP and the header both, a legitimate user from the outside can't bypass both of those check so that also shouldn't be a problem isn't it?

gateway/gateway-runtime/policy-engine/internal/analytics/analytics.go
```diff
@@ -321,6 +352,8 @@ func (c *Analytics) prepareAnalyticEvent(logEntry *v3.HTTPAccessLogEntry) *dto.E
 	if userIP == "" {
 		userIP = Unknown
 	}
+
+	directRemoteIP := logEntry.GetCommonProperties().GetDownstreamDirectRemoteAddress().GetSocketAddress().GetAddress()
```
O-sura, line 368: Need to add a fallback or warn log when directRemoteIP is empty(Let's address via a seperate PR)

gateway/gateway-runtime/policy-engine/internal/analytics/analytics.go
```diff
@@ -144,12 +155,32 @@ func (c *Analytics) Process(event *v3.HTTPAccessLogEntry) {
 	}
 
 	analyticEvent := c.prepareAnalyticEvent(event)
+
+	// Suppress the internal loopback provider hop of an LLM proxy call so a single client
+	// call is counted once. Detection uses the x-wso2-internal-loopback marker the proxy
+	// stamps on its loopback forward (see prepareAnalyticEvent) — a direct provider call
+	// never carries it, so it is never suppressed.
+	if v, ok := analyticEvent.Properties[PropInternalLoopbackProvider].(bool); ok && v {
+		return
```
O-sura, line 176: Better to bring back the debug log mentioning the event suppression(Let's address via a seperate PR)

O-sura (dismissed):

O-sura (approved):

Krishanx92 (approved):

PR #2911, opened 2026-07-27
Add `ap:api_key:all:manage` scope for cross-user API key administration

Thushani-Jayasekera:
- https://github.com/wso2/api-platform/issues/2965

## Summary

API key CRUD is creator-scoped: only the user recorded in `api_keys.created_by` may update or
revoke a key. There is currently no way for an administrator to manage a key created by someone
else — a problem when a user leaves the organization, or when an operator needs to rotate a
compromised key on their behalf.

This PR introduces **`ap:api_key:all:manage`**, an organization-wide scope that lets its holder
perform API key CRUD on any user's keys, alongside the existing creator-based access. Every other
caller keeps the creator-scoped default.

It also removes the gateway-controller's ownership re-check on the platform-api event path, which
cannot work for admin-initiated changes: an `apikey.updated`/`apikey.revoked` event carries only
the acting user's id, never their scopes, so the gateway has no way to tell an org key admin apart
from an unauthorized third party and rejects every admin-initiated rotation.

## Scope model

`ap:api_key:all:manage` grants two things:

1. **Route access** — it is accepted on every API-key operation's `security` list, so a dedicated
   key-admin persona can call them without also holding `ap:rest_api:manage`,
   `ap:llm_provider:manage`, etc.
2. **Ownership override** — it lifts the `created_by` restriction inside the service layer.

Both are required; neither alone is sufficient. Reaching the endpoint (GO-AUTH-007) and being
allowed to touch another user's row are separate decisions, and the service-layer check is the
authoritative one (GO-AUTH-015).

The override widens **which rows within the caller's own organization** are reachable — never
which organization. Every query still sources `organization_uuid` from token claims (GO-AUTH-005).

Granted to `platform-admin` only. `platform-operator`, `platform-developer`, and `platform-viewer`
deliberately do not receive it.

## Surfaces covered

| Surface | Behavior with the scope |
|---|---|
| `PUT`/`DELETE /rest-apis/{id}/api-keys/{keyId}` | Update / revoke any user's key |
| `PUT`/`DELETE /websub-apis`, `/webbroker-apis` `.../api-keys/{keyId}` | Same (shared `APIKeyService`) |
| `GET`/`DELETE /llm-providers/{id}/api-keys` | List and delete any user's key |
| `GET`/`DELETE /llm-proxies/{id}/api-keys` | List and delete any user's key |
| `POST /applications/{id}/api-keys` | Bind any user's key to an application |
| `GET /me/api-keys` | Returns every user's keys in the org |

## Implementation

**One shared predicate.** The ownership rule was previously written out separately in each service,
with three subtly different forms. It is now a single function in `internal/service/apikey.go`
that every path calls, so it cannot drift between artifact kinds:

```go
func canManageAPIKey(createdBy, callerUserID string, keyAdmin bool) bool {
    if keyAdmin {
        return true
    }
    return callerUserID != "" && createdBy == callerUserID
}
```

**Explicit parameters, not implicit context.** Handlers resolve the scope once from the request's
verified claims via `middleware.HasEffectiveScope` (the pattern already used in
`organization.go`), then pass an explicit `keyAdmin bool` down. An authorization input travels as
a visible argument that a future refactor cannot silently drop.

**The repository race guard is unaffected.** `Update`/`Revoke` still re-verify `created_by` in
their `WHERE` clause, and admin paths pass the key's stored creator, so the delete/recreate
protection holds for admin-initiated writes too.

**Gateway side.** `APIKeyUpdateParams` and `APIKeyRevocationParams` gained an
`AuthorizedUpstream bool`, set to `true` only by `UpdateExternalAPIKeyFromEvent` /
`RevokeExternalAPIKeyFromEvent`. The check is skipped there, not deleted — the gateway's own REST
API path still enforces it in full, since the flag defaults to `false`.

## Behavior changes worth reviewer attention

**1. A latent break in the webhook path is fixed.**
`internal/webhook/handlers_apikey.go` calls update/revoke with `userID: ""`. The fail-closed
`userId == ""` check added in 65985191a made both of those webhook events return `403` on every
delivery. They now pass `keyAdmin: true`, justified by the envelope's HMAC signature (verified in
`ReceiveEvent` before dispatch) rather than by key ownership — the payload carries no user
identity at all. **Please confirm this reading matches intent.**

**2. The LLM services' implicit-admin bypass is removed.**
They previously treated an empty caller id as admin (`if userID != "" && key.CreatedBy != userID`),
which fails *open* for any path that forgets to populate the actor. The shared predicate fails
closed instead. All callers go through `resolveActorErr`, which errors on an unresolvable actor,
so no live path is affected.

**3. `/me/api-keys` widening is an explicit flag, not an empty string.**
`ListAPIKeysByUser` takes `allUsers bool` and **errors** if given an empty username without it,
rather than silently dropping the creator filter and returning the whole org's keys. The SQL is
unchanged in shape: the creator predicate is one of two code-controlled literals and the username
is always bound through a `?` placeholder (GO-AUTH-008).

## Known limitation

On the webhook path, `keyAdmin: true` bypasses ownership for **whatever key occupies the target
`(artifact, handle)`** — not only keys that channel provisioned. Since `api_keys` uniqueness is
artifact + name, a Developer Portal event could rotate or revoke a key an interactive user created
via the REST API under a colliding handle. This is documented in the code rather than left
implicit. If it is unacceptable, the tighter fix is to assert `created_by == ""` (system-owned) on
the webhook path instead of passing a blanket flag — happy to switch.

## Documentation

Added **GO-AUTH-019: Ownership-Override Scopes Must Be Named, Service-Layer Enforced, and Never
Silently Assumed** to `.claude/rules/authentication_authorization.md`. This pattern deliberately
cuts against the creator-scoped default and looks like a tenant-isolation hole at a glance, so it
needs a written rationale. The rule covers why an empty caller id must never mean "admin", why the
downstream check is gated rather than deleted, and why the override must not widen past the
caller's own organization. Purely additive — no existing rule renumbered or edited.

## Testing

- **New:** `internal/service/apikey_authz_test.go` — `TestCanManageAPIKey`, 7 cases pinning the
  ownership rule including the fail-closed empty-caller behavior. Passes.
- `go vet` clean across both `platform-api` and `gateway-controller`, test files included.
- Existing `internal/service/application_test.go` call sites updated for the new parameter.

⚠️ **Not run:** the full test suite. The dev machine ran out of disk space, so
`internal/repository` and `plugins/eventgateway/handler` could not be linked. Those two packages
are unverified — CI should be treated as the source of truth for them.

## Configuration / upgrade notes

The scope is appended to every shipped admin scope string (`config.toml`,
`config-template.toml`, `default_config.go`, Helm `values.yaml`, and both e2e configs). Operators
using IDP **role** mode get it automatically via `platform-admin` in `roles.yaml`. Operators using
**scope** mode with a hand-maintained scope list must add `ap:api_key:all:manage` to their admin
client for cross-user key management to work; without it, behavior is unchanged (creator-scoped).

`portals/ai-workspace/configs/config-template.toml` was deliberately **not** updated — it requests
neither `ap:api_key:read` nor this scope today, and defaulting a user-facing portal to an
org-wide admin grant seemed wrong. Easy to add if you disagree.


Thushani-Jayasekera:
@coderabbit review

.claude/rules/authentication_authorization.md
```diff
@@ -4,393 +4,1568 @@ Standards and patterns for ensuring secure authentication, authorization, token
...
+    }
+    if key == nil {
+        return apperror.RESTAPIKeyNotFound.New()
+    }
+    if !canManageAPIKey(key.CreatedBy, userID, keyAdmin) {
+        return apperror.RESTAPIKeyForbidden.New() // Generic 403 — never name the missing scope
+    }
+    // The row-level created_by re-check in the repository's WHERE clause still applies, so an
+    // admin-initiated write keeps the delete/recreate race guard (GO-AUTH-009's atomicity principle).
+    return s.apiKeyRepo.Revoke(artifactID, keyName, key.CreatedBy, userID)
```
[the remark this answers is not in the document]
Thushani-Jayasekera: The reported defect (RevokeAPIKey passing an undefined artifactID to GetByArtifactAndName and Revoke instead of the handle parameter) does not exist in the current code:

- platform-api/internal/service/apikey.go:695 RevokeAPIKey resolves apiHandle → apiMetadata.ID via GetAPIMetadataByHandleAndKind, then passes that same apiId to both s.apiKeyRepo.GetByArtifactAndName(apiId, keyName) (:713) and s.apiKeyRepo.Revoke(apiId, keyName, ...) (:736). No artifactID identifier appears anywhere in the file.
- The other RevokeAPIKey (gateway/gateway-controller/pkg/utils/api_key.go:419) likewise uses params.Handle → config.UUID consistently.
- Both packages build clean (go build ./...), which an undefined variable would have broken outright — further confirming the finding described code that no longer exists (or never did).

platform-api/config/config.toml
```diff
@@ -36,4 +36,4 @@ region       = "us"
...
...cp_delete dp:mcp_content_read dp:mcp_content_create dp:mcp_content_update dp:mcp_content_manage dp:mcp_content_delete dp:api_key_create dp:api_key_read dp:api_key_update dp:api_key_manage dp:api_key_revoke dp:mcp_key_create dp:mcp_key_read dp:mcp_key_update dp:mcp_key_manage dp:mcp_key_revoke dp:api_workflow_create dp:api_workflow_read dp:api_workflow_update dp:api_workflow_delete dp:api_workflow_manage dp:app_create dp:app_read dp:app_update dp:app_manage dp:app_delete dp:app_key_create dp:app_key_manage dp:app_key_revoke dp:app_key_mapping_read dp:app_key_mapping_write dp:app_key_mapping_manage dp:subscription_create dp:subscription_read dp:subscription_update dp:subscription_manage dp:subscription_delete dp:sub_plan_create dp:sub_plan_read dp:sub_plan_update dp:sub_plan_manage dp:sub_plan_delete dp:km_create dp:km_read dp:km_update dp:km_manage dp:km_delete dp:view_create dp:view_read dp:view_update dp:view_manage dp:view_delete dp:label_create dp:label_read dp:label_update dp:label_manage dp:label_delete dp:webhook_subscriber_create dp:webhook_subscriber_read dp:webhook_subscriber_update dp:webhook_subscriber_delete dp:webhook_subscriber_manage dp:event_read dp:delivery_manage"
```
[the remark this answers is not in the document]
Thushani-Jayasekera: Verified — no fix needed. Skipped: both parts are already remediated in the current code.

- platform-api/config/config.toml:37-39 — the bootstrap user is username = '{{ env "APIP_CP_ADMIN_USERNAME" }}' / password_hash = '{{ env "APIP_CP_ADMIN_PASSWORD_HASH" }}'. Neither token carries a default fallback value (compare logging.level, which does: {{ env "APIP_CP_LOGGING_LEVEL" "info" }}), so an unset env var resolves empty, not to a shipped credential.
- platform-api/config/default_config.go:96 — Users: FileBasedUsers{}, with a comment stating exactly the reasoning in the finding (no shipped hash, since that admin now also holds ap:api_key:all:manage).
- Fail-closed startup is in place: validateFileBasedConfig (platform-api/config/config.go:818) rejects len(cfg.Users) == 0, and per-user empty username/password_hash, when auth.mode=file.

So the default admin credential is already gone and startup already refuses to boot without an operator-supplied one.

.claude/rules/authentication_authorization.md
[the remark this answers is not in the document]
Thushani-Jayasekera: - It was introduced in e5c22fc23 ("Implement ownership override for API key management"), where .claude/rules/authentication_authorization.md was 1570 lines. That version's GO-AUTH-019 example did exactly what the reviewer describes:
func (s *APIKeyService) RevokeAPIKey(ctx context.Context, handle, kind, orgID, keyName, userID string, keyAdmin bool) error {
    key, err := s.apiKeyRepo.GetByArtifactAndName(artifactID, keyName)   // undeclared
    ...
    return s.apiKeyRepo.Revoke(artifactID, keyName, key.CreatedBy, userID) // undeclared
- handle is the declared parameter; artifactID is never bound. Two occurrences, as reported.
- The merge commit 68782a557 (upstream/main → fix-api-keu) resolved the file to the condensed 436-line version. That version's GO-AUTH-019 (.claude/rules/authentication_authorization.md:378-413) drops the full handler/service walkthrough entirely and keeps only the canManageAPIKey predicate, the ListAPIKeysByUser repo guard, and the AuthorizedUpstream gate. grep artifactID over the whole repo returns nothing.
- The surviving snippets are internally consistent — params.Handle, params.User, params.AuthorizedUpstream all line up, and there's no function signature left to contradict.

So there's nothing to fix. Your reply is accurate on the substance (documentation-only, no production defect), but it concedes a correction that's already been made by the upstream merge. You may want to say instead that the example the comment refers to was replaced during the merge with upstream/main and the current revision no longer contains it — inviting the reviewer to re-check against the branch tip rather than the earlier commit.

platform-api/internal/apperror/codes.go
```diff
@@ -96,6 +96,8 @@ const (
 	CodeRESTAPINotFound                   = "REST_API_NOT_FOUND"
 	CodeRESTAPIExists                     = "REST_API_EXISTS"
 	CodeRESTAPIDeploymentValidationFailed = "REST_API_DEPLOYMENT_VALIDATION_FAILED"
+	CodeRESTAPIKeyNotFound                = "REST_API_KEY_NOT_FOUND"
+	CodeRESTAPIKeyForbidden               = "REST_API_KEY_FORBIDDEN"
```
malinthaprasan:
```suggestion
	CodeRESTAPIKeyNotFound                = "REST_API_API_KEY_NOT_FOUND"
	CodeRESTAPIKeyForbidden               = "REST_API_API_KEY_FORBIDDEN"
```

gateway/gateway-controller/pkg/utils/api_key.go
```diff
@@ -82,6 +82,11 @@ type APIKeyRevocationParams struct {
 	User          *commonmodels.AuthContext // User who initiated the request
 	CorrelationID string                    // Correlation ID for tracking
 	Logger        *slog.Logger              // Logger instance
+	// AuthorizedUpstream marks this operation as already authorized by the platform API
+	// before it was broadcast, so the gateway must not re-run its own creator check.
+	// Set only on the platform-api event path (see RevokeExternalAPIKeyFromEvent); it is
+	// always false for requests arriving on the gateway's own REST API.
+	AuthorizedUpstream bool
```
malinthaprasan: What does "upstream" mean? Do you mean AuthorizedOrigin? (origin of the call) (or TrustedOrigin - bit shorter)

platform-api/internal/handler/apikey_user.go
```diff
@@ -66,7 +68,11 @@ func (h *APIKeyUserHandler) ListUserAPIKeys(w http.ResponseWriter, r *http.Reque
 
 	limit, offset := parsePagination(r)
 
-	response, err := h.apiKeyUserService.ListAPIKeysByUser(r.Context(), orgID, callerUserID, types, limit, offset)
+	// Holding constants.ScopeAPIKeyAllManage widens this listing from the caller's own keys
+	// to every user's keys in the organization.
+	keyAdmin := middleware.HasEffectiveScope(r, h.authzMode, constants.ScopeAPIKeyAllManage)
```
malinthaprasan, line 73: This name is bit confusing. "key" is used to indicate a "field". So this can be misread as a generic admin check and can be reused in unwanted places. Shall we be explicit here? Fix other places too. 
`hasApiKeyAdmin := ..`

platform-api/internal/service/apikey.go
```diff
@@ -556,29 +616,27 @@ func (s *APIKeyService) UpdateAPIKey(ctx context.Context, apiHandle, kind, orgId
...
 		MaskedAPIKey: maskedAPIKey,
 		APIKeyHashes: apiKeyHashesJSON,
 		Status:       constants.APIKeyStatusActive,
-		UpdatedBy:    userId,
-		ExpiresAt:    expiresAt,
-		Issuer:       req.Issuer,
+		// CreatedBy is required so the repository can re-verify the creator in its WHERE
+		// clause, guarding against a delete/recreate race between the check above and this
+		// write (see APIKeyRepo.Update).
+		CreatedBy: existingKey.CreatedBy,
```
malinthaprasan: Are we always passing the existing Key's CreatedBy? Does the created by check underneath the repository method ever fail in practice?
If not - its better to just drop this as this is confusing. You've already done the authz check earlier.

.claude/rules/authentication_authorization.md
```diff
@@ -375,6 +375,45 @@ os.chmod(socket_path, 0o660)  # defense-in-depth only, not the primary guarantee
 
 ---
 
+## GO-AUTH-019: Ownership-Override Scopes Must Be Named and Service-Layer Enforced — High
+
+Most resource CRUD here is *creator-scoped*: a caller may act only on rows they created (`api_keys.created_by`). Where cross-user administration is genuinely needed, express it as one explicitly named org-wide scope (`ap:api_key:all:manage`) resolved from verified claims at the handler, passed down as an explicit argument, and enforced by a single shared predicate at the service layer that every entry point funnels through (GO-AUTH-015). Never approximate it with an implicit signal — an empty caller id, or "this call came from an internal path so it must be trusted". An empty identity treated as elevated fails *open* for any path that merely forgot to populate the actor. The override widens which rows within the caller's own organization are reachable, never which organization (GO-AUTH-005).
```
malinthaprasan: "one explicitly named org-wide scope" - This should not mean that we should use one scope for everything. Scopes are always specific to resources. Nobody should ever create a scope called "ap:admin"

malinthaprasan: Most resource CRUD here is *creator-scoped* - 
What does "here" mean? Under this topic or platform? Shall we be specific?
In APIP, most CRUD resources are NOT *creator-scoped*

gateway/gateway-controller/pkg/utils/api_key.go
Thushani-Jayasekera: yes, a flag to identify if the request is from a trusted / pre validated origin. Will update the parameter name to `TrustedOrigin ` as  AuthorizedUpstream gives incorrect meaning.

platform-api/internal/service/apikey.go
Thushani-Jayasekera: Ah yeah, before reaching L629, we have this check
```
	if !trustedOrigin && !canManageAPIKey(existingKey.CreatedBy, userId, keyAdmin) {
		return apperror.RESTAPIAPIKeyForbidden.New()
	}
```

created by was passed here as an additional guard when selecting the API Key record to update. But since the validation is already done before reaching the db update state we can remove.

.claude/rules/authentication_authorization.md
Thushani-Jayasekera: ### Updated Rule

## GO-AUTH-019: Ownership-Override Scopes Must Be Named and Service-Layer Enforced — High

Most resources in this repo are *organization-scoped*: any caller in the org with the right resource scope may act on them, and GO-AUTH-005 alone governs access. A few are additionally *creator-scoped* — a caller may act only on rows they created — API keys (`api_keys.created_by`, enforced by `canManageAPIKey` in `platform-api/internal/service`) being the canonical case. This rule applies to those creator-scoped resources; note that a `created_by`/`CreatedBy` column is usually just an audit field and does not by itself make a resource creator-scoped. Where cross-user administration of a creator-scoped resource is genuinely needed, express it as an explicitly named override scope **for that one resource type** (`ap:api_key:all:manage` overrides creator-scoping on API keys and nothing else) — resolved from verified claims at the handler, passed down as an explicit argument, and enforced by a single shared predicate at the service layer that every entry point funnels through (GO-AUTH-015). Never approximate it with an implicit signal — an empty caller id, or "this call came from an internal path so it must be trusted". An empty identity treated as elevated fails *open* for any path that merely forgot to populate the actor. The override widens which rows within the caller's own organization are reachable, never which organization (GO-AUTH-005).

"One named scope" means one *per resource type*, not one for the whole platform. Never define a catch-all override like `ap:admin` or `ap:all:manage` that a handler for any resource can accept: a scope is always bound to the resource it governs, so granting API-key administration cannot silently confer subscription or application administration. A caller who legitimately administers several resource types holds several scopes. Wildcards are not an exception to this, `:*` is **own-level only** — it covers the actions directly at its level and never descends into sub-resources, matches a prefix, or acts transitively (`ap:gateway:*` satisfies `ap:gateway:create` but not `ap:gateway:token:create`). `scopeSatisfies` in `platform-api/internal/middleware/authorization.go` implements exactly that expansion. Wildcards live on the *grant* side only: a handler or service-layer predicate always names the concrete scope it requires and never matches on a wildcard itself.

A scope decision does not survive a broadcast: an event carries the acting user's id, not their scopes, so a downstream `created_by` re-check rejects every legitimate admin-initiated change. Gate that check behind an explicit "trusted origin" flag (the call came from a pre-validated origin) set only by the event path — deleting it outright would also disable it for that component's own directly-authenticated API.

renuka-fernando (approved):

PR #2934, opened 2026-07-28
[Gateway] Fixing Go template rendering for policy parameters

### LINKED ISSUE #2871: Gateway: template expressions unusable in non-string policy params
### Description

A policy parameter declared as `integer`, `number`, or `boolean` cannot be set to a Go-template expression (`{{ env "X" }}`, `{{ secret "X" }}`). The gateway-controller resolves templates *before* validation, but the resolved value is always a **string**, so it fails in two places: JSON-schema validation rejects it as a type mismatch, and — even if validation passed — the string flows unchanged to the policy engine, which type-asserts the value and breaks. Literal numeric/boolean values work; only templated ones fail. This defeats the purpose of resolving templates in the gateway for any non-string param.

Example: the **advanced-ratelimit** `limit`/`burst` params (integers) and `slugify-body`'s `maxLength` (integer) cannot use `{{ env … }}`.

### Root cause (traced)

1. `templateengine.RenderSpec` resolves templates before validation (by design), but `renderValue` returns every rendered leaf as a **string** — `gateway/gateway-controller/pkg/templateengine/spec.go:152-162`. So `limit: "{{ env \"LIMIT\" }}"` becomes the string `"100"`, never the number `100`.
2. Validation order: render then validate — `gateway/gateway-controller/pkg/utils/api_deployment.go:316` (render) → `:326` (validate).
3. `validatePolicyParams` passes params straight to `gojsonschema` with **no coercion** — `gateway/gateway-controller/pkg/config/policy_validator.go:278-286`. A JSON string `"100"` fails `type: integer`/`number`.
4. Even past validation, the transform copies params through untouched — `gateway/gateway-controller/pkg/transform/restapi.go:632-650` — into `RuntimeDeployConfig` (`pkg/models/runtime_deploy_config.go:114`) → xDS → the policy engine, which hands the bare `map[string]interface{}` to the policy factory (`gateway-runtime/policy-engine/internal/registry/registry.go:84-122`). Policies then type-assert values (e.g. `gateway/sample-policies/transform-payload-case/transformpayloadcase.go:47`), so a string where a number is expected breaks at runtime.

### Proposed solution

Add a **schema-aware, recursive coercion** of resolved policy-param values to their declared JSON-schema type, applied **after `RenderSpec` and before `Validate`**, so both validation and everything downstream (xDS, policy engine) receive correctly typed values:

- The policy definition schema is already reachable at this stage — `PolicyValidator` and `RestAPITransformer` both hold `policyDefinitions` (`pkg/models/policy_definition.go:27`), so no new plumbing is needed to look up a param's declared type.
- For each leaf whose schema type is `integer`/`number`/`boolean`, convert a clean scalar string to the typed value; leave everything else untouched.
- It must **walk the schema recursively** (`properties`/`items`), because the real target (`advanced-ratelimit.limit`) is nested under `quotas[].limits[]`. This mirrors the algorithm the AI Workspace frontend already has in `portals/ai-workspace/.../schemaUtils.ts:320` (`coerceValuesToSchemaTypes`).
- Coerce **then** validate: this preserves real validation — if a template resolves to a non-numeric string (e.g. env var `"abc"` for an integer), coercion can't produce a number and `gojsonschema` still rejects it.

**Alternative considered:** schema-unaware re-typing in `templateengine` (JSON-parse a pure-template scalar leaf back to number/bool). Rejected as riskier — it would over-coerce genuine string params that happen to resolve to numeric-looking strings.

### Version

_No response_

### Related Issue

#2872

Milanka00:
## Purpose

Fixes policy parameter validation and runtime handling when Go templates `({{ env "..." }}, {{ secret "..." }})` are used for integer, number, or boolean fields.

RenderSpec JSON-roundtrips the config, so template output is always a string (e.g. "10"). The JSON schema validator and policy engine expect typed values, causing deploy failures like `Expected: integer, given: string `and runtime 500s when string params reach the policy engine.

This PR adds schema-aware coercion that converts rendered strings to the types declared in each policy’s parameter schema (including nested paths such as `quotas[].limits[].limit in advanced-ratelimit`).

## Approach

- Add coerceParamsBySchema / coerceScalarByType — walk policy JSON schema recursively and coerce:
    - integer / number → float64 via strconv.ParseFloat
    - boolean → true / false from "true" / "false"
    - nested object and array (including arrays of objects)

- Call coercion inside validatePolicy before JSON schema validation so deploy-time validation sees typed values

- Expose public helpers for post-render paths:

  - CoerceRestAPIPolicies
  - CoerceMCPProxyPolicies
  - CoerceLLMPolicies

- Unit tests for top-level, nested object, array-of-objects, and array-of-scalar integer coercion
- Deploy-time coercion alone is not enough for runtime. On replica sync, the event listener reloads from DB, re-runs RenderSpec, and pushes config into the in-memory store → xDS → policy engine. That path had no coercion, there for necessary implementations are done to processors.

## Automation tests
- Unit tests are added

## Test environment
- Flows mentioned https://github.com/wso2/api-platform/issues/2871#issuecomment-5100057935 were tested

Fixes https://github.com/wso2/api-platform/issues/2871

### DIFF
--- a/gateway/it/features/api-error-responses.feature
+++ b/gateway/it/features/api-error-responses.feature
@@ -66,7 +66,7 @@ Feature: API Error Responses
           - name: respond
             version: v1
             params:
-              statusCode: "200"
+              statusCode: "not-a-number"
         operations:
           - method: GET
             path: /test
--- a/gateway/gateway-controller/pkg/config/policy_validator.go
+++ b/gateway/gateway-controller/pkg/config/policy_validator.go
@@ -21,6 +21,7 @@ package config
 import (
 	"fmt"
 	"regexp"
+	"math"
 	"strconv"
 	"strings"
 
@@ -418,7 +419,7 @@ func coerceScalarByType(val interface{}, expectedType string) interface{} {
 	}
 	switch expectedType {
 	case "integer", "number":
-		if f, err := strconv.ParseFloat(s, 64); err == nil {
+		if f, err := strconv.ParseFloat(s, 64); err == nil && !math.IsNaN(f) && !math.IsInf(f, 0) {
 			return f
 		}
 	case "boolean":
--- a/gateway/gateway-controller/pkg/config/policy_validator.go
+++ b/gateway/gateway-controller/pkg/config/policy_validator.go
@@ -21,6 +21,7 @@ package config
 import (
 	"fmt"
 	"regexp"
+	"strconv"
 	"strings"
 
 	versionutil "github.com/wso2/api-platform/common/version"
@@ -155,12 +156,14 @@ func (pv *PolicyValidator) validatePolicy(policy api.Policy, fieldPath string) [
 		return errors
 	}
 
-	// Validate policy parameters against JSON schema if schema is defined
+	// Coerce then validate policy parameters against the declared JSON schema.
 	if policyDef.Parameters != nil {
-		// If params is nil, validate against an empty object to enforce required fields
 		params := make(map[string]interface{})
 		if policy.Params != nil {
 			params = *policy.Params
+			// Coerce template-rendered strings to their schema-declared types using the
+			// already-resolved policyDef — avoids a second resolvePolicyVersion call.
+			coerceParamsBySchema(params, *policyDef.Parameters)
 		}
 		schemaErrs := pv.validatePolicyParams(params, *policyDef.Parameters, fieldPath+".params")
 		errors = append(errors, schemaErrs...)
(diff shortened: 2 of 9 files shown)


gateway/it/features/api-error-responses.feature
```diff
@@ -66,7 +66,7 @@ Feature: API Error Responses
           - name: respond
             version: v1
             params:
-              statusCode: "200"
+              statusCode: "not-a-number"
```
renuka-fernando, line 69: We can add an Int Test.

renuka-fernando: Created an issue to track this https://github.com/wso2/api-platform/issues/2967

renuka-fernando (approved):

Krishanx92 (approved):

Issue #2831, opened 2026-07-23
[Task]: Support extending Platform API via a Plugin Development Kit (PDK)


malinthaprasan:
### Please select the area the issue is related to

Platform API

### Please select the aspect the issue is related to

Aspect/API (API backends, definitions, contracts, interfaces, OpenAPI)

### Description

**Problem**

`platform-api` can only be extended by editing it. A separate module cannot add its own REST APIs, config, or middleware without forking or patching platform-api sources.

**Requirement**

Make `platform-api` an importable library with a public, stable extension contract (a Plugin Development Kit), so a wrapper module can run and extend the full platform server by importing it — without changing platform-api.

At a high level:

- A wrapper can run the platform server and register its own plugins.
- A plugin can add REST APIs, an OpenAPI spec, and request-chain middleware.
- Plugins reach the platform through a public capability contract, never internal packages.
- Extension routes run under the platform's existing auth and scope enforcement.
- Extension is compile-time (by import); no runtime plugin loading.
- The default build behaves exactly as before.

**Related:**
N/A

PR #2961, opened 2026-07-28
Add plugin route overrides, and drop the platform pdk re-exports

malinthaprasan:
## Purpose

A wrapper embedding platform-api as a library can add routes and chain middleware, but cannot change what an *existing* core endpoint returns. The only options are forking the handler or shadowing the route, and `ServeMux` panics on a duplicate pattern.

https://github.com/wso2/api-platform/issues/2831

## Goals

Let a plugin decorate one existing core route: its `Wrap` receives the original core handler as `next`. 

## Approach

Six commits, each buildable on its own:

1. **`internal/router`** — a `Router` interface (`*http.ServeMux` satisfies it) and a `Recorder` that records registrations instead of serving them. Core handlers' `RegisterRoutes` now take `router.Router`; the `pdk.Plugin` / `plugin.Plugin` contracts keep `*http.ServeMux`, so no plugin or wrapper changes.
2. **`pdk/override.go`** — `RouteOverride`, `RouteOverrideProvider`, and the capture helpers `Invoke` / `WriteCaptured`.
3. **Server wiring** — core routes are recorded, plugins declare their claims, then `installCoreRoutes` validates them and installs each recorded route on the mux, wrapped where claimed.
4. **Façade removal** — `platform` re-exports nothing from `pdk`. See *Breaking changes in the PDK* below.
5. **`RouteDecorator`** — a defined type of `func(http.Handler) http.Handler` to provide the handler funtion to override the default handler.

Constraints:
- An override cannot change the route's scopes-
- A plugin route colliding with a core pattern gives a startup error

## Breaking changes in the PDK

`platform` no longer re-exports anything from `pdk`. It exports only `New`, `App`, `Run`, and the `With*` options; every contract type has one name, in `pdk`.

Seven symbols are removed, all previously released on `main`:

| Removed | Replace with |
| --- | --- |
| `platform.Plugin` | `pdk.Plugin` |
| `platform.Deps` | `pdk.Deps` |
| `platform.Middleware` | `pdk.Middleware` |
| `platform.ChainPosition` | `pdk.ChainPosition` |
| `platform.PositionedMiddleware` | `pdk.PositionedMiddleware` |
| `platform.BeforePlatformChain` | `pdk.BeforePlatformChain` |
| `platform.AfterPlatformChain` | `pdk.AfterPlatformChain` |

## User stories

- As a wrapper author, I add a field to an existing endpoint's response without forking its handler or duplicating its route.
- As a reviewer, every decorated route is declared explicitly and logged at startup, and a stale pattern stops the server instead of quietly doing nothing.

## Documentation

N/A — extension surface only, no user-facing API change. The contract is documented in the doc comments on `pdk.RouteOverride`, `Invoke`, and `WriteCaptured`.

## Automation tests

- **Unit tests** — 32 new. `internal/router` 100.0% statement coverage, `pdk` 96.8%, `installCoreRoutes` 100.0%. Cover ordering and deferred errors in the recorder, capture/write semantics, and every startup-failure path (unknown pattern, duplicate claims, nil `Wrap`, empty pattern, plugin/core collision). Two assert the header-replacement fix on `Header().Values()` rather than `Get()` — `Get` returns only the first value, so it cannot see a duplicate. Two more pin `RouteDecorator` as a type distinct from `Middleware`, while confirming a plain func literal still assigns to `Wrap`.
- **Integration tests** — none new. Manually verified against a running server with a wrapper declaring two overrides: enriched response on the overridden route, list route unchanged, core's 404 passed through byte for byte, `405` preserved, and a bad pattern refusing startup.

## Security checks

- Followed secure coding standards? **yes**
- Ran FindSecurityBugs plugin? **N/A** — Go module, no Java source; `go vet` clean.
- Confirmed no keys/passwords/tokens/secrets committed? **yes**

An override cannot widen access. Required scopes are keyed by OpenAPI path/method and are untouched, so a decorated route keeps the requirement it had. A decorator cannot re-route a request either — the handler for the pattern is already selected before `Wrap` runs, so rewriting the path changes nothing about what executes. The contract documents that a decorator must read the organization from request context, never from request input. Every malformed or unmatched override aborts startup rather than being skipped.

## Samples

```go
import "github.com/wso2/api-platform/platform-api/pdk"

// Declare the claim. Pattern is matched as an exact string against what core
// registers — a wrong version or wildcard name fails startup.
func (p *MyPlugin) RouteOverrides() []pdk.RouteOverride {
    return []pdk.RouteOverride{
        {Pattern: "GET /api/v0.9/gateways/{gatewayId}", Wrap: p.enrichGateway()},
    }
}

// next is the original core handler, registered under the same pattern — so the
// mux has already resolved r.PathValue("gatewayId") by the time this runs.
func (p *MyPlugin) enrichGateway() pdk.RouteDecorator {
    return func(next http.Handler) http.Handler {
        return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
            res := pdk.Invoke(next, r) // run core, capture its response

            if res.Status != http.StatusOK {
                pdk.WriteCaptured(w, res) // pass errors through untouched
                return
            }

            var base platformapi.GatewayResponse
            if err := json.Unmarshal(res.Body, &base); err != nil {
                trackingID := util.NewToken()
                p.deps.Logger.Error("gateway decode failed",
                    "trackingId", trackingID, "error", err)
                httputil.WriteJSON(w, http.StatusInternalServerError, map[string]any{
                    "error": "internal_error", "tracking_id": trackingID,
                })
                return
            }

            httputil.WriteJSON(w, http.StatusOK,
                toCloudGateway(base, p.environmentFor(base.Id)))
        })
    }
}
```

## Related PRs

None.

## Test environment

Go 1.26.5, macOS. `go build ./...`, `go build ./cmd/main.go`, `go vet`, and the test suite all clean; each of the first three commits was checked out and built independently. The reference wrapper (`apip-cloud-platform-api`) was also smoke-built against this branch through its `.local` symlinks — `go build`, `go vet`, `go test` clean — since nothing in this repo compiles against `pdk` from outside the module.


platform-api/internal/server/overrides.go
```diff
@@ -0,0 +1,90 @@
...
+		if !core.Has(pattern) {
+			unknown = append(unknown, pattern)
+		}
+	}
+	if len(unknown) > 0 {
+		sort.Strings(unknown)
+		p := unknown[0]
+		return fmt.Errorf("plugin %q declared a route override for %q, which is not a core route "+
+			"(patterns are matched exactly, including method and path; %d override(s) unmatched)",
+			overrides[p].plugin, p, len(unknown))
```
renuka-fernando, line 62: Just a minor comment. We can include all the unknown list of patterns.

renuka-fernando (approved):

Issue #2914, opened 2026-07-27
[Task]:  Introduce downstream path, method and upstream status code in SDK


renuka-fernando:
### Description

The SDK's `DownstreamRequest` (introduced to snapshot the client request before policy mutation) currently only captures `Headers`. Similarly, `UpstreamResponse` only captures `Headers`. This is incomplete because policies can also mutate path, method, authority, scheme (via `UpstreamRequestHeaderModifications.Path/Method/Host`) and response status code (via `DownstreamResponseModifications.StatusCode`).

When a policy earlier in the chain rewrites the path or method, subsequent policies see the mutated value in `ctx.Path`/`ctx.Method` — they have no way to access what the original client sent. The same applies to the response status code relative to `Upstream.Response`.

**Changes needed in `sdk/core/policy/v1alpha2/context.go`:**

Add `Path`, `Method`, `Authority`, and `Scheme` to `DownstreamRequest`:

```go
type DownstreamRequest struct {
    Headers   *Headers
    Path      string
    Method    string
    Authority string
    Scheme    string
}
```

Add `StatusCode` to `UpstreamResponse`:

```go
type UpstreamResponse struct {
    Headers    *Headers
    StatusCode int
}
```

**Changes needed in the kernel (`execution_context.go`):**

- `buildRequestContexts`: populate the new `DownstreamRequest` fields from the pre-mutation scalar values (same point where `downstreamHeaders` is snapped).
- `buildResponseContexts`: pass `responseStatus` into `toResponseUpstream` and populate `UpstreamResponse.StatusCode`.

**Note:** `Downstream.Request.Path/Method` etc. are technically recoverable from the downstream headers snapshot via `Headers.Get(":path")` (pseudo-headers are included), but this is undiscoverable and not a documented API. Explicit fields complete the model symmetrically with how headers are already handled.

Body snapshots (`DownstreamRequest.Body`, `UpstreamResponse.Body`) are intentionally excluded — bodies are not mutated between policies in the current chain model, and the memory cost of snapshotting every body is a separate, larger decision.

PR #2974, opened 2026-07-29
feat(sdk): snapshot original path, method, authority, scheme and upstream status code

renuka-fernando:
## Purpose

Policies can mutate the request path, method, authority (via `UpstreamRequestHeaderModifications`/`UpstreamRequestModifications` `Path`/`Method`/`Host`) and the response status code (via `DownstreamResponseModifications.StatusCode`). A policy running later in the chain that needs the *original* client-sent value has no way to recover it — `ctx.Path`/`ctx.Method` already reflect the post-mutation state, and the kernel's downstream/upstream snapshots only captured headers, not these scalar fields.

This closes that gap in the SDK type surface so the snapshot model is symmetric: headers, path/method/authority/scheme (downstream), and status code (upstream) are all recoverable in their pre-mutation form.

This is the SDK-only part, released independently so policy authors can compile against the new fields. The policy-engine kernel wiring and the `python_executor.proto` change that populate them follow in a separate PR.

## Approach

- `DownstreamRequest` (Go + Python): add `Path`, `Method`, `Authority`, `Scheme` alongside the existing `Headers` snapshot.
- `UpstreamResponse` (Go + Python): add `StatusCode` alongside the existing `Headers` snapshot.
- Python uses `status_code: int = 0` (snake_case per convention; `0` default since the field is unset until the kernel populates it).
- No kernel/proto changes here — the new fields are zero-valued until the follow-up policy-engine PR wires them in `buildRequestContexts`/`buildResponseContexts` and the proto.

## Related Issues

Related #2914

The issue will be closed by the follow-up policy-engine PR, not this one.

## Checklist

## Remarks

Additive struct fields only — no behavior change and no breaking change for existing policies. Go builds clean; Python SDK tests pass (3/3).

### DIFF
--- a/sdk-python/src/apip_sdk_core/policy/v1alpha2/types.py
+++ b/sdk-python/src/apip_sdk_core/policy/v1alpha2/types.py
@@ -143,6 +143,10 @@ class DownstreamRequest:
     """
 
     headers: Headers | None = None
+    path: str = ""
+    method: str = ""
+    authority: str = ""
+    scheme: str = ""
 
 
 @dataclass(slots=True)
@@ -182,6 +186,7 @@ class UpstreamResponse:
     """
 
     headers: Headers | None = None
+    status_code: int = 0
 
 
 @dataclass(slots=True)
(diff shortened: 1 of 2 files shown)


Krishanx92 (approved):

O-sura (approved):

Krishanx92 (approved):

renuka-fernando:
@coderabbitai review

PR #2976, opened 2026-07-29
feat(policy-engine): populate downstream path/method and upstream sta…

### LINKED ISSUE #2914: [Task]:  Introduce downstream path, method and upstream status code in SDK

renuka-fernando:
## Purpose

Policies can mutate the request path, method, authority (via `UpstreamRequestHeaderModifications`/`UpstreamRequestModifications`) and the response status code (via `DownstreamResponseModifications.StatusCode`). A policy running later in the chain that needs the *original* client-sent value had no way to recover it — `ctx.Path`/`ctx.Method` already reflect the post-mutation state, and the kernel's downstream/upstream snapshots only carried headers.

The SDK type surface was extended in `sdk/core` v0.3.3 (and Python `apip-sdk-core` 0.2.1). This PR is the policy-engine side that actually populates those fields, completing the snapshot model.

Resolves #2914

## Goals

Expose the pre-mutation request path/method/authority/scheme and the upstream response status code to every policy phase, for both Go and Python policies, symmetrically with how header snapshots already work.

## Approach

- Bump `sdk/core` v0.3.2 → v0.3.3.
- `buildRequestContexts`: snapshot `path`/`method`/`authority`/`scheme` onto `DownstreamRequest` before any policy mutation (captured from the pseudo-headers, before `syncRequestPseudoHeaders` can rewrite the ctx fields).
- `toResponseUpstream`: thread the parsed upstream `:status` into `UpstreamResponse.StatusCode`.
- Extend `python_executor.proto` (add the 4 downstream scalars + `status_code`) and regenerate the Go and Python bindings; wire both bridge translators so Python policies get the same fields.
- Bump `python-executor` pin `apip-sdk-core ~=0.2.0` → `~=0.2.1`.
- No gateway-controller/Envoy change needed — Envoy already sends these as standard pseudo-headers; the kernel only had to surface already-captured values.

## User stories

As a policy author, when an earlier policy rewrites the request path/method or the response status, I can still read the original client-sent path/method/authority/scheme and the original upstream status from the downstream/upstream snapshot.

## Documentation

N/A — additive fields on existing SDK context types; covered by the SDK type docs. No behavior change for existing policies.

## Automation tests
 - Unit tests
   > Kernel `downstream_upstream_test.go`: asserts the new scalar fields populate on all request/response contexts, plus a regression that downstream path/method survive a later path/method rewrite. Python `test_translator.py`: asserts the downstream scalars and upstream `status_code` translate across the proto bridge.
 - Integration tests
   > N/A — no new endpoint/route behavior; the snapshot is a read-only context surface exercised by the unit tests above.

## Security checks
 - Ran FindSecurityBugs plugin and verified report? no (Go/Python change; not a Java module)

## Samples

N/A

## Related PRs

- wso2/api-platform#2974 — the SDK-only change (v0.3.3) that introduced these fields; released independently and consumed here.

## Checklist
- [x] Tests added or updated (unit, integration, etc.)

## Remarks

The regenerated proto bindings require a specific toolchain (protoc 6.30.2, protoc-gen-go 1.36.11, protoc-gen-go-grpc 1.6.1, grpcio-tools 1.76.0) to match CI's `test-proto` git-diff check; generated with those exact versions.


renuka-fernando:
@coderabbitai review

Krishanx92 (dismissed):

Krishanx92 (approved):

Issue #2918, opened 2026-07-28
[Improvement]: Make onboarding "Deploy to Gateway" step reflect real deployment status, and fix Available Proxies cards' clickability


Tharanidk:
### Please select the area the issue is related to

AI Workspace

### Please select the aspect the issue is related to

Aspect/UX (User experience, flows, usability, clarity)

## Improvement 1 — "Deploy to Gateway" onboarding step should require a *successful* deployment

On the onboarding banner (the N/4 step tracker) of the provider, the **Deploy to Gateway** step gets a completion tick as soon as the first Deploy is clicked and a deployment *record* exists, regardless of its status. So even when the deployment **fails**, the step still shows as completed.

<img width="1672" height="741" alt="Image" src="https://github.com/user-attachments/assets/1d6e5328-b77b-4e68-bcf7-d6b9be8fc8f9" />

**Suggested improvement**
Mark the step complete only when there is **at least one active/successful deployment**. A failed or in-progress deploy should leave the step incomplete.

## Improvement 2 — "Available Proxies" cards look clickable but do nothing

In a provider's **Available Proxies** drawer, each proxy card (e.g. "tesrddd", "proxy7") renders as interactive. It has a pointer/hand cursor, a hover state, and a ↗ (up-right arrow) icon that signals "open"  but clicking it does nothing (no navigation, no feedback). This is inconsistent: the card invites a click it never honors.

<img width="761" height="556" alt="Image" src="https://github.com/user-attachments/assets/509cbce5-4b7b-4763-9ee7-1226b920e9d8" />

**Suggested improvement** 
- Make the card actually navigate to the proxy view using the proxy handle 
- If navigation isn't intended, render the card as a non-interactive view (remove the pointer cursor, hover state, and ↗ icon) so it doesn't imply clickability.

PR #2988, opened 2026-07-29
[AIWS] Refactor project ID handling to use project handles instead of UUIDs &  fix UI Issues

Thushani-Jayasekera:
- https://github.com/wso2/api-platform/issues/2918

## Breaking Changes

This PR changes behaviour that existing clients may depend on. The two items in this section are breaks; additive changes are listed separately below, and doc/spec wording updates are omitted.

---

## 1. `projectId` in LLM Proxy and MCP Proxy responses is now a project handle, not a UUID

**Affected responses**

| Schema | Endpoints |
|---|---|
| `LLMProxy` | `GET/POST/PUT /llm-proxies/{proxyId}` and equivalents |
| `LLMProxyListItem` | `GET /llm-proxies`, `GET /llm-providers/{providerId}/llm-proxies` |
| `MCPProxy` | `GET/POST/PUT /mcp-proxies/{id}` and equivalents |
| `MCPProxyListItem` | `GET /mcp-proxies`, `GET /projects/{projectId}/mcp-proxies` |

**Before**

```json
{ "projectId": "550e8400-e29b-41d4-a716-446655440000" }
```

**After**

```json
{ "projectId": "default-project" }
```

**Why:** every other id-shaped field in this API is a handle, and requests already identify a project by handle (`Create`, `List`, `ListByProject`). A client receiving a UUID had no way to resolve it back to a project, so the response value could not be fed into any other call.

**Impact / migration**

- Any consumer that stored, compared, or round-tripped `projectId` as a UUID must switch to the handle. A previously persisted UUID will no longer match the value returned by these endpoints.
- Consumers that passed a response `projectId` straight into a project-scoped request now work correctly (this was previously broken).
- The OpenAPI schema for `projectId` is now constrained to `^[a-z0-9-]+$`, length 3–63 — a UUID no longer satisfies the documented contract.

Implemented via `LLMProxyService.resolveProjectHandle` / `MCPProxyService.resolveProjectHandle`, with a per-request memo so list endpoints don't re-query the same project per item.

---

## 2. Quick Start — the LLM step no longer completes on a non-`DEPLOYED` deployment

**Component:** `LLLMStepBanner.tsx` (AI Workspace Quick Start)

**Before:** the step was marked complete when *any* deployment record existed (`list.length > 0`) — including `DEPLOYING`, `UNDEPLOYED`, and `FAILED`.

**After:** the step is complete only when at least one deployment has `status === 'DEPLOYED'`.

**Impact:** users whose deploy failed or is still in progress will now see the Quick Start LLM step as incomplete where it previously showed as done. This is the intended state, but it is a visible regression in "progress" for anyone in that situation.

---

# Additive Changes (not breaking)

## Proxy card navigation now explains itself instead of failing silently

**Components:** `ProviderMapTab.tsx`, `ServiceProviderOverviewTab.tsx`, `ResourceDrawerCards.tsx`, new `buildProxyPath` helper in `utils/projectRouting.ts`

**Before:** clicking a proxy card whose `projectId` did not match any project in the current organization logged an error and did nothing — the click was silently swallowed with no user-visible feedback.

**After:** the navigation outcome is unchanged (the user stays on the page — a proxy lives inside a project, so there is no correct org-level route to send them to), but an error toast now states why. `buildProxyPath` returns `null` when the project can't be resolved, and both callers surface that rather than navigating.

**Permissions:** the click is gated on `ap:llm_proxy:read` (`SCOPES.LLM_PROXY_READ`), mirroring the gateway-card gating below. Without the scope the card renders non-interactive rather than navigating to a view whose fetch would fail. `ResourceDrawerCards`'s `onProxyClick` became optional for this.

**Note:** the resolution path only behaves correctly in combination with change 1 — matching a proxy's project by handle requires `projectId` to *be* a handle. Before change 1, a UUID `projectId` never matched, so every such click hit the failure branch.

---

## Gateway cards in the provider view are now clickable

**Components:** `ProviderMapTab.tsx` (`GatewayCard`), `ResourceDrawerCards.tsx`, `ServiceProviderOverviewTab.tsx`, new `buildGatewayPath` helper in `utils/projectRouting.ts`

Gateway cards — in both the provider map's gateway column and the "Deployed Gateways" drawer — previously rendered as static, non-interactive cards. They now navigate to the gateway detail view, matching the behaviour proxy cards already had in the same views.

- `GatewayCard` takes an optional `onClick` and, when present, applies the same affordances `ProxyCard` already used: `role="button"`, `tabIndex={0}`, Enter/Space keyboard activation, and a pointer cursor.
- Navigation targets the **existing** `gateways/view/:gatewayName` route (`GatewaysLayout.tsx`). No new route was added. `ViewGateway` resolves that param via `gw.id === gatewayName`, i.e. the gateway **handle** — the same target `GatewaysList` and `AddGateway` already navigate to.
- Gateways are organization-scoped, so `buildGatewayPath` always produces an org-level path (`/organizations/{orgHandle}/gateways/view/{gatewayId}`); there is no project to resolve, unlike `buildProxyPath`.

**Permissions:** the click is gated on `ap:gateway:read` (`SCOPES.GATEWAY_READ`). Without the scope the card renders exactly as before — non-interactive — rather than navigating to a view whose fetch would then fail.

> Note: `GatewaysList` gates its equivalent row-click on `GATEWAY_CREATE`, which it uses as a general "isAdmin" flag also driving the create button and column layout. `GATEWAY_READ` is the semantically correct scope for a view-only navigation; flag in review if you'd prefer these matched exactly.

**Impact:** purely additive — no previously-working interaction changes. Nothing outside the provider view is affected; `LLMProxiesSummaryCardSection`'s `onProxyClick` (a different, single-argument signature) is untouched.

---

# Not a change in this PR

The `gatewayId` spec updates (path params on LLM provider/proxy undeploy & restore, and the `gatewayIdQ` deployment filter) now document a **handle** rather than a UUID. This is the spec catching up to behaviour already shipped in `resolveGatewayFilter` and `toAPIDeploymentResponse` — no runtime behaviour changes here.

---

# Verification

- `tsc --noEmit` on `portals/ai-workspace`: 41 pre-existing errors before these changes, 41 after — none in any touched file. (The project's typecheck is not clean on `main`; the count is unchanged rather than zero.)
- The AI Workspace UI was already handle-based on both the request side (`LLMProxyNew`, `ExternalServersNew`, `ApplicationNew`, and the project-scoped list contexts all send `project.id`) and the routing side (`getProjectSlug` returns `project.id`). Change 1 is what makes the response side agree with them.
- Not yet exercised in a running app — the provider map gateway column and the "Deployed Gateways" drawer are both worth a manual click-through.

### DIFF
--- a/docs/rest-apis/gateway/certificate-management.md
+++ b/docs/rest-apis/gateway/certificate-management.md
@@ -95,7 +95,7 @@ Upload a new TLS certificate (PEM format) to the Gateway. The certificate is loa
 <aside class="warning">
 This operation requires <strong>Basic Auth</strong> authentication.
 
-Required roles: `admin`, `developer`
+Required roles: `admin`
 
 </aside>
 
--- a/docs/rest-apis/platform-api/README.md
+++ b/docs/rest-apis/platform-api/README.md
@@ -1,0 +1,213 @@
+
+<h1 id="wso2-api-platform-platform-api">WSO2 API Platform - Platform API v0.9.0</h1>
+
+This document specifies a **RESTful API** for WSO2 **API Platform** - **Platform API**.
+
+## Authentication
+
+Most API endpoints require a Bearer JWT token in the `Authorization` header. Tokens are
+validated against a configured Identity Provider (IDP) using JWKS-based signature
+verification.
+
+**Required claim**: `organization` — UUID of the caller's organization. All operations are
+automatically scoped to this organization.
+
+**Supported signing algorithms**: RS256, RS384, RS512 (RSA) and ES256, ES384, ES512 (ECDSA).
+Works with any standards-compliant IDP (Keycloak, Azure AD, Okta, Auth0, WSO2 IS, etc.).
+
+## Scope-Based Access Control
+
+Authorization is enforced via OAuth2 scopes carried in the JWT. Both read (GET) and write
+operations require specific scopes declared in the `security` field of each operation.
+
+Base URLs:
+* <a href="https://localhost:9243/api/v0.9">https://localhost:9243/api/v0.9</a>
+* <a href="https://api.platform.com/api/v0.9">https://api.platform.com/api/v0.9</a>
+
+License: <a href="https://www.apache.org/licenses/LICENSE-2.0.html">Apache 2.0</a>
+
+## Table of Contents
+
+### [Authentication](authentication.md)
+
+- [Overview](authentication.md#overview)
+- [Authentication Modes](authentication.md#authentication-modes)
+- [Required Claims](authentication.md#required-claims)
+- [Authorization](authentication.md#authorization)
+- [Error Responses](authentication.md#error-responses)
+- [Troubleshooting](authentication.md#troubleshooting)
+- [Scope Reference](authentication.md#scope-reference)
+
+### [Organizations](organizations.md)
+
+- [Register a new organization](organizations.md#register-a-new-organization)
+- [List organizations](organizations.md#list-organizations)
+- [Get organization by ID](organizations.md#get-organization-by-id)
+- [Check if organization exists](organizations.md#check-if-organization-exists)
+
+### [Projects](projects.md)
+
+- [Create a new project](projects.md#create-a-new-project)
+- [Get all projects for current user's organization](projects.md#get-all-projects-for-current-users-organization)
+- [Get project by UUID](projects.md#get-project-by-uuid)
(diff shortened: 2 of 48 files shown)


Thushani-Jayasekera:
created a new pr to attach these changes also to an existing PR: https://github.com/Thushani-Jayasekera/api-platform/pull/9/

Issue #2930, opened 2026-07-28
[Improvement]: Need to Include subType to Analytics Event


O-sura:
### Please select the area the issue is related to

Gateway

### Please select the aspect the issue is related to

Aspect/Monitoring (Metrics, observability, health checks, alerts), Aspect/Other (Anything else which does not match above categories)

### Suggested Improvement

Currently in the platform side analytics event, only APIType is included in the event sent to the moesif side. To align with the API manager implementation, the new `subType` field also need to be included. 

Ref: [APIM Docs - Moesif Integration](https://apim.docs.wso2.com/en/latest/monitoring/api-analytics/moesif-analytics/moesif-integration-guide/#apim_event_response)

#### Proposed fix
Map the same APIType property to the subType property as well

PR #2995, opened 2026-07-29
Add subType metadata to API event and ensure consistency with apiType

Tharanidk:
## Purpose
Include a `subType` field in the metadata published to Moesif for API analytics events. It mirrors the existing `apiType` value, keeping the two consistent so downstream analytics can rely on `subType` alongside `apiType`.

Resolves:https://github.com/wso2/api-platform/issues/2930

### DIFF
--- a/gateway/gateway-runtime/policy-engine/internal/analytics/publishers/moesif.go
+++ b/gateway/gateway-runtime/policy-engine/internal/analytics/publishers/moesif.go
@@ -193,6 +193,8 @@ func (m *Moesif) Publish(event *dto.Event) {
 	metadataMap["apiName"] = event.API.APIName
 	metadataMap["apiVersion"] = event.API.APIVersion
 	metadataMap["apiType"] = event.API.APIType
+	// subType mirrors apiType
+	metadataMap["subType"] = event.API.APIType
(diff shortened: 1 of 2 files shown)


gateway/gateway-runtime/policy-engine/internal/analytics/publishers/moesif.go
```diff
@@ -193,6 +193,8 @@ func (m *Moesif) Publish(event *dto.Event) {
 	metadataMap["apiName"] = event.API.APIName
 	metadataMap["apiVersion"] = event.API.APIVersion
 	metadataMap["apiType"] = event.API.APIType
+	// subType mirrors apiType
+	metadataMap["subType"] = event.API.APIType
```
Krishanx92, line 197: Why we push same param to subType as well?

Krishanx92 (changes requested):

Krishanx92 (approved):

gateway/gateway-runtime/policy-engine/internal/analytics/publishers/moesif.go
Krishanx92: Llm table provider name

PR #3002, opened 2026-07-29
[Devportal] Support repeatable --config flag

### LINKED ISSUE #2876: Developer Portal: support repeatable --config flag, make it required
### Description

Adopt the layered multi-config-file pattern in the Developer Portal (`portals/developer-portal`, Node.js/Express) so configuration can be a complete base file plus thin environment/deployment overlays, and unify its config-loading behavior with the platform's Go components. This is the JavaScript counterpart of #2865 (gateway-controller + policy-engine), #2873 (AI Workspace BFF), and #2874 (Platform API).

**Current state**
- `src/config/configLoader.js` loads a single, hardcoded path `configs/config.toml` (`path.join(process.cwd(), 'configs', 'config.toml')`) at module import time, via `require('./config/configLoader')` in `src/server.js`.
- If that file is absent, `loadTomlConfig()` returns `{}` and the portal **silently runs on built-in `DEFAULTS`** — the same silent-defaults fallback the other components removed.
- There is no `--config` flag and no `process.argv` handling. `{{ env }}` / `{{ file }}` interpolation and the `mergeOver` deep-merge (over `DEFAULTS`) already exist, and the `APIP_DP_*` env-prefix override was already removed (env reaches config only via explicit `{{ env }}` tokens).

**Behavior (target)**
- Add a repeatable `--config` flag (mirroring the Go components' `-config`). Files are merged in the order given with **last-wins** precedence (a key set in a later file overrides the same key from an earlier file), layered over `DEFAULTS`.
- Merge semantics: nested tables (maps) deep-merge; list/array values are replaced, not appended — document this so overlays that set list keys replace the base list. Confirm/adjust `mergeOver` so arrays replace rather than merge index-wise.
- `{{ env }}` / `{{ file }}` interpolation runs **once, after** all files are merged, so a token declared in the base can be resolved by a later overlay. Secrets stay in interpolation sources, not in the overlay TOMLs.
- `--config` becomes **required** and there is **no default path / no silent-defaults fallback**: omitting `--config`, or passing a path that does not exist/parse, must fail fast (non-zero exit) at startup rather than silently booting on `DEFAULTS`.

**Rationale for unification**
All the platform's config loaders should behave identically: repeatable `--config`, at least one file required, no default path, no silent-defaults fallback. Because there is no env-var provider (env enters only through `{{ env }}` tokens in a file), a missing config file means "running on pure built-in defaults", which for a portal handling auth/session/secret config must fail fast.

**Notes / scope**
- Config currently materializes at **module import** (`configLoader` exports a resolved `config` object). Moving to a required `--config` flag means resolving argv and loading config during startup (e.g. in `src/server.js`) rather than as an import side effect — a larger refactor than the Go components required.
- Deployment: the container relies on `configs/config.toml` being present (mounted); the `docker-entrypoint.sh` ENTRYPOINT and the `developer-portal-ui-helm-chart` must be updated to pass `--config <path>` explicitly once the flag is required (the same adjustment made for the AI Workspace BFF image in #2873).

**Affected code**
- `portals/developer-portal/src/config/configLoader.js` — accept one or more config paths (from argv), loop-load and deep-merge with last-wins, drop the `fs.existsSync` → `{}` silent fallback.
- `portals/developer-portal/src/server.js` — parse `--config` (repeatable, required) and resolve config at startup instead of relying on the import-time singleton.
- `portals/developer-portal/Dockerfile` / `docker-entrypoint.sh` and `kubernetes/helm/developer-portal-ui-helm-chart` — pass `--config` explicitly.

### Version

_No response_

### Related Issue

#2865, #2873, #2874

Piumal1999:
## Purpose
Fix https://github.com/wso2/api-platform/issues/2876

## Summary

- The portal used to read config from one fixed path, configs/config.toml. The path is now given explicitly: --config <path>.
- --config can be passed several times. Files stack in order, later ones winning per key — so one shared base file plus a small per-environment overlay, instead of a full copy per environment. (Nested tables merge; a list in an overlay replaces the base list, it doesn't append.)
- The flag is required. Before, a missing config file meant the portal quietly started on built-in defaults — bad, since that file holds session secrets, encryption keys, and IdP settings. Now no config (or missing / bad TOML / no [developer_portal] table) prints why and exits 1 before the port is bound.
- Built-in defaults no longer run the app on their own; they only fill gaps for keys the config files don't set.
- {{ env }} / {{ file }} tokens are resolved once, after all files merge, instead of per file. That's what lets an overlay override a value the base file wrote as {{ env "..." }}.
- Everything in this repo that starts the portal passes the flag: the Docker entrypoint (base path overridable via APIP_DP_CONFIG_PATH) and the start / debug npm scripts.

### DIFF
--- a/portals/developer-portal/configs/config-template.toml
+++ b/portals/developer-portal/configs/config-template.toml
@@ -6,9 +6,15 @@
 # concrete file this project actually ships, which wires some of these same
 # keys up to environment variables using the syntax below.
 #
+# Config files are loaded via a required, repeatable `--config <path>` flag —
+# at least one is mandatory (no default path, no silent-defaults fallback), and
+# passing it more than once layers overlays with last-wins precedence. Nested
+# tables deep-merge; list/array values are replaced wholesale by a later file.
+#
 # Precedence (lowest to highest):
 #   1. Built-in defaults (src/config/configDefaults.js)
-#   2. configs/config.toml, with any {{ env }} / {{ file }} references resolved
+#   2. The merged --config files, with any {{ env }} / {{ file }} references
+#      resolved once after the merge
 #
 # There is no automatic environment-variable override. A key's value comes
 # from an env var or a file ONLY if you explicitly say so, using Handlebars-
--- a/portals/developer-portal/configs/config.toml
+++ b/portals/developer-portal/configs/config.toml
@@ -4,9 +4,21 @@
 # see configs/config-template.toml for the full list of available keys and
 # their default values.
 #
+# This file is loaded via a required, repeatable `--config <path>` flag (the JS
+# counterpart to the Go components' `-config`) — e.g. the container entrypoint
+# runs `node src/server.js --config /app/configs/config.toml`. At least one
+# --config is mandatory: there is no default path and no silent fallback to the
+# built-in defaults. Pass --config more than once to layer thin overlays over a
+# base file; later files override earlier ones, key by key.
+#
 # Precedence (lowest to highest):
 #   1. Built-in defaults (src/config/configDefaults.js)
-#   2. This file, with any {{ env }} / {{ file }} references resolved
+#   2. The merged --config files (this file, plus any later overlays), with any
+#      {{ env }} / {{ file }} references resolved once after the merge
+#
+# Merge semantics when layering multiple --config files: nested tables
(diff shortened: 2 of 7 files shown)


Piumal1999:
@coderabbitai review

lasanthaS (approved):

PR #3021, opened 2026-07-30
fix(platform-api): release custom policy in-use lock on artifact delete

### LINKED ISSUE #3008: [Bug]: Deleting a provider orphans the policy's "in use" lock forever

AI Workspace

### Description

Once a custom policy is attached to an LLM provider and deployed, the policy is correctly locked from deletion (`409 POLICY_IN_USE`). However, deleting the provider that held the policy does not release the lock. The policy remains permanently stuck in the "in use" state — with zero providers referencing it — and can never be deleted.

## Steps to Reproduce

1. Create an LLM provider that references the custom policy.

   ```bash
   curl -sk -X POST https://localhost:9243/api/v0.9/llm-providers \
     -H "Authorization: Bearer $TOKEN" \
     -H "Content-Type: application/json" \
     -d '{
       "id": "issue1-provider",
       "displayName": "Issue1",
       "version": "v1.0",
       "context": "/issue1",
       "template": "openai",
       "upstream": {
         "main": {
           "url": "http://host.docker.internal:8999"
         }
       },
       "accessControl": {
         "mode": "deny_all",
         "exceptions": [
           {
             "path": "/chat/completions",
             "methods": ["POST"]
           }
         ]
       },
       "globalPolicies": [
         {
           "name": "<policy-name>",
           "version": "v1",
           "params": {
             "value": "x"
           }
         }
       ]
     }'
   ```

2. Deploy the provider to ensure the policy is actively in use.

   ```bash
   curl -sk -X POST https://localhost:9243/api/v0.9/llm-providers/issue1-provider/deployments \
     -H "Authorization: Bearer $TOKEN" \
     -H "Content-Type: application/json" \
     -d '{
       "name": "v1",
       "base": "current",
       "gatewayId": "local-ai-gateway"
     }'
   ```

3. Attempt to delete the custom policy.

   **Expected:** Returns `409 POLICY_IN_USE`.

   ```bash
   curl -sk -i -X DELETE \
     "https://localhost:9243/api/v0.9/gateway-custom-policies/$POLICY_UUID/versions/v1.0.0" \
     -H "Authorization: Bearer $TOKEN"
   ```

4. Delete the LLM provider.

   **Expected:** Returns `204 No Content`.

   ```bash
   curl -sk -i -X DELETE \
     https://localhost:9243/api/v0.9/llm-providers/issue1-provider \
     -H "Authorization: Bearer $TOKEN"
   ```

5. Verify that no providers remain.

   ```bash
   curl -sk https://localhost:9243/api/v0.9/llm-providers \
     -H "Authorization: Bearer $TOKEN"
   ```

6. Attempt to delete the custom policy again.

   **Actual:** The request still returns `409 POLICY_IN_USE`, even though the provider has been deleted and no providers reference the policy.

   ```bash
   curl -sk -i -X DELETE \
     "https://localhost:9243/api/v0.9/gateway-custom-policies/$POLICY_UUID/versions/v1.0.0" \
     -H "Authorization: Bearer $TOKEN"
   ```

Severity/Major (Important functionality is broken. Should be prioritized. Doesn't need immediate attention)

_No response_

ashera96:
## Purpose

- Resolves: https://github.com/wso2/api-platform/issues/3008

Deleting an LLM provider that referenced a custom policy failed to release the policy's in-use lock. As a result, the policy remained stuck in a `409 POLICY_IN_USE` state with zero providers referencing it, making it impossible to delete.

### Root Cause

While `gateway_custom_policy_usages.artifact_uuid` has an `ON DELETE CASCADE` constraint on artifacts, foreign-key enforcement was turning itself off at runtime in SQLite environments:

* `PRAGMA foreign_keys = ON` was previously run as a one-shot `db.Exec` against the database pool. However, `go-sqlite3` applies pragmas per physical connection.
* With the connection pool capped at 1 connection and `conn_max_lifetime` defaulting to 300 seconds, the initial connection was retired ~5 minutes after application startup.
* Replacement connections opened straight from the DSN with foreign key enforcement disabled by default, causing all subsequent `ON DELETE CASCADE` triggers to fail silently for the remainder of the process lifecycle.

---

### Database Connection (`internal/database/connection.go`)
* Moved foreign-key enablement directly into the DSN string (`_foreign_keys=on`) so every connection opened by the pool enforces foreign keys automatically.
* Removed the one-shot pragma execution.
* Added a post-`Ping` verification step that reads `PRAGMA foreign_keys` back on a separately-drawn pooled connection to fail loudly at boot if foreign key enforcement is inactive.

### Repository Layer (`internal/repository/`)
* **`artifact.go`**: `ArtifactRepo.Delete` now explicitly purges the artifact's associated `gateway_custom_policy_usages` rows in the same transaction. This covers all artifact types (REST API, LLM provider, LLM proxy, MCP proxy).
* **`custom_policy.go`**: 
  * `DeleteCustomPolicyIfUnused` now runs in a transaction that first purges orphaned usage rows whose parent artifact no longer exists, allowing stuck policies from pre-fix deployments to become deletable post-upgrade without manual database intervention.
  * `CountCustomPolicyUsages` now ignores artifact-less orphan rows.
  * *Note*: Usage rows backed by active artifacts continue to correctly block deletion with `409 POLICY_IN_USE`.

### Service Layer (`internal/service/gateway.go`)
* Added warning logs indicating the count of purged orphan policy locks to provide visibility during automatic upgrades.

---

## Testing

* **`TestNewConnection_SQLiteForeignKeysSurviveConnectionRecycling`**: Regression test that forces connection recycling via `SetMaxIdleConns(0)` and asserts foreign-key enforcement and cascades remain active on recycled connections.
* **`TestCascade_DeleteArtifactRemovesCustomPolicyUsages`**: Integration test across all three supported database dialects verifying that deleting an artifact releases its policy lock.
* **`TestDeleteCustomPolicyIfUnusedPurgesOrphanUsages`**: Verifies that pre-existing stale orphan rows are cleaned up without blocking policy deletion.
* **`TestDeleteCustomPolicyIfUnusedBlocksWhenArtifactExists`**: Confirms that active artifact usages continue to enforce the `POLICY_IN_USE` lock.
* **`TestCountCustomPolicyUsagesIgnoresOrphans`**: Verifies usage counts ignore orphan rows.

---

## Notes

The foreign-key enforcement gap was global across SQLite tables (all `ON DELETE CASCADE` statements were inert after connection recycling). The DSN configuration fix resolves this behavior platform-wide going forward.


Thushani-Jayasekera (approved):

renuka-fernando:
Are we relying on `ON DELETE CASCADE` for this or deleting from code itself?

Thushani-Jayasekera:
> Are we relying on `ON DELETE CASCADE` for this or deleting from code itself?

This is implemented at both levels.

ashera96:
We have done both, code level deletion is also handled

PR #3129, opened 2026-08-04
Add Secret Management UI Implementation

npamudika:
## Purpose
Platform API already supports encrypting upstream credentials for LLM providers, LLM proxies, MCP proxies, and REST APIs via {{ secret "handle" }} placeholders (added in https://github.com/wso2/api-platform/issues/2149, but two gaps made that feature incomplete in practice:

- No UI to manage secrets directly. The only way to create, rotate, or delete a secret was a raw API/CLI call — there was no screen in AI Workspace for it, "no dedicated Secrets screen in any portal").
- Rotation/deletion wasn't live. A gateway only picked up a rotated or deleted secret on its next reconnect-triggered poll sync. An already-connected gateway kept serving the old credential — e.g. rotating a leaked API key didn't actually stop it being used until that gateway happened to reconnect or restart.
- While building and testing the fix for (2), a related latent bug surfaced: BuildAPIDeploymentYAML silently dropped upstream.main/sandbox.auth from the REST API deployment YAML, so secret-backed REST API upstream credentials never reached the gateway at all — it just quietly forwarded requests with no auth header.
- Hard delete instead of soft-delete. SecretRepository.FindRefsAndSoftDelete → FindRefsAndDelete: same transactional lock-then-check pattern, but the final step is now DELETE FROM secrets instead of UPDATE ... SET status='DEPRECATED'. This alone fixes your second point too — cleanupRotatedSecret (which already ran this same check-then-delete after a rotation) now actually removes the old secret's row when nothing else references it, instead of leaving it around forever as DEPRECATED.
- Orphan cleanup on resource deletion. LLMProviderRepo.Delete, LLMProxyRepo.Delete, and MCPProxyRepo.Delete now capture every secret handle the resource referenced before removing it, and return that list. Each corresponding service's Delete method then calls a new best-effort SecretService.CleanupOrphanedSecrets, which re-checks each handle and permanently deletes any that are no longer referenced by anything else — silently skipping ones still in use.

## Resolves: 
https://github.com/wso2/api-platform/issues/2149

## Goals
- Ship a dedicated Secrets UI in AI Workspace (list/search, create, view, rotate, delete) under Organization Settings, scope-gated on ap:secret:read.
- Push secret.updated / secret.deprecated events from platform-api to every connected gateway-controller over the existing control-plane WebSocket, so rotation/deletion takes effect in about a second instead of waiting for the next reconnect.
- Fix the REST API deployment YAML gap so secret-backed REST API upstreams actually work.
- Close a set of correctness, UX, and hardening gaps found via a dedicated live-testing pass against a real running stack (docker platform-api + ai-workspace + gateway-controller + gateway-runtime): a rotate-form field-name mismatch that silently dropped display-name edits, missing server-side handle validation (a handle containing / was creatable but then permanently unreachable via the API), a broadcast-ordering bug, an unhelpful validation error message, and frontend accessibility/robustness gaps.

### Backend (Go, platform-api):

- SecretService — full CRUD, broadcastSecretEvent (best-effort, fans out to every gateway in the org via GatewayEventsService), server-side handle validation (constants.SecretHandlePattern / SecretHandleMaxLength, mirroring the frontend's slug rule), and ValidateSecretRefs now distinguishes a missing handle from a deprecated one in its error message.
- GatewayEventsService.BroadcastSecretUpdatedEvent / BroadcastSecretDeprecatedEvent — new event types on the existing EventHub WebSocket channel used for all other control-plane push events.
dto.UpstreamTarget gained an Auth field, and BuildAPIDeploymentYAML now copies it through — the deployment-YAML fix.

### Backend (Go, gateway-controller):

- client.go dispatch: handleSecretUpdatedEvent fetches the new plaintext on demand and re-encrypts it into local storage; handleSecretDeprecatedEvent evicts the local copy. Both fall back cleanly to the existing poll-based sync if the push is missed (disconnected gateway, delivery failure, etc.).

### Frontend (AI Workspace):

- Five new pages under portals/ai-workspace/src/pages/appShell/appShellPages/secret/: SecretsList, CreateSecret, SecretOverview, RotateSecret, DeleteSecretDialog — wired into App.tsx routing and the Settings nav.
- Accessibility: list rows are keyboard-navigable (role="link", tabIndex, Enter/Space handling, aria-label).
- Hardening: autoComplete="new-password" on credential inputs (avoids browser password-manager prompts), and a new useIsMounted hook guards state updates in Create/Rotate/Delete after the component has unmounted.

## UIs
Screenshot 2026-08-10 at 10 50 59
Screenshot 2026-08-10 at 10 51 15
Screenshot 2026-08-10 at 10 52 43
Screenshot 2026-08-10 at 10 52 59
Screenshot 2026-08-10 at 10 53 14
Screenshot 2026-08-10 at 10 53 24
Screenshot 2026-08-10 at 10 54 16

## UI Recording
https://github.com/user-attachments/assets/0d42c43f-4cac-4d33-a11b-3df5ee6a6d8a

## User Stories
- As an AI Workspace admin, I can create, view, rotate, and delete secrets from the UI without needing API/CLI access.
- As an admin rotating a leaked or expiring credential, the new value takes effect on every already-connected gateway within about a second, not just on that gateway's next reconnect.
- As an admin trying to delete a secret still in use, I see exactly which resources reference it instead of a generic failure.
- As a keyboard/assistive-technology user, I can navigate and operate the Secrets list without a mouse.

## Documentation
https://github.com/wso2/api-platform/blob/main/docs/rest-apis/platform-api/secrets.md

## Automation Tests
- Unit tests (Go): 42 test functions across the secret-management surface
- Integration/e2e tests (Cucumber/godog, tests/integration-e2e/):

1. features/secret_lifecycle.feature (2 scenarios) — rotation and deletion pushing live to an already-connected gateway, verified against the gateway's own management API.
2. features/rest_api_secret.feature, features/policy_secret.feature (1 each) — on-demand secret resolution at deploy time for REST APIs and policy params.
3. Additionally hand-verified live against a real docker stack (platform-api + ai-workspace + gateway-controller + gateway-runtime) this session: full create/rotate/delete UI flow, XSS-payload rendering safety, delete-blocked-by-reference dialog, disconnected-gateway reconnect catch-up, and every fix above reproduced pre-fix and confirmed post-fix.

## Related PRs
Refer the PRs mentioned in https://github.com/wso2/api-platform/issues/2149

## Test Environment
- Go: 1.26.5 (per platform-api/go.mod)
- Node: 20.20.2 (per portals/ai-workspace/.nvmrc)
- OS: macOS 15.7.3 (Darwin 24.6.0, arm64)
- Database: SQLite (local dev); the shared ValidateSecretRefs/handle-validation logic is DB-agnostic and covered against SQLite in tests — not separately re-verified against PostgreSQL/SQL Server in this pass.
- Browser: Chromium-based embedded test browser (headless automation) against the built AI Workspace frontend bundle.
- Containers: Docker Compose stack — platform-api, ai-workspace, gateway-controller, gateway-runtime, sample-backend, all built from this branch's source.

### DIFF
--- a/gateway/gateway-controller/pkg/controlplane/client.go
+++ b/gateway/gateway-controller/pkg/controlplane/client.go
@@ -159,6 +159,7 @@ type Client struct {
 	webhookSecretSnapshotManager WebhookSecretSnapshotRefresher
 	secretSyncer                 secretSyncer
 	secretHashCache              sync.Map // handle → last-known Platform API hash (string)
+	secretRevisionCache          sync.Map // handle → last-applied event Revision (int64); never cleared on evict, so a stale event can't undo a later one
 	eventGatewayHooks            ControlPlaneEventGatewayHooks
 
 	// DP->CP push retry tuning.
@@ -3786,6 +3787,10 @@ func (c *Client) handleSecretUpdatedEvent(event map[string]interface{}) {
 // *utils.APIUtilsService (mirroring syncSecretsIncrementalFromMetas in
 // sync_secrets_test.go, which works around the same constraint).
 func (c *Client) applySecretUpdatedPayload(payload SecretUpdatedEventPayload, logger *slog.Logger, fetchValue func(handle string) (string, error)) {
+	if c.isStaleSecretEvent(payload.Handle, payload.Revision, logger) {
+		return
+	}
+
 	plaintext, err := fetchValue(payload.Handle)
 	if err != nil {
 		logger.Error("Failed to fetch rotated secret value", slog.Any("error", err))
@@ -3798,7 +3803,28 @@ func (c *Client) applySecretUpdatedPayload(payload SecretUpdatedEventPayload, lo
 	}
 
 	c.secretHashCache.Store(payload.Handle, payload.Hash)
+	c.secretRevisionCache.Store(payload.Handle, payload.Revision)
 	logger.Info("Applied secret rotation from secret.updated event")
+}
+
+// isStaleSecretEvent reports whether revision is older than the last revision this
+// gateway already applied for handle. The comparison is strict-less-than so a
+// redelivery of the same event (revision equal to the cached one — e.g. at-least-once
+// retry from the EventHub) still applies normally, preserving existing idempotency.
+// The cache entry is never cleared on eviction (see handleSecretDeprecatedEvent), so a
+// deletion followed by a reactivation followed by a stale, redelivered deletion cannot
+// undo the reactivation: the stale deletion's revision is lower than the reactivation's.
+func (c *Client) isStaleSecretEvent(handle string, revision int64, logger *slog.Logger) bool {
+	if cached, ok := c.secretRevisionCache.Load(handle); ok {
+		if lastApplied, ok := cached.(int64); ok && revision < lastApplied {
+			logger.Warn("Ignoring stale secret event",
+				slog.Int64("event_revision", revision),
+				slog.Int64("last_applied_revision", lastApplied),
+			)
+			return true
+		}
+	}
+	return false
 }
 
 // handleSecretDeprecatedEvent processes secret.deprecated events, pushed when a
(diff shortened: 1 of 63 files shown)


seminalworks:
Reviewed the Secrets UI flow end to end. A few things worth addressing before this merges.

**CTA placement**
On Settings, Secrets, the Create button currently sits inline above the search bar. For consistency with how create actions are placed elsewhere in the app, move it to the top right of the page header instead.

**Two left navs**
Two fully visible left navs at once is not recommended. Settings currently has 3 items, well under the 5-to-7 range where a second sidebar starts paying for itself. If trivial, I'd recommend a drop-up menu instead of a full second nav at this point. If not feasible, then the primary nav should collapse to an icon rail while the Settings nav is open.

**Contextual entry point (P2)**
Putting these items in Settings is fine; though we should consider contextual discoverability. I've attached two examples below (note the Secrets link is shown here before typing):
image
Moreover, we should facilitate creating these items within the flows themselves, rather than having to navigate away from the relevant context. See example below—perhaps the user could click one of these links for a popup where they could title the secret:
image

**Empty state and RBAC**
More of an open question than feedback, but do we have empty states for theses pages? Is their ability role-based? If so, what would each role see? Let me know if you have any mockups to review, or need input.

PR #3147, opened 2026-08-05
feat(sdk): add snapshot header writers for request/response contexts

renuka-fernando:
## Purpose

Policies can read the pre-mutation client-request snapshot via `DownstreamHeaders()` and the upstream-response snapshot via `UpstreamHeaders()`, but there was no sanctioned way to *write* those snapshots. The only option was the kernel-only `UnsafeInternalValues()` escape hatch, which policies must never call. This blocked legitimate inter-policy coordination — for example, an earlier policy injecting a header that a later JWT/auth policy validates against the client-request snapshot.

## Goals

Give policies a safe, first-class API to mutate the request/response snapshots that later policies in the same request read, without exposing the unsafe internal map and without changing what is sent to the backend or the client.

## Approach

- Add imperative, chain-internal setters that mirror the existing read accessors:
  - `SetDownstreamHeader` / `AddDownstreamHeader` / `RemoveDownstreamHeader` on `RequestHeaderContext` and `RequestContext` (client-request snapshot).
  - `SetUpstreamHeader` / `AddUpstreamHeader` / `RemoveUpstreamHeader` on `ResponseHeaderContext` and `ResponseContext` (upstream-response snapshot).
- These follow the same imperative, chain-internal pattern already used for `SharedContext.Metadata`. The kernel shares one snapshot object across every phase context of a request, so a header-phase write is visible to the body/stream phases, and a request-phase downstream write is visible to the response phase.
- They are never translated into an Envoy header mutation: to change what the backend receives, return `UpstreamRequestModifications.HeadersToSet`; to change what the client receives, return `DownstreamResponseModifications`.
- Snapshot-only, never the live fallback: unlike the read accessors, the setters do not fall back to the live working headers when the gateway provides no snapshot, since mutating live request headers would leak to the backend and mutating live response headers would leak to the client. They no-op instead.
- `Headers` stays read-only for policies; `UnsafeInternalValues()` remains confined to these SDK-internal helpers. This is a pure SDK change — no policy-engine code is touched, because the engine already builds a stable, shared, mutable snapshot across all phase contexts.

## User stories

As a policy author, I want to inject or modify a header on the client-request snapshot so that a later policy (e.g. JWT validation) reads the value I set, without resorting to unsafe internal APIs.

## Documentation

N/A — SDK-level API addition; policy authoring docs can reference the new setters alongside the existing `DownstreamHeaders()`/`UpstreamHeaders()` accessors in a follow-up.

## Automation tests

 - Unit tests
   > Added `context_header_writers_test.go` covering: read-back through `DownstreamHeaders()`/`UpstreamHeaders()`, the guarantee that a snapshot write does not leak into the live working headers (backend/client), no-op behavior when the gateway provides no snapshot, lazy `Headers` creation on a header-less snapshot, and Add/Remove semantics.
 - Integration tests
   > N/A for this change — no policy-engine or wire behavior changes. Existing SDK package tests pass (`go test ./core/policy/v1alpha2/`), `go vet` clean.

## Security checks

 - Ran FindSecurityBugs plugin and verified report? no — Go SDK change; FindSecurityBugs is a Java tool and not applicable.

## Samples

N/A — no sample changes in this PR. A follow-up can wire a JWT/sample policy to demonstrate `SetDownstreamHeader` usage.

## Related PRs

N/A

## Test environment

Go (SDK module `sdk/core/policy/v1alpha2`); macOS (darwin). Verified via `go build`, `go vet`, and `go test`.

## Checklist

### DIFF
--- a/sdk/core/policy/v1alpha2/context_header_writers.go
+++ b/sdk/core/policy/v1alpha2/context_header_writers.go
@@ -1,0 +1,226 @@
+/*
+ *  Copyright (c) 2026, WSO2 LLC. (http://www.wso2.com) All Rights Reserved.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ */
+
+package policyv1alpha2
+
+import "strings"
+
+// This file provides the WRITE counterpart of the read accessors in
+// context_accessors.go. It lets an earlier policy influence what a LATER policy
+// in the same request observes when it reads the pre-mutation snapshot via
+// DownstreamHeaders() (request phase) or UpstreamHeaders() (response phase) —
+// for example, injecting a header that a downstream JWT/auth policy validates
+// against the client-request snapshot.
+//
+// These setters are deliberately IMPERATIVE and CHAIN-INTERNAL, mirroring how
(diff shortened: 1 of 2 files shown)


Thushani-Jayasekera (approved):

Krishanx92:
@coderabbitai review

PR #3212, opened 2026-08-12
Bypass JWT signature validation via a build-time flag instead of config

### LINKED ISSUE #3211: [Improvement]: Add a config option to disable JWT signature validation in internal_token mode

Platform API

### Problem
In `internal_token` auth mode, JWT signature validation is always on and cannot
be turned off. There is no configuration to accept a token without verifying its
signature, even in deployments where verification is intentionally not wanted.

### Request
Add an `auth.jwt.skip_validation` boolean config option. When enabled in
`internal_token` mode, the platform API should parse the token for its claims but
skip signature and issuer verification (so no public key is required), while still
requiring the organization claim.

### Notes
- Must default to `false` — validation stays strict unless explicitly disabled.
- Must be ignored in `file` and `idp` modes.
- The option is dangerous on an internet-facing listener, so it should log a loud
  warning at startup when enabled and be clearly documented as such.

### Related Issues

_No response_

dakshina99:
## Purpose

Some deployments front `platform-api` (in `internal_token` mode) with a **trusted mediation layer on a private network** that has already authenticated the caller and forwards an unsigned internal token carrying only the org context. Those deployments need `platform-api` to read the token's claims without verifying its signature.

The concern is how that bypass is *exposed*. A `config.toml` field (`auth.jwt.skip_validation`) is a runtime toggle any operator could flip — including on an internet-facing listener, where it would let anyone forge claims. Signature-bypass should not be an operator-flippable setting at all.

## Goals

Make the signature-validation bypass a **property of a specific build**, not a runtime configuration option — so no configuration change on a normally-built binary can ever disable signature validation.

## Approach

The bypass is now a **build-time variable stamped in via ldflags**, mirroring how the binary's version is injected:

```
-X github.com/wso2/api-platform/platform-api/config.skipJWTValidation=true
```

- `config.skipJWTValidation` is an unexported package-scope string var; `config.SkipJWTValidation()` reports whether the build set it to `true` (case-insensitive, trimmed).
- The empty default that every normal build carries keeps strict RS256 validation. There is no `config.toml` key — the field was removed from the `JWT` config struct and the config template.
- `buildAuthenticator` reads `config.SkipJWTValidation()` in `internal_token` mode; when true it logs a prominent warning and wires the local-JWT middleware to accept unsigned tokens (claims still read; org claim still required). When false it loads the public key and validates as before. File mode is unaffected — it is handled by a separate branch that always loads the public key and validates strictly.

This replaces the earlier `auth.jwt.skip_validation` config field from this PR's first revision.

## Documentation

N/A — the removed config key was documented inline in `config-template.toml`; that block is replaced with a note pointing to the build-time flag.

## Automation tests

- Unit tests
  > `config`: `TestSkipJWTValidation` covers the accessor's parsing (empty/false/arbitrary → strict; `true`/`TRUE`/padded → bypass). `middleware`: existing `auth_skip_validation_test.go` covers the middleware behavior (unsigned token accepted + org resolved with bypass; missing org claim still 401; unsigned rejected under strict). `go build ./...`, `go test ./config/... ./internal/middleware/...` pass; symbol verified linkable under the ldflag.
- Integration tests
  > N/A — covered by unit tests at the config and middleware layers.

## Security checks

- Followed secure coding standards? **yes** — the bypass is no longer runtime-configurable; a normal build cannot disable signature validation, and the default is strict RS256. When bypass is compiled in, the org claim is still required and a loud warning is logged at startup.
- Ran FindSecurityBugs plugin and verified report? **N/A** (Go module; not a Java/FindSecurityBugs target).
- Confirmed that this PR doesn't commit any keys, passwords, tokens, usernames, or other secrets? **yes**.

## Related PRs

N/A — the consuming build passes the ldflag from its own build tooling.

## Test environment

Go (module build); linux/amd64 + linux/arm64 build targets.

- Resolves https://github.com/wso2/api-platform/issues/3211

### DIFF
--- a/platform-api/config/config.go
+++ b/platform-api/config/config.go
@@ -330,6 +330,37 @@ type JWT struct {
 	TokenTTL       time.Duration `koanf:"token_ttl"`
 }
 
+// skipJWTValidation is stamped in at BUILD time via ldflags, exactly like the
+// binary's version (-X ...config.skipJWTValidation=true) — it is deliberately NOT
+// a config-file field. Disabling JWT signature validation is a property of a
+// specific build (the cloud build fronted by a trusted mediation layer on a
+// private network that has already authenticated the caller and forwards an
+// unsigned internal token carrying the org context), never a runtime toggle an
+// operator could flip on an internet-facing deployment. The empty default that
+// every normal build carries means strict validation.
+var skipJWTValidation string
+
+// skipJWTValidationEnabled is the parsed skipJWTValidation, evaluated once at
+// package initialization (the ldflags value is fixed at link time) so
+// SkipJWTValidation() is a plain field read on the hot authentication path.
+var skipJWTValidationEnabled = parseSkipJWTValidation(skipJWTValidation)
+
+// parseSkipJWTValidation reports whether the build-time flag value enables the
+// bypass: true only for the literal "true" (case-insensitive, surrounding space
+// trimmed); any other value keeps strict validation.
+func parseSkipJWTValidation(v string) bool {
+	return strings.EqualFold(strings.TrimSpace(v), "true")
+}
+
+// SkipJWTValidation reports whether this build disables JWT signature and issuer
+// verification in "internal_token" mode (see skipJWTValidation). DANGEROUS: it
+// makes unsigned ("none") tokens acceptable, so it must only be true in a build
+// deployed behind a trusted mediation layer on a private network. Ignored in
+// "file" and "idp" modes.
+func SkipJWTValidation() bool {
+	return skipJWTValidationEnabled
+}
+
 // LoadPublicKey reads and parses the PEM-encoded RSA public key from
 // PublicKeyFile. The file is read fresh on every call rather than cached,
 // so PublicKeyFile is a mounted-file path, never inlined PEM content.
--- a/platform-api/config/skip_jwt_validation_test.go
+++ b/platform-api/config/skip_jwt_validation_test.go
@@ -1,0 +1,45 @@
+/*
+ *  Copyright (c) 2025, WSO2 LLC. (http://www.wso2.org) All Rights Reserved.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ */
+
+package config
+
+import "testing"
+
+// parseSkipJWTValidation backs the ldflags-stamped build var. It is true only for
+// the literal "true" (case-insensitive, surrounding space trimmed) — the empty
+// value every normal build carries, and any other value, keep strict validation.
+func TestParseSkipJWTValidation(t *testing.T) {
+	cases := []struct {
+		name string
+		val  string
+		want bool
+	}{
+		{"empty default is strict", "", false},
+		{"explicit false", "false", false},
+		{"arbitrary value is strict", "1", false},
+		{"true enables bypass", "true", true},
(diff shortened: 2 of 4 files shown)


malinthaprasan:
Shall we use ldflags instead of using config.toml config? 
eg: https://github.com/wso2/api-platform/blob/main/gateway/gateway-builder/cmd/builder/main.go#L48-L53
cc: @renuka-fernando

platform-api/config/config.go
```diff
@@ -328,6 +328,15 @@ type JWT struct {
...
 	TokenTTL       time.Duration `koanf:"token_ttl"`
+	// SkipValidation disables JWT signature and issuer verification in
+	// "internal_token" mode: the token is parsed for its claims but its
+	// signature is not checked, so unsigned ("none") tokens are accepted and no
+	// public key is loaded. This exists ONLY for deployments where a trusted
+	// mediation layer on a private network has already authenticated the caller
+	// and forwards an unsigned internal token carrying the org context. It is
+	// DANGEROUS on any internet-facing listener — anyone can then forge claims.
+	// Defaults to false; ignored in "file" and "idp" modes.
+	SkipValidation bool `koanf:"skip_validation"`
```
renuka-fernando: I was thinking of not making this a config, instead a build-time linker flag which set a package level bool.
So in the On Prem Platform API, there is no skip validation

platform-api/config/config.go
```diff
@@ -330,6 +330,25 @@ type JWT struct {
...
+// unsigned internal token carrying the org context), never a runtime toggle an
+// operator could flip on an internet-facing deployment. The empty default that
+// every normal build carries means strict validation.
+var skipJWTValidation string
+
+// SkipJWTValidation reports whether this build disables JWT signature and issuer
+// verification in "internal_token" mode (see skipJWTValidation). DANGEROUS: it
+// makes unsigned ("none") tokens acceptable, so it must only be true in a build
+// deployed behind a trusted mediation layer on a private network. Ignored in
+// "file" and "idp" modes.
```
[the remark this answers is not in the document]
dakshina99: This is a build-time flag, off by default. config.skipJWTValidation is empty unless a build explicitly stamps it via -ldflags -X, so every normal/default build does strict RS256 validation and rejects none.

Also the file mode is not affected. File mode is handled by its own branch in server.go (AuthModeFile) that always loads the public key and passes SkipValidation: false. It never calls buildAuthenticator, so the flag cannot take effect there. The != AuthModeIDP guard inside buildAuthenticator is only reached for internal_token mode (file is peeled off by the caller; idp fails the guard).

platform-api/config/config.go
dakshina99: Addressed

platform-api/config/config-template.toml
```diff
@@ -263,6 +263,11 @@ private_key_file = "/etc/platform-api/keys/jwt_private.pem"
 # syntax). Not used for "internal_token" tokens — their expiry is whatever
 # "exp" claim the issuer set.
 token_ttl = "1h"
+# NOTE: JWT signature validation cannot be disabled from configuration. The
+# signature-validation bypass used by trusted-mediation-layer deployments is a
+# BUILD-TIME flag baked into the binary via ldflags
+# (-X github.com/wso2/api-platform/platform-api/config.skipJWTValidation=true),
+# never a config-file field — so no internet-facing deployment can turn it on.
```
renuka-fernando: Do we need this comment in config tempate file?

platform-api/config/config.go
```diff
@@ -330,6 +330,25 @@ type JWT struct {
...
+func SkipJWTValidation() bool {
+	return strings.EqualFold(strings.TrimSpace(skipJWTValidation), "true")
```
renuka-fernando: Maybe we can store this in a separate bool variable, so no need to check this `strings.EqualFold` on each time this func is called. I mean we can run this `strings.EqualFold` once, and set to that variable

platform-api/config/config-template.toml
dakshina99: Removed

platform-api/config/config.go
dakshina99: Yes. We can. Improved.

renuka-fernando (approved):

dakshina99:
Closing this since this is done by a config via[1]

[1] https://github.com/wso2/api-platform/pull/3232

Issue #3163, opened 2026-08-06
[Bug]: API Key name needs 3 or more chars in Gateway - but not validated in CP


malinthaprasan:
### Please select the area the issue is related to

AI Workspace, Gateway

### Please select the aspect the issue is related to

Aspect/UI (Frontend layouts, components, styling), Aspect/API (API backends, definitions, contracts, interfaces, OpenAPI)

### Description

<img width="1143" height="743" alt="Image" src="https://github.com/user-attachments/assets/12deb819-4354-4c34-9c3c-3e65e57a8aed" />

```
gateway-controller-1  | time=2026-08-06T04:55:29.467Z level=INFO source=/build/pkg/policyxds/server.go:228 msg="Policy xDS stream request" stream_id=1 type_url=api-platform.wso2.org/v1.RouteConfig version=3 resource_names=[]
gateway-controller-1  | time=2026-08-06T04:55:29.467Z level=INFO source=/build/pkg/policyxds/snapshot.go:446 msg="[linear cache] open watch 3 (delta: false) for api-platform.wso2.org/v1.RouteConfig all resources"
gateway-controller-1  | time=2026-08-06T04:55:29.467Z level=INFO source=/build/pkg/policyxds/server.go:228 msg="Policy xDS stream request" stream_id=1 type_url=api-platform.wso2.org/v1.PolicyChainConfig version=3 resource_names=[]
gateway-controller-1  | time=2026-08-06T04:55:29.467Z level=INFO source=/build/pkg/policyxds/snapshot.go:446 msg="[linear cache] open watch 3 (delta: false) for api-platform.wso2.org/v1.PolicyChainConfig all resources"
gateway-controller-1  | time=2026-08-06T04:56:16.244Z level=INFO source=/build/pkg/controlplane/client.go:1369 msg="Received WebSocket event" type=apikey.created payload="[REDACTED - contains sensitive API key data]"
gateway-controller-1  | time=2026-08-06T04:56:16.244Z level=INFO source=/build/pkg/controlplane/client.go:2941 msg="API Key Created Event received" correlation_id=d609dbbc-01bf-444d-935c-f80d49e7ebed timestamp=2026-08-06T04:56:16Z
[...]

PR #3215, opened 2026-08-13
Fix API key name validation mismatch and align max length to 128 characters

Isuranga-2001:
## Purpose
This PR resolves: 
- #3163

The root cause was a validation mismatch across the API key creation flow: the gateway enforced a set of API key naming rules, but the platform API could still accept invalid caller-supplied names. As a result, names that were valid in one layer could be rejected or silently dropped downstream by the gateway, creating inconsistent behavior and avoidable API failures.

This change aligns the validation logic across the relevant components so that API key names are validated consistently at the source and the accepted range is enforced uniformly.

## Goals

- Align API key name validation rules across the gateway, platform API, and UI
- Enforce a consistent valid range for API key names:
  - minimum length: 1 character
  - maximum length: 128 characters
- Prevent invalid API key names from being accepted by the platform API and then rejected later by the gateway
- Preserve existing generation behavior for display-name-derived names while validating direct caller-supplied IDs
- Add regression coverage to prevent future drift between validation rules

## Approach

- updates the API key name validation logic to keep the lower bound at 1 and the upper bound at 128 characters
- keeps the naming pattern consistent with the gateway rule: lowercase alphanumeric names with hyphen-based separators, without invalid leading/trailing or repeated separators
- validates caller-supplied IDs before persisting or generating downstream values
- adds regression tests to cover valid and invalid inputs, including edge cases around max-length boundaries and invalid characters

## User stories
N/A

## Documentation
N/A

## Automation tests
 - Unit tests 
   - Added validation regression coverage for API key name constraints and invalid caller-supplied IDs
   - Verified valid and invalid edge cases around the 1-character minimum and 128-character maximum
 - Integration tests
   - No dedicated integration test was added for this narrow validation fix because the change is covered at the service validation layer where the mismatch originally occurred.

## Security checks
 - Ran FindSecurityBugs plugin and verified report? no

## Samples
N/A

## Related PRs
N/A

## Test environment

- Validation-tested in the local development environment for the affected Go service and UI logic
- No special hardware or external environment requirements beyond the standard repository development setup


platform-api/internal/database/schema.sql
```diff
@@ -395,7 +395,7 @@ CREATE TABLE IF NOT EXISTS mcp_proxies (
 CREATE TABLE IF NOT EXISTS api_keys (
     uuid VARCHAR(40) PRIMARY KEY,
     artifact_uuid VARCHAR(40) NOT NULL,
-    handle VARCHAR(40) NOT NULL,
+    handle VARCHAR(128) NOT NULL,
```
thivindu, line 398: This requires database schema migrations. Shall we keep this on hold until then?

thivindu (changes requested):
Introduces DB schema changes to the platform API db. Need to discuss

Thushani-Jayasekera:
In the Platform API OpenAPI, the id (handle) is already defined with a 3–40 character limit, and the database handle column is also VARCHAR(40). API keys created from the CP to the DP therefore should follow the 3–40 character constraint.

```
   id:
      type: string
      description: URL-safe handle (identifier) of the API key (generated from displayName when not supplied)
      minLength: 3
      maxLength: 40
   displayName:
      type: string
      description: Human-readable display name of the API key
      minLength: 1
      maxLength: 128
```

```
CREATE TABLE IF NOT EXISTS api_keys (
    uuid VARCHAR(40) PRIMARY KEY,
    artifact_uuid VARCHAR(40) NOT NULL,
    handle VARCHAR(40) NOT NULL,
```

The Gateway OpenAPI currently allows up to 63 characters, and minimum 3.
```
        name:
          type: string
          description: Identifier of the API key. If not provided, a default identifier will be generated
          pattern: "^[a-z0-9]+(-[a-z0-9]+)*$"
          minLength: 3
          maxLength: 63
          example: my-production-key
```

There is currently an inconsistency in the maximum number of characters allowed. However, can we implement the `validateAPIKeyName` function on the Platform API side, along with corresponding UI validation, to ensure that API key handles always comply with the **3–40 character** constraint and are accepted by both the Platform API and Gateway?

malinthaprasan:
> In the Platform API OpenAPI, the id (handle) is already defined with a 3–40 character limit, and the database handle column is also VARCHAR(40). API keys created from the CP to the DP therefore should follow the 3–40 character constraint.
> 
> ```
>    id:
>       type: string
>       description: URL-safe handle (identifier) of the API key (generated from displayName when not supplied)
>       minLength: 3
>       maxLength: 40
>    displayName:
>       type: string
>       description: Human-readable display name of the API key
>       minLength: 1
>       maxLength: 128
> ```
> 
> ```
> CREATE TABLE IF NOT EXISTS api_keys (
>     uuid VARCHAR(40) PRIMARY KEY,
>     artifact_uuid VARCHAR(40) NOT NULL,
>     handle VARCHAR(40) NOT NULL,
> ```
> 
> The Gateway OpenAPI currently allows up to 63 characters, and minimum 3.
> 
> ```
>         name:
>           type: string
>           description: Identifier of the API key. If not provided, a default identifier will be generated
>           pattern: "^[a-z0-9]+(-[a-z0-9]+)*$"
>           minLength: 3
>           maxLength: 63
>           example: my-production-key
> ```
> 
> There is currently an inconsistency in the maximum number of characters allowed. However, can we implement the `validateAPIKeyName` function on the Platform API side, along with corresponding UI validation, to ensure that API key handles always comply with the **3–40 character** constraint and are accepted by both the Platform API and Gateway?

Yeah. I am not sure why we need to change this to 1-128 range. All we need is proper validations to the existing range.

Isuranga-2001:
Thanks @thivindu, @Thushani-Jayasekera and @malinthaprasan for pointing this out.

For context on why this PR went with 1–128 instead of the existing 3–40 range: the original issue (#3163) was that the Gateway validates and enforces its own API key naming rules, but the Platform API had no code-level validation on caller-supplied IDs at all — so a name could pass through the Platform API and only fail later at the Gateway, causing inconsistent behavior. While fixing that gap, the Dev Portal's displayName field already allows up to 128 characters, so the max length was aligned to 128 to avoid rejecting names the Dev Portal itself allows a user to enter.

Looking at this more closely now, here's the full picture across components:

- Dev Portal: displayName allows up to 128 characters.
- Platform API: OpenAPI id/handle is constrained to 3–40 characters, and api_keys.handle is VARCHAR(40) across all four DB schemas (postgres, sqlite, sqlserver, and the base schema.sql) — but again, no service-layer validation currently enforces this.
- Gateway: OpenAPI name allows 3–63 characters, enforced both in validation logic and at the DB level.

So raising the Platform API bound to 128 (as done here) closes the original validation gap, but it does so by moving the max length further out of sync with the Gateway's 63-character limit, and it requires a schema migration on api_keys.handle to actually support 128-character values at the DB level — which this PR doesn't include.

Flagging this here so the inconsistency is captured before deciding how to proceed on this PR.

Isuranga-2001:
Closing this PR.

The original problem (#3163) was that the Gateway enforces API key name validation, but the Platform API had no code-level validation on caller-supplied IDs — so names accepted by the Platform API could still be rejected downstream by the Gateway. This PR aligned the max length to 128 to match the Dev Portal's displayName field while fixing that gap, but that surfaced a broader inconsistency in the max-length constraint across components:

- Dev Portal: 128 characters
- Platform API: 40 characters (api_keys.handle is VARCHAR(40) across all four DB schemas, with no service-layer validation enforcing it)
- Gateway: 63 characters (enforced at both the logic and DB level)

Raising the Platform API bound to 128 requires a DB schema migration on api_keys.handle across all four schemas (Postgres, SQLite, SQL Server, and the default schema), which isn't something we can do as part of this fix right now. Since the length mismatch itself can't be resolved in this PR, I'm closing it rather than merging a partial fix.

I've raised #3235 to track this underlying inconsistency in max API key name length across Dev Portal (128), Platform API (40), and Gateway (63) — resolving it properly needs a DB migration and a decision on a single target length, and it will remain unresolved even after the next step below.

Next steps:

- Opening a new PR scoped to adding validateAPIKeyName validation on the Platform API and ai-workspace, enforcing the current 3–40 character range already defined in the OpenAPI spec and DB schema. This resolves #3163 — Platform API and Gateway will reject invalid names consistently — without requiring a schema migration.

PR #3241, opened 2026-08-17
Tag component logs at the producer so JSON log output stays parseable

### LINKED ISSUE #3206: [Bug]: [pol] log prefix makes JSON traffic-log lines unparseable

Gateway

Aspect/Logging (Log formats, instrumentation, improvements)

### Description

Traffic-log lines reach the container log with a `[pol] ` prefix in front of the JSON, so the line is not valid JSON and cannot be parsed by any log processor (Splunk, Fluent Bit, Loki, Datadog):

```
[pol] {"timestamp":"2026-08-11T11:06:22.120Z","correlationId":"335eb001-3544-4ac5-ab91-fd2acd479450","status":200,"api":{"name":"Echo No Auth","version":"v1","context":"/echo","kind":"RestApi"},"operation":{"method":"GET","path":"/echo/anything"},"target":{"statusCode":200,"destination":"sample-backend/anything"},"client":{"ip":"192.168.128.1","userAgent":"curl/8.4.0"},"latencies":{"durationUs":13268}}
```

`jq` fails at column 5; removing the five leading characters makes the same line parse:

```console
$ jq . tl.txt
jq: parse error: Invalid numeric literal at line 1, column 5

$ sed 's/^\[pol\] //' tl.txt | jq -c '{status, api: .api.name}'
{"status":200,"api":"Echo No Auth"}
```

Since the whole purpose of the JSON output mode is machine consumption, the prefix defeats the feature — operators have to insert a preprocessing step to strip it before anything can parse it.

**Root cause.** The policy engine writes the record as bare, valid JSON to its own stdout — `gateway-runtime/policy-engine/internal/analytics/publishers/log.go`:

```go
func (l *Log) write(data []byte) {
	l.mu.Lock()
	defer l.mu.Unlock()
	if _, err := fmt.Fprintln(l.out, string(data)); err != nil { ... }
}
```

Nothing is wrong at that point. The corruption happens afterwards, outside the process: `gateway-runtime` runs three processes in one container (Envoy, policy engine, Python executor), all writing to the same container stdout, so the entrypoint wraps each one's output in a shell loop that prepends a component tag — `gateway/gateway-runtime/docker-entrypoint.sh`:

```bash
/app/policy-engine -xds-server "${PE_XDS_SERVER}" "${PE_ARGS[@]}" \
    > >(while IFS= read -r line; do echo "[pol] $line"; done) \
    2> >(while IFS= read -r line; do echo "[pol] $line" >&2; done) &
```

That loop is a blind line filter — it cannot distinguish a human-readable log message from a machine-destined JSON record, so it stamps both identically. `docker-entrypoint-debug.sh` does the same.

**This is broader than the traffic log.** Envoy is wrapped by the same mechanism (`docker-entrypoint.sh`, `[rtr]` tag) and its file access log writes to `/dev/stdout`, i.e. the same pipe. So `router.access_logs.format = "json"` is corrupted identically — confirmed locally:

```
[rtr] {"bytesRx":0,"bytesTx":449,"dur":13,"host":"sample-backend","meth":"GET","path":"/echo/anything","respCd":200,"t":"2026-08-11T11:06:22.120Z",...}
```

The policy engine's application logs are also affected when `policy_engine.logging.format = "json"`. A fix that only special-cases the policy engine's traffic log will leave the router's JSON access log to resurface as a separate report.

The prefix itself is load-bearing — the `[rtr]`/`[pol]`/`[pye]` tags are how the three interleaved streams are told apart when debugging — so it cannot simply be removed without a replacement.

### Steps to Reproduce

1. Enable traffic logging in `gateway/configs/config.toml`:

   ```toml
   [traffic_logging]
   enabled = true

   [router.access_logs]
   enabled = true
   ```

2. Start the gateway and deploy any API.
3. Invoke the API once.
4. Inspect the runtime log:

   ```bash
   docker compose logs --tail 20 gateway-runtime | grep correlationId
   ```

   The traffic-log line is prefixed with `[pol] ` and does not parse as JSON.

To see the same defect on the router's access log, additionally set `format = "json"` under `[router.access_logs]`, restart, and invoke again — the `[rtr] {…}` line is equally unparseable.

Severity/Major (Important functionality is broken. Should be prioritized. Doesn't need immediate attention)

API Platform 1.2.0 (reproduced on the 1.2.0 gateway images built from source)

Tharsanan1:
Fixes #3206.

## Problem

`gateway-runtime` runs Envoy, the policy engine and the Python executor in one container, all writing to the same container stdout. To keep the three interleaved streams readable, the entrypoint wrapped each process's stdout **and** stderr in a shell loop that prepended a component tag:

```bash
> >(while IFS= read -r line; do echo "[pol] $line"; done)
```

That loop is a blind line filter — it runs outside the processes, after every line has been written, so it cannot tell a human-readable log message from a machine-destined JSON record and stamps both. The policy engine's traffic log is always JSON, so every line arrived as:

```
[pol] {"timestamp":"2026-08-17T14:09:07.293Z","correlationId":"…","status":200,…}
```

`jq` rejects that at column 5, and so do Splunk, Fluent Bit and Loki. The whole point of the JSON output mode is machine consumption, and the prefix defeats it.

Two more outputs are affected by the same mechanism: Envoy's access log whenever `router.access_logs.format = "json"`, and the policy engine's own application logs when `policy_engine.logging.format = "json"`.

## Approach

Tag at each producer, where prose and machine records are distinguishable, and stop wrapping stdout.

| Output | How it is tagged now |
|---|---|
| Envoy access log (text) | `[rtr] ` in the default `TextFormat` |
| Envoy access log (JSON) | `"component": "rtr"` in the default `JSONFields` |
| Envoy application log | entrypoint's stderr wrapper (Envoy logs to stderr) |
| PE application log (text) | `componentPrefixWriter` on the `slog` handler |
| PE application log (JSON) | `component` attribute on the logger |
| PE traffic log (always JSON) | `component` field on the record |
| Python executor | `[pye] ` in the existing `logging.Formatter` |
| PE bootstrap `init()` | prefixed writer in `plugin_registry.go.tmpl` |

Two points worth reviewer attention:

- **The policy engine's tag is applied to the logger, not the stream.** The traffic-log publisher keeps writing to `os.Stdout` directly, so tagged prose and untouched JSON coexist on one descriptor. That separation is impossible from the shell, which sees one undifferentiated stream.
- **JSON never gets a text prefix** — it gets a field. A prefix would just relocate the bug.

`json_fields` merges with the defaults, so `component` survives unless a deployer overrides that key explicitly. A deployer who overrides `text_format`, however, is responsible for keeping the `[rtr] ` tag — noted in a code comment.

## stderr stays wrapped, deliberately

stdout is unwrapped; **stderr is still tagged, unconditionally.** Nothing writes JSON there (measured: 0 of 151 lines in JSON mode), and it carries the output that bypasses every logger — Go runtime dumps, panics, Python tracebacks, Envoy fatals. No producer-side mechanism can reach that. Forcing `SIGQUIT` on the policy engine:

```
[pol] SIGQUIT: quit
[pol] goroutine 0 gp=0x2ffc1a0 m=0 mp=0x2ffd340 [idle]:
```

Untagged stderr lines during the crash: **0**. Without the wrapper that is a few hundred lines of anonymous stack trace on a stream shared by three processes, at exactly the moment you need to know which one died.

Note for future changes: because the wrapper owns stderr, do **not** also add `[rtr] ` to an Envoy `--log-format` default, or application logs would read `[rtr] [rtr] …`. Each stream has exactly one tagging owner.

## Verification

Built and run from this branch.

**Text mode**
```
traffic log:          {"component":"pol","timestamp":…}   jq parses: YES
prefixed JSON lines:  0
untagged stdout:      0        untagged stderr: 0
present and tagged:   [pol] 153 · [rtr] 4 stdout + 151 stderr · [pye] 9 · [ent] 27
access log:           [rtr] [2026-08-17T14:09:07.293Z] "GET /echo/anything HTTP/1.1" …
```

**All-JSON mode** (`access_logs.format` and `policy_engine.logging.format` both `json`)
```
stdout JSON lines: 44, unparseable 0
prefixed JSON lines: 0     untagged stdout/stderr: 0/0
{"component":"rtr","meth":"GET","respCd":200}
{"component":"pol","level":"INFO"}
{"component":"pol","status":200,"api":"Echo No Auth"}
```

**No functional regression**: `/echo/anything` 200 · `/_gateway-health/ready` 200 · unmatched path 404 · 52 policies registered · all three processes running.

**Tests** — 6 new, plus existing suites green (`cmd/policy-engine`, `internal/analytics/...`, controller `pkg/config`, `pkg/xds`, `pkg/utils`):

- `componentPrefixWriter` returns `len(p)`; any other count reads as a failed write to `io.Writer` callers
- tag and record reach the descriptor in **one** `Write`, so a concurrent writer cannot interleave between them
- text-mode `slog` output starts with `[pol] `
- JSON-mode `slog` output has no prefix, parses, and carries `"component":"pol"`
- traffic-log line starts with `{` **and** carries `"component":"pol"`

## Known limitation

The logger installed by the generated `init()` is fixed to text format: `policy_engine.logging.format` is not known during `init()`, since config loads in `main()`. In JSON logging mode those startup lines are tagged **text** rather than JSON — a processor can recognise and skip them, but they will not parse. Closing that fully needs the log format resolved before policy registration, which is a larger restructure than this change should carry.


Tharsanan1:
@coderabbit review

.agents/skills/gateway-debug/SKILL.md
```diff
@@ -467,13 +467,20 @@ Most bugs surface in logs or config dumps without needing to step through code.
 | Envoy router (in Docker) — `[rtr]` lines | `cd <REPO_ROOT>/gateway && docker compose logs --no-log-prefix gateway-runtime 2>&1 \| grep '^\[rtr\]'` |
 | Python executor (Option 2B) | `/tmp/python_executor.log` |
 
-> Why the `grep`: the `gateway-runtime` container stamps every log line with
-> one of three prefixes — `[rtr]` (Envoy router), `[pol]` (in-container PE,
-> still receives xDS pushes even in debug mode), unprefixed (the entrypoint).
+> Why the `grep`: the `gateway-runtime` container stamps every human-readable log
+> line with one of these prefixes — `[rtr]` (Envoy router), `[pol]` (in-container PE,
+> still receives xDS pushes even in debug mode), `[pye]` (Python executor, started
+> only when compiled Python policies are present), unprefixed (the entrypoint).
```
O-sura: Here it's mentioned that the entrypoint is unprefixed but seems like in the docker-entrypoint.sh the logging function for entrypoint messages(`log()) still appends the `[ent]` prefix. Is that intentional?

renuka-fernando (dismissed):

gateway/gateway-runtime/python-executor/main.py
```diff
@@ -121,9 +121,9 @@ def setup_logging():
     handler = logging.StreamHandler(sys.stdout)
     handler.setLevel(level)
 
-    # Simple format with [pye] prefix for the entrypoint to identify
+    # [pye] identifies this process on the container's shared stdout.
     formatter = logging.Formatter(
-        fmt='%(asctime)s [%(levelname)s] %(name)s: %(message)s',
+        fmt='[pye] %(asctime)s [%(levelname)s] %(name)s: %(message)s',
```
O-sura: Looks like multiline messages (especially exception tracebacks) don’t get tagged with the `[pye]` prefix properly. Since [`logging.Formatter`](https://github.com/python/cpython/blob/af49df919dafc3767ae956767dce0482f9cd6d4e/Lib/logging/__init__.py#L699) appends the traceback after applying the format string, the prefix only ends up on the first line, while the following lines are untagged. Was this the intended behavior, or should we prefix each physical line of the final formatted message?

O-sura (approved):

renuka-fernando (approved):

PR #3277, opened 2026-08-20
feat(platform-api): include user-agent in access logs

renuka-fernando:
## Purpose

The platform-api access log is produced by gin's built-in logger, whose format is fixed and does not capture the client User-Agent, making it hard to identify which clients are calling the API.
This PR extends the access log to include the request User-Agent.

## Approach

- Replace `gin.Default()` with `gin.New()` plus explicit logger and recovery middleware so the access-log format can be customized.
- Add a custom log formatter that mirrors gin's default access-log format and appends the request User-Agent.

## Related Issues

N/A

## Checklist


renuka-fernando:
@coderabbitai review

ashera96 (approved):

renuka-fernando:
@coderabbitai review

PR #3281, opened 2026-08-21
chore(APICP): revamp the portal UI with a scope-aware app shell

ShavinAnjithaAlpha:
## Purpose

Revamp the API Control Plane portal's app shell so it reads and behaves like Bijira, and restructure the source tree around it. The sidebar, header and layout are the visible half; underneath, the scope model (organization → project → API) is made explicit in the route builders, the navigation registry and a new `ScopeGate`, so what the sidebar offers and what a page can actually render can no longer disagree.

Resolves https://github.com/wso2-enterprise/apim-saas/issues/2897, https://github.com/wso2-enterprise/apim-saas/issues/2907

### Sidebar

- Replaced per-scope sidebar headings with three divider-separated clusters: Place, API, and Global.
- Added nested submenus for Test, Insights, Observability, and Manage, with active-child expansion and explicit navigation in the collapsed rail.
- Made all items available at every scope via scope-less aliases, with ScopeGate resolving missing scope.
- Made Overview scope-adaptive and highlighted across nested scopes.
- Limited capability gating to API scope so API-level entry points remain accessible while an API is unloaded.

### Header

- **Organization / Project / API switchers**, each a new searchable `SearchableComplexSelect` (icon + primary/secondary rows, bounded trigger width, ellipsized option text) in place of the plain `ComplexSelect`.
- **Quick selectors** (`ProjectQuickSelector`, `APIQuickSelector`) appear for the levels not yet selected, with their own loading/error states.
- **Step-back affordances**: an inline clear button on the project and API switchers navigates back to organization / project level.
- Options resolve `displayName` with a fallback to the handle, and keep the current value selectable before the list loads so the switcher never renders out of range.
- All shell strings moved to `react-intl` (brand title, switcher labels, search / empty / no-results copy, notification tooltip, footer links, panel title).

### Layout

- Breadcrumbs moved inside `PageContent` (full width, `Stack`-spaced above the outlet) instead of floating above it, and now label from `displayName`.
- Footer and notification-panel strings i18n'd.

## Scope model, routes and gating

- **`projectPath` / `apiPath` builders with an explicit unresolved-scope marker.** A `null` handle produces the page's scope-less alias by substituting a reserved `select-scope` segment (`/organizations/acme/select-scope/apis`) rather than dropping segments, which would make `/organizations/acme/projects/apis` read positionally as *project `apis`*, and would collide project Settings/Home with their organization-level equivalents.
- **`ScopeGate`** renders a project/API picker in place of the page body until the route carries the scope the page needs, then navigates to the fully-scoped URL.
- **`projectScopedPaths` / `apiScopedPaths` / `apiScopeSelectPaths`** generate both the route patterns `AppRoutes` registers and the `match` predicates the sidebar highlights on, from the same builder, the route table and the highlight can't drift apart. Submenu parents match only the aliases, so parent and child never both claim active.
- **`ConsoleScopeProvider` / `ConsoleScopeContext`** now expose `isProjectScope` / `isApiScope`, projects and their error state, and API capabilities, typed against the generated resource clients (`RestApi`, `Organization`, `Project`) rather than hand-written domain types.
- Runtime logs moved from project-wide `observe/runtimelogs` to API-scoped `observability/logs`.
- New `ComingSoon` placeholder pages for Admin, API Insights, Compliance, Metrics, Alerts, Monetize, LifeCycle, API Console and API Chat, so every sidebar destination resolves to something.

## Restructuring

- `src/features/*` → `src/pages/appShell/appShellPages/*` (apis, deploy, gateways, projects, organizations, observability, insights, manage, test, admin, settings), with each page's own `components/` and `utils/` subfolders.
- `src/features/auth` → `src/contexts/auth`; `features/billing/ProductActivation` → `src/hooks`; layouts → `src/pages/appShell`; `src/pages/auth` for login/callback.
- API resource index modules added under `src/api/resources/*` (organizations, projects, restApis, apiKeys, gateways).
- Project listing gains pagination, sorting and a reworked card grid; project forms moved onto Oxygen form controls.

## Theme

- Shared `sx` recipes (`glassSurfaceSx`, `hairline`, `interactiveCardSx`, `stickyBottomBarSx`) so surfaces stay consistent across pages.

## Documentation 

N/A

## Tests

- New: `AppSidebar.test.tsx`, `ScopeGate.test.tsx`, `paths.test.ts`, `useNavigationItems.test.tsx`.
- Expanded: `navigationRegistry.test.ts` (clusters, submenus, adaptive items, alias matching), `ApiListPage.test.tsx`, `ProjectListPage.test.tsx`, `NewProjectDialog.test.tsx`, `apiCapabilities.test.ts`.

## Related

- Builds on #3259 (settings pages + Slot/Port extension seam), already on `main`.

## Test environment

- Node.js 24.x (per engines in package.json), npm
- Vite 5.0.11, TypeScript 5.8.3, React 19.2.3, react-intl 10.1.20
- Vitest 2.1.8 on jsdom 25.0.1


portals/api-control-plane/src/api/resources/apiKeys/index.ts
```diff
@@ -0,0 +1,51 @@
...
+    UpdateApiKeyResponse,
+    ApiKeyArtifactType
+}
+
+export {
+    useMyApiKeys,
+    useCreateApiKey,
+    useUpdateApiKey,
+    useRevokeApiKey,
+}
```
Induwara04: Add one extra line

portals/api-control-plane/.claude/oxygen-ui/components.md
```diff
@@ -0,0 +1,1431 @@
+# Oxygen UI Component Reference
+
```
Induwara04: Initiate a one skill based on this all like 'how to develop UI'

ShavinAnjithaAlpha: Pushed a fixed in commit [3572c88](https://github.com/wso2/api-platform/pull/3281/commits/3572c881c4024cffcaeffa871bab69a4d17cbc92), review it.

portals/api-control-plane/src/api/resources/apiKeys/index.ts
ShavinAnjithaAlpha: Pushed a fixed in commit [3572c88](https://github.com/wso2/api-platform/pull/3281/commits/3572c881c4024cffcaeffa871bab69a4d17cbc92), review it.

portals/api-control-plane/package.json
```diff
@@ -38,19 +38,24 @@
...
-    "@wso2/oxygen-ui": "0.11.0",
-    "@wso2/oxygen-ui-icons-react": "0.11.0",
+    "@wso2/oxygen-ui": "0.13.1",
+    "@wso2/oxygen-ui-icons-react": "0.13.1",
     "axios": "1.19.0",
     "js-yaml": "4.1.0",
     "react": "19.2.3",
     "react-dom": "19.2.3",
-    "react-intl": "^10.1.20",
+    "react-intl": "10.1.20",
```
AnuGayan, line 55: **Severity:** 🔴 blocker
**Subject:** Versions pinned here without regenerating `package-lock.json` — `npm ci` fails

**Description**
Commit `c8fba03` pinned `react-intl` `^10.1.20`→`10.1.20`, `eslint-plugin-formatjs` `^6.4.0`→`6.4.0`, `babel-plugin-formatjs` `^11.3.16`→`11.3.16` and `@formatjs/cli` `^6.16.18`→`6.16.18`, but the lockfile still records the resolved caret versions (`react-intl@10.1.22`, `eslint-plugin-formatjs@6.5.0`, …). `npm ci` refuses to install. `Dockerfile:27` runs `npm ci --ignore-scripts` and `.github/workflows/apip-api-control-plane-release.yml:53` runs `npm ci`, so the production image and the release job both break on this branch. Confirmed by running `npm ci` at `c8fba03` and by reading both files.

**How to verify**
```
cd portals/api-control-plane && npm ci
# npm error `npm ci` can only install packages when your package.json
# and package-lock.json ... are in sync.
# npm error Invalid: lock file's react-intl@10.1.22 does not satisfy react-intl@10.1.20
```

**Suggested fix**
Regenerate the lockfile against the pins (do **not** unpin):
```bash
cd portals/api-control-plane
npm install --package-lock-only
git add package-lock.json
npm ci   # must now succeed
```

**Prompt for Claude**
```text
In portals/api-control-plane, package.json and package-lock.json are out of sync.
1. First confirm: run `npm ci` in portals/api-control-plane. If it succeeds, stop and tell me — the finding is wrong.
2. If it fails with "can only install packages when your package.json and package-lock.json ... are in sync",
   regenerate the lockfile: `npm install --package-lock-only`. Do NOT change any version in package.json —
   the pins in commit c8fba03 are intentional.
3. Confirm afterwards with `npm ci` (exit 0), then `npm run typecheck` and `npm test` (both must still pass).
4. The only expected diff in package-lock.json is the resolved versions of react-intl, eslint-plugin-formatjs,
   babel-plugin-formatjs, @formatjs/cli and their transitive deps.
```

portals/api-control-plane/src/extensions.tsx
```diff
@@ -16,43 +16,37 @@
...
+ * decides which sidebar section it's grouped under — mirrors
+ * `NavigationDefinition` so it can be merged straight into the existing
+ * nav pipeline in `navigation/useNavigationItems.ts`.
  */
-export type ApiControlPlaneExtension = SlotEntry & {
+export type ApiControlPlaneExtension = {
+  id: string;
   routePath: string;
-  render: (port: CloudHostPort) => ReactNode;
+  element: ReactNode;
```
AnuGayan: **Severity:** 🟠 major
**Subject:** `render(port)` → `element` reverts the #3259 Slot/Port seam and orphans `src/slots/`

**Description**
The PR body says this "builds on #3259 (settings pages + Slot/Port extension seam), already on `main`". It replaces it. On `main`, `ApiControlPlaneExtension = SlotEntry & { render: (port: CloudHostPort) => ReactNode, scope, ... }` dispatched through `SlotEntriesProvider`/`useSlotEntries`. This PR rewrites it to `{ element: ReactNode, level, ... }` on a flat `ExtensionsContext`, and deletes `navigation/useSettingsTabs.tsx` (the `settings.project.tabs` consumer) plus `routes/AppRoutes.settingsTab.test.tsx`. `src/slots/index.tsx` survives with no consumer, and extensions no longer receive `CloudHostPort` (`hostPort` importers drop 3 → 1). That is a breaking change to the contract `apim-saas` injects against, made silently — nothing goes red because the orphaned slot tests still pass in isolation. Confirmed via `git diff 44a453e..c8fba03 -- src/extensions.tsx` and importer greps on both refs.

**How to verify**
```bash
git diff 44a453e..c8fba03 -- portals/api-control-plane/src/extensions.tsx
git grep -l "slots" 44a453e -- portals/api-control-plane/src | grep -v src/slots/
#   -> extensions.tsx, hostPort.tsx, navigation/useSettingsTabs.tsx
grep -rn "from '.*slots'" portals/api-control-plane/src | grep -v "src/slots/"
#   -> no consumer remains on this branch
```

**Suggested fix**
No snippet — this is a product decision, not a mechanical edit. Either keep the `SlotEntry`/`render(port)` contract and adapt the new nav pipeline to it, or, if the revert is deliberate, say so in the PR body, delete `src/slots/` and its test in the same commit, and flag the breaking change to the `apim-saas` host.

**Prompt for Claude**
```text
src/extensions.tsx on this branch replaces the Slot/Port extension contract that PR #3259 added on main.
1. First confirm: `git diff 44a453e..HEAD -- portals/api-control-plane/src/extensions.tsx`. Check whether main's version
   uses SlotEntry, useSlotEntries and `render: (port: CloudHostPort)`, and whether HEAD's uses a plain `element: ReactNode`
   on its own context. If HEAD still uses the slot seam, stop and tell me — the finding is wrong.
2. Then search the apim-saas consumer (or ask me for it) for objects passed to ExtensionsProvider, and report which fields
   would break: `scope`->`level`, `render`->`element`, `slot` removed.
3. Do NOT change extensions.tsx yourself. Report the breakage surface and whether src/slots/index.tsx has any remaining
   importer, so we can decide between restoring the seam and deleting it deliberately.
4. Confirm src/slots/index.test.tsx still passes either way (`npx vitest run src/slots`).
```

portals/api-control-plane/src/pages/appShell/appShellPages/apis/ApiDetailPage.tsx
```diff
@@ -0,0 +1,195 @@
...
+              <Stack
+                direction="column"
+                justifyContent="space-between"
+                alignItems="flex-end"
+                sx={{ alignSelf: 'stretch' }}
+              >
+                {/* Deployments remain viewable for gateway-created proxies (deploy/
+                    redeploy/restore/undeploy are disabled on the page itself), so the
+                    button navigates but is relabelled "View Deployments". */}
+                <Button
```
AnuGayan: **Severity:** 🟠 major
**Subject:** "Deploy to Gateway" is a primary button with its navigation commented out — it does nothing

**Description**
This `<Button variant="contained">` has `component={RouterLink}` and `to={...}` both commented out and no `onClick`. It is the most prominent control on the API overview page and it is inert — clicking it does nothing, with no disabled state to signal that. The whole file is a copy-paste from the AI Workspace portal: its message ids are `aiWorkspace.pages.appShell.appShellPages.proxies.LLMProxyOverview.*`, and the surrounding comments reference `isReadOnlyProxy`, `canUpdateProxy`, `proxiesPath` and "gateway-created proxies", none of which exist here. Confirmed by reading the file; it replaces a deleted `ApiDetailPage.tsx` that had a test.

**How to verify**
Open this file at line 159. Run the app, open any API's overview, click **Deploy to Gateway** — no navigation, no console error.

**Suggested fix**
Wire it to the deploy route the sidebar already owns:
```tsx
<Button
  variant="contained"
  component={RouterLink}
  to={routes.apiDeploy(params.orgHandle!, params.projectHandler!, params.apiHandler!)}
>
  <FormattedMessage
    id="apiControlPlane.apiDetail.deployToGateway"
    defaultMessage="Deploy to Gateway"
  />
</Button>
```
Requires importing `Link as RouterLink` from `react-router-dom`, `routes` from `../../../../routes/paths`, and reading `params` from `useConsoleScope()`.

**Prompt for Claude**
```text
In src/pages/appShell/appShellPages/apis/ApiDetailPage.tsx around line 159, the "Deploy to Gateway" Button has
`component={RouterLink}` and `to=` commented out and no onClick, so it is inert.
1. First confirm: read lines 150-185 and check the Button has neither onClick nor an active component/to. If it already
   navigates, stop and tell me — the finding is wrong.
2. If inert, wire it to routes.apiDeploy(orgHandle, projectHandler, apiHandler), reading handles from
   useConsoleScope().params. Import `Link as RouterLink` from react-router-dom and `routes` from ../../../../routes/paths.
3. While here, replace the hardcoded "Deploy to Gateway" text with a FormattedMessage under an `apiControlPlane.*` id
   (NOT the `aiWorkspace.*` prefix this file uses), and delete the commented-out Tooltip/IconButton/DisabledActionTooltip
   blocks at lines 102-121 and 167-181 — they reference identifiers that do not exist in this portal.
4. Do not change the OverviewTab render at the bottom of the file.
5. Confirm with `npm run typecheck` and `npm run lint` (lint currently fails at line 139 of this file — fixing that is expected).
```

portals/api-control-plane/src/pages/appShell/AppHeader.tsx
```diff
@@ -0,0 +1,358 @@
...
+              >
+                <X size={14} />
+              </IconButton>
+            </Box>
+          )}
+
+          {!params.projectHandler && (
+            <ProjectQuickSelector
+              disabled={!params.orgHandle}
+              isProjectsLoading={isLoading || projectsError !== undefined}
```
AnuGayan: **Severity:** 🟠 major
**Subject:** Folding the error into the loading flag makes `ProjectQuickSelector`'s error branch unreachable

**Description**
`isProjectsLoading={isLoading || projectsError !== undefined}` feeds a component whose render order is `isProjectsLoading ? <Loading/> : projectsError ? <Failed to load projects/> : …` (`ProjectQuickSelector.tsx:150-160`). Because the error is OR-ed into the loading flag, the first branch always wins when `projectsError` is set, and the "Failed to load projects" message is dead code. A user whose project list fails to load sees a spinner that never resolves, with no signal that the request failed. Confirmed by reading both files; the sibling `APIQuickSelector` is passed `isApisLoading={apisQuery.isLoading}` and does not have the bug.

**How to verify**
Open `AppHeader.tsx:240` beside `ProjectQuickSelector.tsx:150`. To reproduce: block the projects request in devtools, then open the project quick-selector at org scope — it shows "Loading…" indefinitely.

**Suggested fix**
```suggestion
              isProjectsLoading={isLoading && projectsError === undefined}
```

**Prompt for Claude**
```text
In src/pages/appShell/AppHeader.tsx line 240, `isProjectsLoading={isLoading || projectsError !== undefined}` makes
ProjectQuickSelector's error branch unreachable.
1. First confirm: read ProjectQuickSelector.tsx lines 148-165 and check the render order is
   `isProjectsLoading ? Loading : projectsError ? Failed : ...`. If projectsError is checked first, stop and tell me —
   the finding is wrong.
2. If loading is checked first, change AppHeader.tsx:240 to `isProjectsLoading={isLoading && projectsError === undefined}`
   so the error branch wins. Keep passing `projectsError={projectsError}` unchanged.
3. Do not touch the APIQuickSelector props below it — that one is already correct.
4. Confirm with `npm run typecheck` and `npm test`.
```

AnuGayan (commented):
## Review — scope-aware app shell

A large app-shell rewrite: three-cluster sidebar with submenus, searchable org/project/API header switchers, an explicit scope model (`projectPath`/`apiPath` + `ScopeGate` + the `select-scope` marker), and a `src/features/*` → `src/pages/appShell/appShellPages/*` restructure.

**The scope model is the strongest part.** Generating route patterns *and* sidebar `match` predicates from one builder (`projectScopedPaths`/`apiScopedPaths`/`apiScopeSelectPaths`) closes a whole class of drift, and the `SELECT_SCOPE_SEGMENT` rationale in `paths.ts` — spelling out why dropping segments would make `/projects/apis` read positionally as *project `apis`* — is exactly the reasoning that usually gets lost. `toRouteRegex` escapes metacharacters *before* substituting `:param`, in the right order. `ConfirmDialog`'s rework to a real `<form>` with `useId()`-bound labels is a genuine a11y improvement, and `ProjectListPage` (debounced server-side search, page reset on filter change, `page > lastPage` clamping after a delete, `defineMessages` for every string) reads like the standard the other new pages should be held to.

Against that, the branch does not build reproducibly, ships a placeholder page copy-pasted from the AI Workspace portal with an inert primary button, reverts the extension seam #3259 just landed, and deletes ~875 lines of tests for pages it rewrites.

### Checks I ran on `c8fba03`

| Command | Result |
|---|---|
| `npm ci` | ❌ **fails** — lockfile out of sync with the pinned versions |
| `npm run typecheck` | ✅ clean |
| `npm test` | ✅ 59 files, 732 tests pass (17 MUI out-of-range warnings) |
| `npm run lint` | ❌ 2 errors, both in files added by this PR |
| `npm run i18n:extract` | ⚠️ 2 duplicate ids; `en.json` 21 lines stale |

### Findings

13 inline comments below, each with a reproduction and a paste-ready prompt. Headlines:

- 🔴 **`npm ci` fails** — `c8fba03` pinned versions without regenerating `package-lock.json`. `Dockerfile:27` and `apip-api-control-plane-release.yml:53` both run `npm ci`, so the image and the release job break.
- 🟠 **The #3259 Slot/Port seam is reverted, not built on** — `render(port)`→`element`, `scope`→`level`, `useSettingsTabs.tsx` and its test deleted, `src/slots/` orphaned.
- 🟠 **"Deploy to Gateway" does nothing** — `ApiDetailPage.tsx:159` has its link commented out.
- 🟠 **A failed project load shows "Loading…" forever** — `AppHeader.tsx:240` folds the error into the loading flag.

### One finding I could not anchor inline (deleted files)

**Severity:** 🟠 major
**Subject:** Six test files deleted with no replacement, and the PR body describes one of them as "expanded"

**Description**
Deleted with no equivalent anywhere on the branch: `ApiCreatePage.test.tsx` (100), `ApiDetailPage.test.tsx` (101), `ApiListPage.test.tsx` (118), `OverviewTab.test.tsx` (218), `DeployPage.test.tsx` (232), `AppRoutes.settingsTab.test.tsx` (106). The pages themselves survive in rewritten form — `ApiListPage.tsx` is +86/−112, `DeployPage.tsx` +37/−11, `ApiDetailPage.tsx` is brand new — so the rewrite lands untested. The PR body's Tests section lists `ApiListPage.test.tsx` under "Expanded", but no file by that name exists on the branch.

**How to verify**
```bash
git ls-tree -r --name-only 44a453e -- portals/api-control-plane/src | grep -E '\.test\.tsx?$' \
  | sed 's|.*/||' | sort -u > /tmp/base.txt
git ls-tree -r --name-only c8fba03 -- portals/api-control-plane/src | grep -E '\.test\.tsx?$' \
  | sed 's|.*/||' | sort -u > /tmp/pr.txt
comm -23 /tmp/base.txt /tmp/pr.txt
find portals/api-control-plane/src -name 'ApiListPage.test.tsx'   # no output
```

**Suggested fix**
Port the deleted suites to the new file locations. At minimum restore `ApiListPage.test.tsx` and `DeployPage.test.tsx`, whose pages changed substantially, and correct the PR body's Tests section.

**Prompt for Claude**
```text
This branch deletes six test files without replacing them. Restore the two highest-value ones.
1. First confirm they are gone: `find portals/api-control-plane/src -name 'ApiListPage.test.tsx'
   -o -name 'DeployPage.test.tsx'`. If either exists, stop and tell me which — the finding is
   partly wrong.
2. Recover the originals:
   `git show 44a453e:portals/api-control-plane/src/features/apis/ApiListPage.test.tsx`
   `git show 44a453e:portals/api-control-plane/src/features/deploy/DeployPage.test.tsx`
3. Port each to its page's new location (src/pages/appShell/appShellPages/apis/ and .../deploy/),
   updating import paths and any assertions the rewrite invalidated. Where the page's behaviour
   genuinely changed, update the assertion rather than deleting the test.
4. Do not weaken assertions just to make them pass, and do not add `.skip`.
5. Confirm with `npm test` — all suites must pass and the file count must rise from 59.
```

### Nits, not worth their own thread

- **`aiWorkspace.*` message ids leak in**: `ApiDetailPage.tsx:127,138` (one names `LLMProxyOverview`), `ProjectQuickSelector.tsx:126,153,159,167`, `APIQuickSelector.tsx`, `ExploreMoreCard.tsx:99`. Every other id uses `appShell.*` / `apiControlPlane.*` — worth renaming before these reach a translation memory.
- **`ExploreMoreCard.tsx:80-90` links to AI Workspace docs** — "Manage Existing App LLM Proxies" pointing at `.../ai-workspace/llm-proxies/manage-proxy/`. Wrong product for this page.
- **Hardcoded English despite the "all shell strings moved to react-intl" claim**: every `ScopeGate` `prompt` (12 call sites — `prompt` is typed `string`, so `<FormattedMessage>` can't be passed; widening it to `ReactNode` would fix that), `ScopeGate`'s `DEFAULT_PROMPT` / `"Unable to load projects"` / `"Loading projects"`, all of `restApiDisplay.ts` (`'REST API'`, `'Published'`, `'Not deployed'`, `'Unknown'` — these render straight into chips), `SettingsPage.tsx` in full, `AppLayout.tsx:164` `"You're all caught up."`, `SaveBar.tsx` `'Save changes'` / `'Saving…'`. `ProjectQuickSelector.tsx:134` hardcodes `placeholder="Search"` where the API one uses `intl.formatMessage`.
- **`src/theme/receipes.ts` is misspelled** (`recipes`), and `glassSurfaceSx` is exported from `theme/index.ts` with no consumer.
- **`src/theme/emotionCache.ts` carries authoring leftovers**: a redundant `// src/theme/emotionCache.ts` path header, and `// ...read the per-request nonce from a meta tag (see §10)` — a cross-reference to a document that isn't in the repo. The code also reads `.nonce` off a `meta[property="csp-nonce"]` element; if the nonce is carried in `content=` (the usual convention) that yields `''`, not the nonce. *Unverified — I found no `csp-nonce` meta in `index.html`, so this path may never fire today.*
- **`AppHeader.tsx:229,308`** — the clear `IconButton` is `width: 20, height: 10`; the 10px height looks like a typo and gives a sub-minimum tap target.
- **`ApiDetailPage.tsx:38-48`** — `truncateProviderDisplayName` is declared inside the component body, after the early returns, at the wrong indent level. It's a pure helper that belongs at module scope, and it truncates an API display name, not a provider's.
- **`useNavigationItems.ts:139-146`** — the sort comparator calls `combinedRegistry.find()` twice per comparison. Fine at ~20 items, but a precomputed `Map<id, order>` would be faster and clearer.

---
*Reviewed at `c8fba03` against base `44a453e`, with a local install (`npm install`), typecheck, full test run, lint and i18n extract. Findings marked "Confirmed" were reproduced; anything I could not reproduce is labelled unverified inline.*

portals/api-control-plane/src/scope/ScopeGate.tsx
```diff
@@ -0,0 +1,364 @@
...
+                        <MenuItem value="__loading__" disabled>
+                          <FormattedMessage
+                            id="scopeGate.loadingProjects"
+                            defaultMessage="Loading projects..."
+                          />
+                        </MenuItem>
+                      ) : projects.length === 0 ? (
+                        <MenuItem value="" disabled>
+                          <FormattedMessage
+                            id="scopeGate.noProjects"
```
AnuGayan: **Severity:** 🟡 minor
**Subject:** `scopeGate.noProjects` is reused here for different copy, so this menu item renders the wrong string

**Description**
`scopeGate.noProjects` is used at line 168 for the empty-state paragraph ("You have no projects yet. Create a project to continue.") and again here for a dropdown item ("No projects available"). `scopeGate.noApis` has the same problem at lines 293 and 355. react-intl resolves by id, so `en.json` wins for both call sites. Confirmed against `src/i18n/messages/en.json:356,359`: `noApis` holds the long sentence and `noProjects` holds "No projects available" — so the dropdown shows a full sentence and the empty-state paragraph shows a three-word label. `formatjs extract` reports both as `Duplicate message id`.

**How to verify**
```
cd portals/api-control-plane && npm run i18n:extract
# Duplicate message id: "scopeGate.noProjects"
# Duplicate message id: "scopeGate.noApis"
```
Then compare `en.json:356-361` with the four `defaultMessage`s here.

**Suggested fix**
```suggestion
                            id="scopeGate.projectSelect.empty"
                            defaultMessage="No projects available"
```
Same treatment for the API menu item at line 293 (`scopeGate.apiSelect.empty`), then re-run `npm run i18n`.

**Prompt for Claude**
```text
In src/scope/ScopeGate.tsx, the ids "scopeGate.noProjects" (lines 168 and 233) and "scopeGate.noApis" (lines 293 and 355)
are each used for two different sentences.
1. First confirm: run `npm run i18n:extract` in portals/api-control-plane and check it prints `Duplicate message id` for
   both. If it does not, stop and tell me — the finding is wrong. Restore en.json afterwards with
   `git checkout -- src/i18n/messages/en.json`.
2. Rename only the two dropdown MenuItem occurrences (line 233 -> scopeGate.projectSelect.empty,
   line 293 -> scopeGate.apiSelect.empty), keeping their short "No projects available" / "No APIs available"
   defaultMessages. Leave the two paragraph occurrences on the existing ids.
3. Then run `npm run i18n` to regenerate en.json and commit the result.
4. Confirm: `npm run i18n:extract` prints no Duplicate warnings, and `git diff --exit-code src/i18n/messages` is clean
   after `npm run i18n`.
```

portals/api-control-plane/src/i18n/messages/en.json
```diff
@@ -1 +1,380 @@
...
+  "scopeGate.goToProjects": {
+    "defaultMessage": "Go to Projects"
+  },
+  "scopeGate.loadingApis": {
+    "defaultMessage": "Loading APIs..."
+  },
+  "scopeGate.loadingProjects": {
+    "defaultMessage": "Loading projects..."
+  },
+  "scopeGate.noApis": {
```
AnuGayan, line 890: **Severity:** 🟡 minor
**Subject:** `en.json` is stale — six ids added by this PR were never extracted

**Description**
`DocumentsPage`, `PoliciesPage` and `RoutingPage` (all added here) define six ids under `apiControlPlane.pages.appShell.appShellPages.apis.develop.*` that are absent from this file. English still renders via `defaultMessage`, so nothing breaks visibly — but the strings never reach translators, and the repo's own `i18n:check` script fails. Confirmed by running `npm run i18n:extract` on `c8fba03`, which adds 21 lines to this file. (The same run also reports the duplicate `scopeGate.noApis`/`scopeGate.noProjects` ids — see the ScopeGate comment; fix that first.)

**How to verify**
```
cd portals/api-control-plane && npm run i18n:extract && git diff --stat src/i18n/messages/en.json
# 1 file changed, 21 insertions(+)
```

**Suggested fix**
```bash
cd portals/api-control-plane && npm run i18n && git add src/i18n/messages/en.json
```
Run this **after** fixing the duplicate ids, so the wrong copy isn't baked in.

**Prompt for Claude**
```text
portals/api-control-plane/src/i18n/messages/en.json is missing six ids added by this branch.
1. First confirm: `cd portals/api-control-plane && npm run i18n:extract && git diff --stat src/i18n/messages/en.json`.
   If the diff is empty, stop and tell me — the finding is wrong.
2. Fix the duplicate-id issue in ScopeGate.tsx FIRST (scopeGate.noProjects / scopeGate.noApis), otherwise the regenerated
   file bakes in the wrong copy.
3. Then run `npm run i18n` and commit src/i18n/messages/en.json plus any compiled output the script writes.
4. Confirm with `npm run i18n:check` (exit 0) and `npm test`.
```

portals/api-control-plane/src/scope/ScopeGate.tsx
```diff
@@ -0,0 +1,364 @@
...
+                <FormControl fullWidth sx={{ maxWidth: 300 }}>
+                    <FormLabel id="scope-gate-api-label">
+                      <FormattedMessage
+                        id="scopeGate.api"
+                        defaultMessage="API"
+                      />
+                    </FormLabel>
+                    <Select
+                      labelId="scope-gate-api-label"
+                      value={chosenApi || '__loading__'}
```
AnuGayan: **Severity:** 🟡 minor
**Subject:** `chosenApi || '__loading__'` selects a sentinel whose `MenuItem` only exists while loading

**Description**
The `__loading__` `MenuItem` is rendered only inside the `apisQuery.isPending` branch. Once the query settles with nothing chosen, `value` is still `'__loading__'` but no option carries it, so MUI logs an out-of-range warning and the trigger renders blank instead of a placeholder. Line 207 has the same shape for the project select. Confirmed empirically: `npm test` on this branch emits 17 such warnings, several naming `ScopeGate.test.tsx` and reading ``out-of-range value `__loading__` ... available values are `orders-api` ``.

**How to verify**
```
cd portals/api-control-plane && npm test 2>&1 | grep -c out-of-range   # 17
cd portals/api-control-plane && npm test 2>&1 | grep -B2 out-of-range | grep '^stderr'
```

**Suggested fix**
```suggestion
                      value={apisQuery.isPending ? '__loading__' : chosenApi}
```
Same shape at line 207: `value={isLoading ? '__loading__' : chosenProject}`. `displayEmpty` is already set, so the empty value still renders a placeholder row.

**Prompt for Claude**
```text
Fix out-of-range MUI Select values in src/scope/ScopeGate.tsx.
1. First confirm the warnings are real: `cd portals/api-control-plane && npm test 2>&1 | grep -c out-of-range`.
   Expect 17. If it is 0, stop and tell me — the finding is wrong.
2. At line 207 and line 276 the sentinel '__loading__' is used as the value even when the matching disabled MenuItem is
   not rendered. Change each to select the sentinel only while that query is loading, and '' otherwise:
     line 207 -> value={isLoading ? '__loading__' : chosenProject}
     line 276 -> value={apisQuery.isPending ? '__loading__' : chosenApi}
   Keep `displayEmpty` so the empty value still renders a placeholder row.
3. Do not change the submit/navigate logic.
4. Confirm: `npm test` must still pass AND the out-of-range count must drop. Report any that remain rather than
   suppressing them.
```

portals/api-control-plane/src/pages/appShell/AppHeader.tsx
```diff
@@ -0,0 +1,358 @@
...
+      ? projects.map((project) => ({ handler: project.id, name: project.displayName || project.id })) 
+      : params.projectHandler && project ? [{ handler: params.projectHandler, name: project.displayName || params.projectHandler }] 
+        : [];
+
+  const apisQuery = useRestApis(
+    {},
+    { projectId: project?.id, orgId: organization?.id }
+  );
+  const apis = apisQuery.data?.list ?? [];
+  const apiOptions: { handler: string; name: string }[] =
```
AnuGayan: **Severity:** 🟡 minor
**Subject:** `apiOptions` has no current-value fallback, so the API switcher renders out of range on first paint

**Description**
`apiOptions` is `[]` unless `projects.length > 0 && project` have both resolved, while the switcher below is given `value={params.apiHandler}`. Until the project resolves there is no matching option, so MUI warns and the switcher renders blank. `orgOptions` (lines 93-99) already carries exactly the fallback that prevents this, and the PR body claims the switchers "keep the current value selectable before the list loads so the switcher never renders out of range" — that holds for org, partly for project, and not at all for API. Confirmed by the test run: `AppRoutes.orgSettings.test.tsx` emits ``out-of-range value `retail-apis` ... available values are `` ``.

**How to verify**
```
cd portals/api-control-plane && npm test 2>&1 | grep -B2 out-of-range | grep orgSettings
```
Or hard-reload any API-scoped URL and watch the API switcher flash empty before the list arrives.

**Suggested fix**
Give `apiOptions` the same fallback shape `orgOptions` uses:
```tsx
: params.apiHandler
  ? [{ handler: params.apiHandler, name: component?.displayName || params.apiHandler }]
  : [];
```
`component` is already on `useConsoleScope()`.

**Prompt for Claude**
```text
In src/pages/appShell/AppHeader.tsx around lines 109-119, apiOptions is [] until both `projects.length > 0` and `project`
resolve, while the API switcher's value is params.apiHandler — so it renders out of range on first paint.
1. First confirm: `cd portals/api-control-plane && npm test 2>&1 | grep -c out-of-range` (expect 17), and read lines
   93-119 to confirm orgOptions has a current-value fallback and apiOptions does not. If apiOptions already has one,
   stop and tell me — the finding is wrong.
2. Add the same fallback to apiOptions, sourcing the label from `component?.displayName` (available on useConsoleScope())
   falling back to the handle.
3. Do not change how changeApi/changeProject navigate.
4. Confirm: `npm test` must still pass and the out-of-range count must drop.
```

portals/api-control-plane/src/pages/appShell/AppHeader.tsx
```diff
@@ -0,0 +1,358 @@
...
+                )}
+                searchPlaceholder={intl.formatMessage({ id: 'appShell.header.api.placeholder', defaultMessage: 'Search APIs...' })}
+                emptyMessage={intl.formatMessage({ id: 'appShell.header.api.empty', defaultMessage: 'No APIs found' })}
+                noResultsMessage={intl.formatMessage({ id: 'appShell.header.api.noResults', defaultMessage: 'No matching APIs' })}
+                sx={SWITCHER_SELECT_SX}
+              />
+
+              <IconButton
+                size="small"
+                aria-label={intl.formatMessage({ id: 'appShell.header.api.goToApiLevel', defaultMessage: 'Go to API level' })}
```
AnuGayan: **Severity:** 🟡 minor
**Subject:** Clear button is labelled "Go to API level" but navigates to project level

**Description**
This `X` is a step-*back* affordance: its handler `clearApiSelection` (line 84) navigates to `routes.projectHome(...)`, leaving API scope. The label — and the message id `appShell.header.api.goToApiLevel` — say the opposite. The sibling project clear button at line 210 is labelled correctly ("Go to organization level"). Screen-reader users are told the reverse of what the control does. Confirmed by reading `AppHeader.tsx:84-88` and `:284-302`.

**How to verify**
Open this file at line 292 and compare with `clearApiSelection` at line 84 and the project button's label at line 210.

**Suggested fix**
```suggestion
                aria-label={intl.formatMessage({ id: 'appShell.header.api.goToProjectLevel', defaultMessage: 'Go to project level' })}
```
Then re-run `npm run i18n` to pick up the renamed id.

**Prompt for Claude**
```text
In src/pages/appShell/AppHeader.tsx line 292, the API switcher's clear button is aria-labelled "Go to API level" but its
onClick (clearApiSelection, line 84) navigates to routes.projectHome.
1. First confirm: read lines 84-88 and 284-302 and check clearApiSelection targets projectHome, not an API route.
   If it targets an API route, stop and tell me — the finding is wrong.
2. If confirmed, change the id to 'appShell.header.api.goToProjectLevel' and the defaultMessage to 'Go to project level',
   matching the project switcher's own pattern at line 210.
3. Remove the now-unused 'appShell.header.api.goToApiLevel' entry by running `npm run i18n`, not by hand-editing en.json.
4. Confirm with `npm run typecheck`, `npm test` and `npm run i18n:check`.
```

portals/api-control-plane/src/pages/appShell/appShellPages/projects/ProjectListPage.tsx
```diff
@@ -0,0 +1,476 @@
...
+                <ProjectsGrid
+                  onDelete={setToDelete}
+                  onOpen={openProject}
+                  orgHandle={orgHandle}
+                  projects={projects}
+                />
+              </Box>
+              {total > PAGE_SIZE_OPTIONS[0] && (
+                <Box
+                  sx={[
```
AnuGayan: **Severity:** 🟡 minor
**Subject:** `bottom` without `position`, so the pagination bar never sticks and `useFooterHeight()` is dead

**Description**
`bottom` has no effect on a `position: static` element, and nothing here sets a position — so the comment above ("keeps the pagination bar at the bottom on a half-empty page") describes behaviour that does not happen, and the `useFooterHeight()` call at line 224 is measured for nothing. The `stickyBottomBarSx` recipe supplies exactly the missing `position: 'sticky'`, and its own docstring names "a list page's pagination row" as an intended call site — but its only consumer is `SaveBar.tsx:51`. Confirmed by reading `theme/receipes.ts:95-114`, `SaveBar.tsx:47-53` and this block.

**How to verify**
Open `src/theme/receipes.ts:105-113` (which sets `position: 'sticky'`) beside this line. In the browser, set rows-per-page to 12 with more than 12 projects and scroll — the bar scrolls away instead of pinning.

**Suggested fix**
```suggestion
                  sx={[
                    stickyBottomBarSx,
                    { bottom: `${footerHeight}px`, mx: -8, px: 8 },
                  ]}
```
Add `import { stickyBottomBarSx } from '../../../../theme';`. If the bar was never meant to stick, drop the `bottom` key and the `useFooterHeight()` call instead — but not both halves as they stand.

**Prompt for Claude**
```text
In src/pages/appShell/appShellPages/projects/ProjectListPage.tsx line 424, the pagination wrapper sets `bottom` but never
`position`, so useFooterHeight() at line 224 has no effect.
1. First confirm: read the sx at lines 424-426 and check no `position` is set, and read src/theme/receipes.ts:105-113 to
   confirm stickyBottomBarSx is what supplies `position: 'sticky'`. If the sx already sets a position, stop and tell me.
2. If confirmed, ask me which is intended before editing — either
   (a) make it sticky: add stickyBottomBarSx as the first element of the sx array, importing it from '../../../../theme'
       the way SaveBar.tsx:22 does; or
   (b) it was never meant to stick: drop the `bottom` key and the useFooterHeight() call at line 224, and delete the
       misleading comment above line 420.
   Do NOT guess — report both options and the visual difference.
3. Confirm with `npm run typecheck` and `npm test` (ProjectListPage.test.tsx has 9 tests that must keep passing).
```

portals/api-control-plane/src/pages/appShell/appShellPages/apis/ApiDetailPage.tsx
```diff
@@ -0,0 +1,195 @@
...
+                      </Typography>
+                      <Typography variant="body2">
+                        {detail.context || '/'}
+                      </Typography>
+                    </Stack>
+                    <Stack direction="row" alignItems="center" gap={2}>
+                      <Typography variant="caption" color="text.secondary">
+                        <FormattedMessage
+                          id="aiWorkspace.pages.appShell.appShellPages.proxies.LLMProxyOverview.last.updated"
+                          defaultMessage={'Last updated :'}
```
AnuGayan: **Severity:** 🟡 minor
**Subject:** `defaultMessage={'…'}` fails `formatjs/enforce-default-message` — `npm run lint` exits 1

**Description**
The rule requires a bare string literal attribute, not a JSX expression. This line and `ExploreMoreCard.tsx:99` are the only two lint errors on the branch, and both are in files this PR adds. Lint is not currently a CI gate (`make test-api-control-plane` runs only `npm test` + `bff-test`), so this will not go red on its own — it just leaves `npm run lint` permanently broken for everyone after this merges. Confirmed by running `npm run lint` at `c8fba03`.

Related, in the same area: this PR adds a `.eslintignore`, which ESLint 9's flat config does not read — the run emits `ESLintIgnoreWarning: The ".eslintignore" file is no longer supported` and ignores the file, so its `/public/**` entry has no effect. `eslint.config.js:238` already covers build/coverage/node_modules/generated. Both new ignore files also carry the typo "Pleae".

**How to verify**
```
cd portals/api-control-plane && npm run lint
# ✖ 2 problems (2 errors, 0 warnings)   + ESLintIgnoreWarning at the top
```

**Suggested fix**
```suggestion
                          defaultMessage="Last updated :"
```
Same shape at `ExploreMoreCard.tsx:99`. Then either delete `.eslintignore` or fold `/public/**` into `eslint.config.js`'s `ignores` array.

**Prompt for Claude**
```text
`npm run lint` fails in portals/api-control-plane with 2 formatjs/enforce-default-message errors.
1. First confirm: `cd portals/api-control-plane && npm run lint`. Note the two file:line pairs. If it exits 0, stop and
   tell me — the finding is wrong.
2. At each reported location, change `defaultMessage={'...'}` to a plain string literal attribute `defaultMessage="..."`,
   preserving the exact text. Do not change the id.
3. Separately: .eslintignore is new in this branch but ESLint 9 flat config ignores it (see the ESLintIgnoreWarning).
   Either delete .eslintignore, or move its '/public/**' entry into the `ignores` array at eslint.config.js:238.
   Ask me which before deleting.
4. Also fix the "Pleae" -> "Please" typo in .prettierignore (and .eslintignore if it survives).
5. Confirm with `npm run lint` (exit 0, no ignore warning) and `npm test`.
```

portals/api-control-plane/src/navigation/navigationTypes.ts
```diff
@@ -40,26 +48,33 @@ export type NavigationDefinition = {
    */
   pinned?: boolean;
```
AnuGayan, line 49: **Severity:** 🟡 minor
**Subject:** `pinned` is documented as a `Sidebar.Footer` feature but nothing reads it

**Description**
The doc comment above says this "pins this item to the sidebar's fixed bottom section (`Sidebar.Footer`) instead of the scrolling main nav list. Used for Settings". Nothing sets it (no registry entry), `useNavigationItems.resolve()` never copies it onto the returned item, and `AppSidebar.tsx` never reads it or renders a `Sidebar.Footer` — Settings is an ordinary item in the `global` cluster. Comments in `navigationRegistry.tsx:453` and `AppRoutes.orgSettings.test.tsx:29,65,79` still describe pinning, so the next reader will believe the feature is live. Confirmed by `grep -rn "pinned" src` — every hit is a type declaration or a comment.

**How to verify**
```bash
grep -rn "pinned" portals/api-control-plane/src
grep -rn "Sidebar.Footer" portals/api-control-plane/src   # only this doc comment
```

**Suggested fix**
Delete both `pinned?: boolean` declarations (here and line 77) and reword the comments that reference pinning — or implement it in `AppSidebar`. Deleting is the smaller change unless a bottom-pinned Settings is still wanted.

**Prompt for Claude**
```text
In portals/api-control-plane, `pinned` is declared on both NavigationDefinition and NavigationItem
(src/navigation/navigationTypes.ts:49 and :77) and documented as pinning an item to Sidebar.Footer, but nothing
implements it.
1. First confirm: `grep -rn "pinned" portals/api-control-plane/src` and check every hit is a type declaration or a
   comment — no registry entry sets it, useNavigationItems.ts never copies it onto the resolved item, AppSidebar.tsx
   never reads it. If any code path reads it, stop and tell me.
2. Ask me whether Settings is meant to sit in a pinned bottom section before changing anything. If I say no, remove both
   `pinned?: boolean` declarations and reword the "pinned" wording in navigationRegistry.tsx:453 and
   AppRoutes.orgSettings.test.tsx:29,65,79 (comments only — do not change what those tests assert).
3. Confirm with `npm run typecheck` and `npm test` (AppRoutes.orgSettings.test.tsx must still pass unchanged).
```

portals/api-control-plane/src/pages/appShell/ProjectQuickSelector.tsx
```diff
@@ -0,0 +1,191 @@
...
+  };
+
+  return (
+    <>
+      <Tooltip title={intl.formatMessage({ id: 'appShell.header.project.select', defaultMessage: 'Select Project' })}>
+        <span>
+          <IconButton
+            size="small"
+            aria-label={intl.formatMessage({ id: 'appShell.header.project.select', defaultMessage: 'Select Project' })}
+            // disabled={disabled}
```
AnuGayan: **Severity:** 🔵 nit
**Subject:** `// disabled={disabled}` — the prop is accepted and wired into an effect, but never applied

**Description**
`disabled` is a required prop on this component, is read by the `useEffect` at line 60, and is passed `disabled={!params.orgHandle}` by `AppHeader.tsx:239` — but the line that would apply it to the button is commented out, so the trigger is always enabled. The sibling `APIQuickSelector.tsx:90` applies it live. No user-visible difference today (this component only renders inside a `params.orgHandle && …` guard, so `disabled` is always `false`), but it is dead wiring plus commented-out code. `APIQuickSelector.tsx:97` has the same issue with `// borderRadius: 1,`. Confirmed by reading both files and their call sites.

**How to verify**
Compare line 90 here with `APIQuickSelector.tsx:90`.

**Suggested fix**
```suggestion
            disabled={disabled}
```
If the button is deliberately meant to stay clickable, drop the `disabled` prop and its `useEffect` instead — but not leave it half-wired.

**Prompt for Claude**
```text
In src/pages/appShell/ProjectQuickSelector.tsx line 90, `disabled={disabled}` is commented out, so the `disabled` prop is
accepted and used by the useEffect at line 60 but never applied to the IconButton.
1. First confirm: read lines 40-100 here and lines 40-100 of APIQuickSelector.tsx, and check the sibling applies
   `disabled={disabled}` while this one does not. If both apply it, stop and tell me.
2. Ask me whether the button should be disabled before editing. If yes, uncomment line 90. If no, remove the `disabled`
   prop from the Props type, its destructure, the useEffect at line 60, and the `disabled={!params.orgHandle}` passed at
   AppHeader.tsx:239.
3. Either way, delete the commented-out `// borderRadius: 1,` at APIQuickSelector.tsx:97.
4. Confirm with `npm run typecheck`, `npm run lint` and `npm test`.
```

ShavinAnjithaAlpha: Pushed a fix in commit [b86efcb](https://github.com/wso2/api-platform/pull/3281/commits/b86efcb6310d5aa70bfd887410b1b760477aab20).

portals/api-control-plane/src/pages/appShell/appShellPages/apis/ApiDetailPage.tsx
ShavinAnjithaAlpha: Pushed a fix in commit [b86efcb](https://github.com/wso2/api-platform/pull/3281/commits/b86efcb6310d5aa70bfd887410b1b760477aab20).

portals/api-control-plane/src/pages/appShell/appShellPages/projects/ProjectListPage.tsx
ShavinAnjithaAlpha: Pushed a fix in commit [b86efcb](https://github.com/wso2/api-platform/pull/3281/commits/b86efcb6310d5aa70bfd887410b1b760477aab20).

portals/api-control-plane/src/pages/appShell/AppHeader.tsx
ShavinAnjithaAlpha: Pushed a fix in commit [b86efcb](https://github.com/wso2/api-platform/pull/3281/commits/b86efcb6310d5aa70bfd887410b1b760477aab20).

portals/api-control-plane/src/pages/appShell/AppHeader.tsx
ShavinAnjithaAlpha: Pushed a fix in commit [b86efcb](https://github.com/wso2/api-platform/pull/3281/commits/b86efcb6310d5aa70bfd887410b1b760477aab20).

portals/api-control-plane/src/scope/ScopeGate.tsx
ShavinAnjithaAlpha: Pushed a fix in commit [b86efcb](https://github.com/wso2/api-platform/pull/3281/commits/b86efcb6310d5aa70bfd887410b1b760477aab20).

portals/api-control-plane/src/i18n/messages/en.json
ShavinAnjithaAlpha: Pushed a fix in commit [b86efcb](https://github.com/wso2/api-platform/pull/3281/commits/b86efcb6310d5aa70bfd887410b1b760477aab20).

portals/api-control-plane/src/scope/ScopeGate.tsx
ShavinAnjithaAlpha: Pushed a fix in commit [b86efcb](https://github.com/wso2/api-platform/pull/3281/commits/b86efcb6310d5aa70bfd887410b1b760477aab20).

portals/api-control-plane/src/pages/appShell/AppHeader.tsx
ShavinAnjithaAlpha: Pushed a fix in commit [b86efcb](https://github.com/wso2/api-platform/pull/3281/commits/b86efcb6310d5aa70bfd887410b1b760477aab20).

portals/api-control-plane/src/pages/appShell/appShellPages/apis/ApiDetailPage.tsx
ShavinAnjithaAlpha: Pushed a fix in commit [b86efcb](https://github.com/wso2/api-platform/pull/3281/commits/b86efcb6310d5aa70bfd887410b1b760477aab20).

portals/api-control-plane/package.json
ShavinAnjithaAlpha: Pushed a fix in commit [b86efcb](https://github.com/wso2/api-platform/pull/3281/commits/b86efcb6310d5aa70bfd887410b1b760477aab20).

portals/api-control-plane/src/navigation/navigationTypes.ts
ShavinAnjithaAlpha: Confirmed unimplemented, but intentional. The pinned `Sidebar.Footer` section for Settings is still planned, and the declaration records that intent. 

Leaving it in place; implementing it needs pinned.

Induwara04 (dismissed):

portals/api-control-plane/src/extensions.tsx
ShavinAnjithaAlpha: Pushed a fix in commit [1d4557b](https://github.com/wso2/api-platform/pull/3281/commits/1d4557b018e8ae565d4a25da5d12fe566fbc4d68).

Induwara04 (approved):

Issue #3287, opened 2026-08-23
Remove or implement unwired APIGateway CRD spec fields


renuka-fernando:
### Description

The `APIGateway` CRD (`gateway.api-platform.wso2.com/v1`) defines several typed `spec` fields that the API server accepts but the operator's reconcile/deploy path silently ignores, so setting them has no effect on the deployed gateway. The only field that actually configures the gateway today is `spec.configRef` (a ConfigMap of Helm values deep-merged over the operator default). This is a confusing false affordance — e.g. a user sets `spec.infrastructure.image` or `spec.infrastructure.replicas` and nothing changes.

**Root cause:** `buildCRValuesOverlay` (`kubernetes/gateway-operator/internal/controller/apigateway_controller.go:757-779`), the only function that turns `spec.infrastructure` into Helm values, reads only `infra.Labels` and `infra.Annotations`. Nothing reads `infra.Image`/`RouterImage`/`Replicas`/`Resources`/`NodeSelector`/`Tolerations`/`Affinity`, `spec.storage`, or `spec.controlPlane.tls`/`tokenSecretRef`. `spec.controlPlane.host` is only recorded in the operator's in-memory registry (`registerGatewayInRegistry`), never applied to the deployed gateway.

**What is wired vs not (APIGateway CRD path):**

| Field | Status |
|---|---|
| `spec.apiSelector` | ✅ implemented (required) |
| `spec.configRef` | ✅ implemented (the real config surface) |
| `spec.infrastructure.labels` / `annotations` | ✅ implemented (→ `commonLabels`/`commonAnnotations`) |
| `spec.controlPlane.host` | ⚠️ recorded in operator registry only, not deployed |
| `spec.controlPlane.tls` / `tokenSecretRef` | ❌ ignored |
| `spec.infrastructure.image` / `routerImage` | ❌ ignored |
| `spec.infrastructure.replicas` / `resources` | ❌ ignored |
| `spec.infrastructure.nodeSelector` / `tolerations` / `affinity` | ❌ ignored |
| `spec.storage.*` | ❌ ignored |

**Sample CR showing the gap:**

```yaml
apiVersion: gateway.api-platform.wso2.com/v1
kind: APIGateway
metadata:
  name: sample-gw
  namespace: sample-gw
spec:
  apiSelector:                 # ✅ implemented (required)
    scope: Namespaced
  configRef:                   # ✅ implemented — the real config surface
    name: sample-gw-values
  infrastructure:
[...]

PR #3288, opened 2026-08-23
docs(operator): add debug guide and per-gateway version sample

renuka-fernando:
## Purpose

The gateway operator had no debug guide and no sample for running different gateway versions per `APIGateway`, and `make install` was broken because its CRD kustomization still referenced the pre-rename `_gateways.yaml` and omitted most CRDs. This PR adds the missing docs and sample and fixes the CRD kustomization. It also documents that the typed `spec.infrastructure.image`/`routerImage` fields are not wired yet (tracked in #3287) and that the gateway version is set via the `configRef` ConfigMap.

Related #3287

## Goals

- Give contributors a working host-run (VS Code) debug workflow for the operator.
- Show how to run different gateway (data-plane) versions per `APIGateway` from a single operator.
- Make `make install` work again by listing the current CRDs.

## Approach

- Add `kubernetes/gateway-operator/DEBUG_GUIDE.md` for running the operator from VS Code against a cluster: install the operator's own CRDs plus the Gateway API standard CRDs, provision the mandatory AES-256 at-rest encryption key Secret per namespace, launch the debugger, and trigger a reconcile.
- Add `config/samples/api_v1_apigateway_multiversion.yaml` showing one gateway that tracks the operator default version and one pinned to a specific version via its own `configRef`, i.e. per-gateway image/tag overrides from a single operator.
- Document per-gateway versioning in the operator README: value precedence (configRef > CR labels/annotations > operator default > chart defaults), using `spec.infrastructure.annotations` rather than `metadata.annotations` to force a redeploy, and the caveat that operator-default changes only auto-apply to new gateways.
- Fix `config/crd/kustomization.yaml`, which referenced the pre-rename `_gateways.yaml` and listed only 2 of 12 CRDs (breaking `make install`); it now lists all 12 current CRD bases.

## User stories

- As a contributor, I want to run and debug the gateway operator locally against a real cluster.
- As an operator user, I want one gateway to track the operator default version and another pinned to a specific version, from a single operator.

## Documentation

This PR is itself operator documentation (`DEBUG_GUIDE.md` and a new README section). No separate product-documentation impact.

## Automation tests

- Unit tests: N/A — this PR only changes docs, a sample manifest, and the CRD kustomization resource list; no Go code is changed.
- Integration tests: N/A. The `make install` fix was verified manually via `kubectl kustomize config/crd`, which now renders all 12 CRDs.

## Security checks

- Followed secure coding standards: yes.
- Ran FindSecurityBugs plugin and verified report: N/A — no Java/compiled code changed.
- Confirmed that this PR doesn't commit any keys, passwords, tokens, usernames, or other secrets: yes. The sample and debug guide instruct generating an AES-256 key locally via `openssl` (created owner-only with `umask 077`) and never commit key material.

## Samples

- `kubernetes/gateway-operator/config/samples/api_v1_apigateway_multiversion.yaml` — `gw-default` (inherits the operator default version) and `gw-121` (pins both gateway images to `1.2.1`).

## Related PRs

N/A

## Test environment

N/A — documentation and sample manifests. The debug workflow steps were validated manually with `kubectl` against a local cluster.

### DIFF
--- a/kubernetes/gateway-operator/DEBUG_GUIDE.md
+++ b/kubernetes/gateway-operator/DEBUG_GUIDE.md
@@ -1,0 +1,73 @@
+# Gateway Operator — Debug Guide
+
+Run the operator from your host (VS Code debugger) against a real cluster. All commands run from `kubernetes/gateway-operator/`.
+
+## Prerequisites
+
+1. Go toolchain + VS Code with the Go extension.
+2. `kubectl` pointing at the target cluster (the operator uses your current kubeconfig context).
+3. A reachable cluster (kind / k3s / colima / minikube / remote).
+
+## Steps
+
+1. **Confirm your cluster context:**
+   ```bash
+   kubectl config current-context && kubectl cluster-info
+   ```
+
+2. **Install the Gateway API standard CRDs** (the operator watches `Gateway`, `GatewayClass`, `HTTPRoute`, `ReferenceGrant`). Skip if already present:
+   ```bash
+   kubectl get crd gateways.gateway.networking.k8s.io >/dev/null 2>&1 || \
+     kubectl apply -f ../helm/operator-helm-chart/files/gateway-api-standard/standard-crds.yaml
+   ```
+   > Uses the in-repo bundle (v1.5.1) the operator compiles against.
+
+3. **Install the operator's own CRDs** (`gateway.api-platform.wso2.com`):
+   ```bash
+   make install
+   ```
+   > Regenerates the CRD manifests (controller-gen) and applies them via kustomize. Installs only this operator's CRDs — the Gateway API set in step 2 is separate.
+   > Fallback (no regeneration / avoids client-side apply limits): `kubectl apply --server-side -f config/crd/bases/`
+
+4. **Verify all CRDs are registered:**
+   ```bash
+   kubectl get crd | grep -E 'gateway.api-platform.wso2.com|gateway.networking.k8s.io'
+   ```
+
+5. **Launch the debugger:** VS Code → Run and Debug → **Gateway Operator**.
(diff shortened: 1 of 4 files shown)


renuka-fernando:
Thanks for the review. Addressed all 5 findings in `48378dceb`:

1. **`gw-default` default-tracking wording** — corrected in the sample header: it inherits the operator default only at (re)deploy time, and an existing gateway needs an `APIGateway` generation bump to pick up a default change (it does not auto-move). The README already stated this (the "Changing the operator default only auto-applies to new gateways" note), so no README change was needed.
2. **Temp AES key file permissions** — wrapped the key generation in `( umask 077 … )` in both the sample header and `DEBUG_GUIDE.md`, so the temp `default-aesgcm256-v1.bin` is owner-only regardless of the caller's umask.
3. **Gateway API CRD guard** — replaced the single-CRD `kubectl get … || apply` guard with an unconditional `kubectl apply` (idempotent), so a partial CRD set is no longer silently skipped.
4. **Stale namespace name** — `gw-120` → `gw-default` in the `DEBUG_GUIDE.md` step-8 secret instruction, matching the sample.
5. **CRD delete warning** — added a data-loss warning that `kubectl delete -f config/crd/bases/` removes cluster-scoped CRDs and cascades to every custom resource in all namespaces; use only on a disposable cluster.

Also expanded the PR description to the full template (goals, user stories, documentation, tests, security, samples, test environment) to clear the description check.

renuka-fernando:
@coderabbitai review

renuka-fernando:
@coderabbitai full review

renuka-fernando:
@coderabbitai review

Thushani-Jayasekera (dismissed):

renuka-fernando:
Addressed the follow-up finding in `607a38fcd`: switched both the sample and `DEBUG_GUIDE.md` to write the AES-256 key into a unique `mktemp` file (mode 600) with an EXIT-trap cleanup, instead of redirecting into a fixed `default-aesgcm256-v1.bin` path (which `umask 077` wouldn't tighten if the file already existed). The Secret's data key remains `default-aesgcm256-v1.bin` via `--from-file=default-aesgcm256-v1.bin="$keyfile"`.

Thushani-Jayasekera (approved):

PR #3310, opened 2026-08-25
[Backend] Add GraphQL API Support for API Platform

### LINKED ISSUE #3195: [Feature]: GraphQL support for API Platform

Gateway, Developer Portal, Platform API

Aspect/API (API backends, definitions, contracts, interfaces, OpenAPI), Aspect/UI (Frontend layouts, components, styling), Aspect/UX (User experience, flows, usability, clarity)

### Suggested Feature

Currently, API Platform Gateway supports only REST APIs. To remain competitive and address diverse customer needs, it is essential to onboard support for other API types. GraphQL, being an emerging API standard and as it operates over existing transport protocols, must be implemented.

This will include;

- Support GraphQL as a first-class API type
- Allow creation of a GraphQL API similar to approaches used for REST APIs
- Provide QoS features for GraphQL
- API Console and API Portal support for GraphQL APIs

### Related Issues

_No response_

### Steps to Verify

- [x] **Design Document** — A detailed design document has been created and reviewed, covering architecture, data flow, and edge cases.
- [x] **Design Mail** — A design summary email has been sent to relevant stakeholders for awareness and feedback.
- [x] **Code Review** — All code changes have been peer-reviewed and approved according to the project's review standards.
- [ ] **Testing Complete** — Adequate unit, integration, and/or end-to-end tests have been written and are passing.
- [ ] **Documentation Review** — User-facing and/or developer documentation has been updated to reflect the new feature and reviewed.
- [ ] **Feature Complete** — The feature is fully implemented, all checklist items above are done, and it is ready for release.

npamudika:
## Purpose
Adds first-class support for GraphQL as a managed API kind across the platform, matching the existing capabilities REST APIs already have. Previously, exposing a GraphQL backend through the gateway required treating it as a generic REST API, which meant no purpose-built handling for GraphQL's single-endpoint, single-route shape (one POST route regardless of query/mutation), no gateway-controller or control-plane CRUD for it, no API key management, and no CLI support. This closes that gap end-to-end: gateway-only proxying, control-plane management, API key lifecycle, and CLI tooling.

Resolves https://github.com/wso2/api-platform/issues/3195

## Goals

- Introduce GraphQLApi as a core artifact kind (not a plugin), deployable and manageable the same way RestApi/Mcp/LlmProvider/LlmProxy already are.
- Support the full lifecycle gateway-only (direct against gateway-controller, no control plane): create/update/list/get/delete, policy attachment, and API key management (create/list/regenerate/update/revoke).
- Support the full lifecycle via the control plane (platform-api): CRUD, deployment (deploy/undeploy/restore), SDL resolution (inline, file upload, fetch-by-URL, or live introspection), and API key management, reusing existing shared infrastructure (repository interfaces, scope/role mapping, gateway event translation) rather than duplicating it.
- Add ap CLI parity for the gateway-only path (ap gateway graphql-api ... and ap gateway graphql-api api-key ...), mirroring the existing rest-api command family.
- Ensure a GraphQL API's single-route shape is correctly represented at every layer (Envoy routing, policy chain resolution, deployment YAML) without carrying over REST-specific assumptions (e.g. no per-operation operations[] list, no SDL stored gateway-side).

## Approach

- DB schema (e5a7be003): added graphql_apis tables across all supported dialects (SQLite/PostgreSQL/SQL Server) on both gateway-controller and platform-api.
- OpenAPI specs (f23e9bd13): added the /graphql-apis resource family to platform-api's control-plane spec (CRUD, API keys, deployments, gateway associations) and to gateway-controller's management spec, regenerating server code for both.
- Gateway-only path (b8f7686d6): GraphQLAPITransformer builds exactly one Exact-match POST route per API (no per-operation routing); wired into the existing generic transform/policy-chain-resolution pipeline with no core routing changes needed. New handler, storage wiring, and a 30-scenario E2E suite (graphql_deploy.feature) covering CRUD, routing, policy attachment (auth/CORS/rate-limit/set-headers), sandbox routing, and confirmed known limitations (CORS preflight doesn't work for a single-route API — documented, not fixed, as it would require adding an OPTIONS route).
- CLI (d424e8285): new ap gateway graphql-api command group mirroring rest-api's structure. apply (create/update) needed no new code — it's already a generic, kind-dispatched command.
- API key management on gateway-controller (05217e6e5): added the /graphql-apis/{id}/api-keys endpoints, reusing the existing kind-agnostic APIKeyService (only the HTTP handlers are new). Found and fixed a real bug in the process: new routes were missing from a hand-maintained auth route-map, causing every request to be denied as 404 regardless of the OpenAPI spec — a regression-guard test was added.
- Control-plane support (04d726a4c): full CRUD/deployment/SDL-resolution/API-key stack on platform-api, wired into the same shared cross-kind infrastructure every other artifact kind uses (no parallel/duplicate plumbing).

## User stories

- As a gateway operator, I can deploy, update, and manage a GraphQL API directly against the gateway, with policies (auth, rate limiting, header mediation) applying the same way they do for REST APIs.
- As a platform administrator, I can onboard a GraphQL API through the control plane by providing its schema inline, via file upload, via URL, or via live introspection, and manage its full deployment lifecycle.
- As an API consumer/integrator, I can generate, list, rotate, and revoke API keys scoped to a GraphQL API, on both the gateway-only and control-plane paths.
- As a CLI user, I can manage GraphQL APIs and their API keys with ap gateway graphql-api ... the same way I already do for REST APIs with ap gateway rest-api ....

## Documentation
N/A for now

## Automation tests

- Unit tests

  -   gateway/gateway-controller: transformer tests (graphql_test.go), 22 API key handler tests (graphql_apikey_handler_test.go) covering all 5 operations × {no-auth, invalid-body, success, DB-error, not-found}, plus a route-auth regression guard in main_test.go.
  -   cli/: 24 tests across graphqlapi/commands_test.go (8) and graphqlapi/apikey/commands_test.go (16), using a new testutil.NewGatewayServer harness.
  -   platform-api: extensive coverage across graphql_api_test.go, graphql_deployment_test.go, graphql_apikey_test.go, graphql_gateway_test.go, graphql_multipart_test.go, artifact_tables_test.go, pagination_test.go, resources_test.go (~35 files touched, most new).
  -   All existing unit test suites in every touched module re-verified green after these changes (go build/go vet/go test ./... clean in gateway-controller, cli, and platform-api).

- Integration tests

  -   graphql_deploy.feature — 30 scenarios: CRUD, filtering/pagination, validation errors, single-route-exactly enforcement, policy attachment (jwt-auth, cors, set-headers, basic-ratelimit), sandbox routing, mutation pass-through.
  -   graphql-api-keys.feature — 15 scenarios mirroring the existing REST api-keys.feature scenario-for-scenario: full lifecycle, multi-key, empty-list, 404 cases, invalid JSON, special characters, pagination.
  -   All verified against a real, freshly-built gateway-controller (and gateway-runtime where relevant) — not mocked.

## Security checks

- Followed secure coding standards in the WSO2 secure engineering guidelines? Partially — followed this repo's own codified security rules (.claude/rules/*.md, covering auth/authz, SSRF, file access, XXE, dependency management, etc.) throughout; have not separately checked against the WSO2 guidelines document itself.
- Ran FindSecurityBugs plugin and verified report? N/A — this is a Go/TypeScript codebase; FindSecurityBugs is a Java/SpotBugs plugin and doesn't apply.
- Confirmed that this PR doesn't commit any keys, passwords, tokens, usernames, or other secrets? Yes — verified via git status/diff review before each commit; no credential files were staged.

## Samples
gateway/examples/countries-graphql-api.yaml and gateway/examples/blog-graphql-api.yaml — sample gateway-only GraphQL API deployment manifests, added in b8f7686d6.

## Related PRs
N/A

## Test environment

- OS: macOS 15.7.3 (Darwin 24.6.0), arm64
- Go: go1.26.5 (darwin/arm64)
- Database: SQLite (local dev/test default) — verified live; PostgreSQL and SQL Server schema files were added for all three dialects but not exercised against real Postgres/SQL Server instances in this round of testing
- Container runtime: Docker (via Rancher Desktop) for gateway-controller/gateway-runtime integration tests
- JDK / browser: not applicable — no JVM or browser-facing component in this change


platform-api/resources/openapi.yaml
```diff
@@ -6182,6 +6988,278 @@ components:
...
+            the endpoint introspected at creation time (see `sdl`/`introspectionMode` below).
+        lifeCycleStatus:
+          type: string
+          x-go-type: RESTAPILifeCycleStatus
+          description: |
+            Current lifecycle status of the API. Reuses REST APIs' lifecycle enum
+            unmodified (STAGED, CREATED, PUBLISHED, DEPRECATED, RETIRED, BLOCKED) —
+            declaring a second identically-valued enum schema here would collide
+            with it at Go-constant generation time.
+          example: "CREATED"
```
malinthaprasan: Not required

platform-api/resources/openapi.yaml
```diff
@@ -6182,6 +6988,278 @@ components:
...
+        kind:
+          type: string
+          description: Kind of the API based on its communication protocol or architectural style
+          example: GraphQLApi
+          default: GraphQLApi
+        sdl:
```
malinthaprasan, line 7125: Without exposing here, expose from GET /graphql-apis/{id}/sdl

platform-api/resources/openapi.yaml
```diff
@@ -6182,6 +6988,278 @@ components:
...
+            counterpart to how an OpenAPI document can be supplied by reference for
+            other artifact kinds (see LlmProviderTemplate's `metadata.openapiSpecUrl`).
+            Distinct from `upstream.main.url`: this is a plain HTTP(S) GET of a static
+            schema file, not a live introspection query against a GraphQL server, and
+            is fetched with the same public-internet-only SSRF hardening as an
+            OpenAPI-spec-by-URL fetch (loopback/private/link-local/metadata addresses
+            refused) — it is not meant for a tenant's own in-cluster backend. Mutually
+            exclusive with `sdl`. Never stored or echoed back; only the fetched `sdl`
+            text is persisted and returned.
+          example: https://raw.githubusercontent.com/example/countries-api/main/schema.graphql
```
malinthaprasan, line 7166: Not required to store after the API is created.

platform-api/resources/openapi.yaml
```diff
@@ -6182,6 +6988,278 @@ components:
...
+          type: string
+          format: date-time
+          readOnly: true
+        # sdl is deliberately omitted from list responses — unlike a REST API's
+        # operations[] (a handful of entries), a GraphQL schema's SDL can run to
+        # many KB; embedding it in every list-page row would bloat GET
+        # /graphql-apis for no listing-view benefit. Fetch GET
+        # /graphql-apis/{graphqlApiId} for the full schema.
+
+    GraphQLAPIListResponse:
```
malinthaprasan, line 7423: Lets remove this comment

platform-api/resources/openapi.yaml
npamudika: sdlUrl is a write-only convenience field — a third way to supply the schema on Create/Update, alongside pasting sdl directly or uploading a file (sdlFile via multipart). Instead of the caller fetching a raw .graphql schema file themselves and pasting its contents into sdl, they can just point the server at the file's URL (GitHub raw link) and let the server fetch it server-side.

AnuGayan (commented):
## Review — GraphQL API support

Reviewed the full diff at `eab9583` (97 files, +15,519/-405), with the branch checked out locally. `go build` and `go vet` are clean on both Go modules and the new unit tests pass (`./gateway/gateway-controller/pkg/transform/...`, `./platform-api/internal/utils/...`).

### Summary

The quality of the individual layers is high — `graphql_apis` matches its peer tables column-for-column across all four dialects, `GraphQLAPITransformer` correctly models GraphQL's single-route shape, the SSRF-guarded fetch client is reused rather than reinvented, and the shared gateway-association SQL in `repository/api.go` was properly extracted instead of copy-pasted (I checked every public method signature — no call-site regressions).

The problem is the **control-plane ↔ data-plane seam**, which the description lists as a goal ("Support the full lifecycle via the control plane … deployment (deploy/undeploy/restore)") but which is not wired in either direction:

- **CP → DP:** platform-api broadcasts `graphqlapi.deployed` / `.undeployed` / `.deleted`, and `gateway/gateway-controller/pkg/controlplane/client.go:1413-1478` has no case for any of them. They hit `default:` and are logged as "unknown event type". The deployment returns `201 DEPLOYING` and stays there forever; no route is ever programmed. `grep -rn "graphql" gateway/gateway-controller/pkg/controlplane/` returns zero hits.
- **DP → CP:** the gateway pushes GraphQL artifacts on every create/update (`graphql_api_handler.go:317`), and `ArtifactImportService.importers` has no `GraphQLApi` entry, so the push is rejected with `Unsupported artifact kind "GraphQLApi"` — visible only in logs.

Both fail silently, which is what makes them severe rather than merely incomplete. The gateway-only path (which the 30-scenario `graphql_deploy.feature` suite actually exercises) looks sound; it's the control-plane path — untested end to end by design, since the integration suite runs against gateway-controller directly — where the gaps sit.

### Verdict

🔧 **Request changes** — 1 blocker, 3 major, 3 minor, 4 nits, all inline.

| # | Sev | Finding |
|---|-----|---------|
| 1 | 🔴 | No gateway handler for `graphqlapi.*` events → CP deployment is a silent no-op |
| 2 | 🟠 | No `GraphQLApi` artifact importer → every DP→CP push rejected |
| 3 | 🟠 | Deployment YAML drops `upstream.sandbox`, which every other layer supports |
| 4 | 🟠 | `sdlUrl` documents public-only SSRF hardening the code does not implement |
| 5 | 🟡 | Introspection's 15s timeout is discarded; no request context (effective bound 60s) |
| 6 | 🟡 | `validateGraphQLUpstream` doesn't reject `ref` though the spec says it does |
| 7 | 🟡 | CORS-preflight limitation documented only in the `.feature` file |
| 8-11 | 🔵 | `gofmt` fails on 3 files; comment cites a non-existent `maxGraphQLSDLFileBytes`; main/sandbox share one `*PolicyChain`; event-gateway comment overstates route reachability |

Note on #3: `graphql_api_test.go:602-662` asserts a stored sandbox upstream (with auth) round-trips through the read response, so the CP genuinely accepts and returns a field that never reaches the gateway. And on #5/#6: I confirmed the `Update`-re-introspects-when-`sdl`-is-omitted behavior is deliberate (`graphql_api_test.go:1179-1202` asserts the `SDL`→`ENDPOINT` flip), so I have **not** flagged that.

### What's good

- **Schema discipline.** `VARCHAR(40)` ids, `VARCHAR(255)` display name, `VARCHAR(30)` version, `VARCHAR(1023)` description, `BYTEA`/`BLOB`/`VARBINARY(MAX)` configuration, `TIMESTAMPTZ`/`DATETIME2(7)` per dialect, both peer indexes present in all four files, and the SQL Server `ON DELETE NO ACTION` workaround for error 1785 carried over with a comment explaining why. Nothing diverges from `rest_apis`/`mcp_proxies`.
- **The `repository/api.go` refactor is done right** — the four `artifact_gateway_mappings` helpers became kind-agnostic package functions with every `*APIRepo` method signature preserved.
- **Security posture of the new input surfaces.** `ParseGraphQLAPIMultipartRequest` bounds the body with `MaxBytesReader` independently of `Content-Length`, re-checks the read length against the untrusted `fileHeader.Size`, and cleans up spilled temp files. The API-key handlers pass `constants.GraphQLApi` + `orgId` on every call — no cross-kind or cross-org IDOR. Errors map to a sterile 422 rather than echoing parser output.
- **The auth route-map fix and its regression guard.** Catching that a route present in the spec and `ServerInterface` but absent from the hand-maintained map is denied by default, then locking all ten new routes down in `main_test.go` — I verified all ten spec routes are covered.
- **Comments that carry real weight**, e.g. why `resolveUpstreamCluster` became a package function, why `collectAPIPolicies` was duplicated instead, and why the route key must follow `METHOD|PATH|VHOST` (`transform/graphql.go:105-109`).

Every inline comment carries a **Prompt for Claude** that tells the agent what evidence to check *first*, so a mistaken finding of mine can't turn into a bad commit.

---

platform-api/internal/service/graphql_deployment.go
```diff
@@ -0,0 +1,592 @@
...
+		return nil, fmt.Errorf("failed to set deployment status for GraphQL API: %w", err)
+	}
+
+	if s.gatewayEventsService != nil {
+		deploymentEvent := &model.GraphQLAPIDeploymentEvent{
+			ApiId:        apiModel.ID,
+			DeploymentID: deploymentID,
+			PerformedAt:  performedAt,
+		}
+		if err := s.gatewayEventsService.BroadcastGraphQLAPIDeploymentEvent(gatewayID, deploymentEvent); err != nil {
```
AnuGayan, line 285: **Severity:** 🔴 blocker
**Subject:** No gateway-controller event handler for `graphqlapi.deployed` / `.undeployed` / `.deleted`

**Description**
This broadcast (and the ones at lines 366 and 446) is the *only* mechanism by which a control-plane deployment reaches the gateway. But the gateway's WebSocket dispatcher at `gateway/gateway-controller/pkg/controlplane/client.go:1413-1478` has cases for `api.*`, `llmprovider.*`, `llmproxy.*`, `mcpproxy.*`, `websub.*` and `webbroker.*` — and none for `graphqlapi.*`. All three fall into `default:` and are logged as "Received unknown event type". No route is programmed, no `sendDeploymentAck` is issued, and the deployment row stays at `DEPLOYING` forever while the API returns `201`. There is no CP→DP artifact reconciliation to compensate — the only `sync*` functions on the client cover subscription plans, subscriptions and API keys. Confirmed against `client.go:1413-1478`, this file's lines 230-297, and a repo-wide grep for the event names.

**How to verify**
```bash
grep -rn "graphqlapi\." gateway/gateway-controller/pkg/controlplane/   # no output
```
End to end: `POST /graphql-apis`, then `POST /graphql-apis/{id}/deployments`. Response is `201 status: DEPLOYING`; gateway log shows `Received unknown event type type=graphqlapi.deployed`; `POST <context>` 404s at the router; status never leaves `DEPLOYING`.

**Suggested fix**
Add the three cases plus handlers mirroring the MCP trio (`handleMCPProxyDeploymentEvent` at `client.go:2692`, `handleMCPProxyUndeploymentEvent` at `:2806`, `handleMCPProxyDeletedEvent`). This also needs `APIUtilsService.FetchGraphQLAPIDefinition` — no such method exists (`FetchMCPProxyDefinition` at `pkg/utils/api_utils.go:605` is the template) — and the corresponding gateway-facing definition endpoint on the control plane. I have not validated a snippet against a live deploy, so I'm describing the shape rather than handing over untested code.

**Prompt for Claude**
```text
In gateway/gateway-controller/pkg/controlplane/client.go, the eventType switch at
line 1413 has no case for the graphqlapi.* events that platform-api broadcasts.

1. FIRST confirm the finding: run
   `grep -rn "graphqlapi\." gateway/gateway-controller/pkg/controlplane/`
   and confirm it returns nothing, and that
   platform-api/internal/service/gateway_events.go defines
   EventTypeGraphQLAPIDeployed/Undeployed/Deleted as "graphqlapi.deployed",
   "graphqlapi.undeployed", "graphqlapi.deleted". Also confirm no CP->DP artifact
   reconciliation covers GraphQL. If a handler or reconciler already exists, STOP
   and tell me -- the finding is wrong.
2. If confirmed, add `case "graphqlapi.deployed"/"graphqlapi.undeployed"/
   "graphqlapi.deleted"` and implement handleGraphQLAPIDeploymentEvent /
   UndeploymentEvent / DeletedEvent modeled exactly on
   handleMCPProxyDeploymentEvent (line 2692), handleMCPProxyUndeploymentEvent
   (line 2806) and handleMCPProxyDeletedEvent -- including the sendDeploymentAck
   calls on every failure branch, with resourceType "graphqlapi".
3. This needs a definition fetch: add APIUtilsService.FetchGraphQLAPIDefinition
   in pkg/utils/api_utils.go mirroring FetchMCPProxyDefinition (line 605), and
   verify the control plane actually serves a gateway-facing
   /graphql-apis/{id} definition endpoint. If it does not, report that back
   rather than inventing one.
4. Do not change the gateway-only (direct-to-gateway-controller) path, the
   GraphQLAPITransformer, or any existing event case.
5. Confirm afterwards: deploy a GraphQL API through the control plane and check
   the deployment status reaches DEPLOYED and POST <context> proxies to the
   backend -- not just that unit tests pass.
```

gateway/gateway-controller/pkg/api/handlers/graphql_api_handler.go
```diff
@@ -0,0 +1,585 @@
...
+	if result.IsStale {
+		return
+	}
+	if s.controlPlaneClient == nil || !s.controlPlaneClient.IsConnected() || s.controlPlaneClient.IsOnPrem() ||
+		!s.systemConfig.Controller.ControlPlane.DeploymentSyncEnabled {
+		return
+	}
+	cfgID := result.StoredConfig.UUID
+	deployedAt := result.StoredConfig.DeployedAt
+	s.controlPlaneClient.SubmitArtifactPush(func() {
```
AnuGayan, line 334: **Severity:** 🟠 major
**Subject:** `ArtifactImportService` has no `GraphQLApi` importer, so every DP→CP push is rejected

**Description**
This push runs on every GraphQL create/update whenever `DeploymentSyncEnabled` is set. On the receiving side, `ArtifactImportService.importers` (`platform-api/internal/service/artifact_import.go:137-143`) registers importers for `RestApi`, `LLMProvider`, `LLMProviderTemplate`, `LLMProxy` and `MCPProxy` only; `artifact_import.go:228-232` then returns `ValidationFailed: Unsupported artifact kind "GraphQLApi".` A GraphQL API created directly on a CP-attached gateway works locally but never appears in the control plane, and the failure surfaces only as a warning log and a `cp_sync_status` row — never to the user. This is the mirror image of the missing event handler: the CP↔DP contract is unimplemented in both directions. Confirmed against `artifact_import.go:137-143` and `:225-232`, plus `grep -i graphql platform-api/internal/service/artifact_import*.go` returning nothing.

**How to verify**
```bash
grep -ri graphql platform-api/internal/service/artifact_import*.go   # no output
```
With `DeploymentSyncEnabled=true`, create a GraphQL API on the gateway and watch the platform-api log for `Failed to import gateway artifact in bulk push … kind=GraphQLApi`.

**Suggested fix**
Add a `graphqlAPIImporter` alongside the others, modeled on `artifact_import_mcp.go` (the closest peer — MCP likewise has no operations list):
```go
        constants.MCPProxy:            newMCPProxyImporter(mcpProxyRepo, artifactRepo, mcpServerInfo),
        constants.GraphQLApi:          newGraphQLAPIImporter(graphqlAPIRepo, artifactRepo),
```
This requires threading `graphqlAPIRepo` into `NewArtifactImportService` at `platform-api/internal/server/server.go` (~line 305) and checking `utils.ArtifactImportRank` gives `GraphQLApi` a sensible ordering.

**Prompt for Claude**
```text
platform-api/internal/service/artifact_import.go:137-143 registers gateway-artifact
importers by kind and has no entry for constants.GraphQLApi, while the gateway
pushes GraphQL artifacts (gateway/gateway-controller/pkg/api/handlers/
graphql_api_handler.go:317 pushDeployableGraphQLArtifact).

1. FIRST confirm: check the importers map has no constants.GraphQLApi key, and that
   artifact_import.go:228-232 returns ValidationFailed for an unregistered kind.
   If an importer exists elsewhere (a plugin, a build-tagged file), STOP and say so.
2. If confirmed, create platform-api/internal/service/artifact_import_graphql.go
   defining a graphqlAPIImporter modeled directly on artifact_import_mcp.go's
   mcpProxyImporter (Kind() returns constants.GraphQLApi), and register it in the
   importers map. Thread repository.GraphQLAPIRepository into
   NewArtifactImportService and its call site in internal/server/server.go.
3. Check utils.ArtifactImportRank handles GraphQLApi so batch ordering is defined.
4. Do not change the behavior of any existing importer or the continue-on-error
   semantics of ImportArtifacts.
5. Confirm afterwards with a unit test in the style of
   artifact_import_lifecycle_test.go asserting a GraphQLApi push lands as an
   artifact with Origin=OriginDP and Type=constants.GraphQLApi.
```

platform-api/internal/dto/graphql_api.go
```diff
@@ -0,0 +1,72 @@
...
+	Context           string           `yaml:"context"`
+	SubscriptionPlans []string         `yaml:"subscriptionPlans,omitempty"`
+	Upstream          *GraphQLUpstream `yaml:"upstream,omitempty"`
+	Policies          []Policy         `yaml:"policies,omitempty"`
+}
+
+// GraphQLUpstream represents the upstream configuration for the GraphQL API
+// deployment YAML — a single logical endpoint (no sandbox split, unlike REST).
+type GraphQLUpstream struct {
+	Main *GraphQLUpstreamTarget `yaml:"main,omitempty"`
```
AnuGayan: **Severity:** 🟠 major
**Subject:** Deployment YAML silently drops `upstream.sandbox` for control-plane GraphQL APIs

**Description**
`GraphQLUpstream` has only a `Main` field, and `generateGraphQLAPIDeploymentYAML` (`platform-api/internal/service/graphql_deployment.go:98-108`) populates only `Main`. But the sandbox upstream is supported at every *other* layer: the CP OpenAPI `GraphQLAPI.upstream` `$ref`s the shared `Upstream` schema (which has `sandbox`), `mapUpstreamAPIToModel` persists it, `mapUpstreamConfigToDTO` echoes it back — `graphql_api_test.go:622-662` explicitly asserts a stored sandbox upstream with auth survives to the read response — and `GraphQLAPITransformer` builds a full sandbox route for it (`gateway/gateway-controller/pkg/transform/graphql.go:135-170`). A user can configure a sandbox upstream, see it accepted and returned by the API, and never get a sandbox route, with no warning. The doc comment above ("no sandbox split, unlike REST") contradicts the gateway spec's own `GraphQLAPIConfigData.Upstream.Sandbox` field. Confirmed against this file's lines 55-59, `graphql_deployment.go:98-108`, `transform/graphql.go:137-170`, generated `GraphQLAPIConfigData` in `pkg/api/management/generated.go:772-780`, and `graphql_api_test.go:602-662`.

**How to verify**
Create a GraphQL API with `upstream.sandbox.url` set, `GET` it (sandbox is present in the response), then deploy and inspect the stored deployment content: `SELECT content FROM deployments WHERE artifact_id = '<uuid>';` — the YAML `spec.upstream` has only `main:`. Compare with `transform/graphql.go:137`: `apiData.Upstream.Sandbox` is always nil for a CP-deployed API.

**Suggested fix**
```suggestion
	Main    *GraphQLUpstreamTarget `yaml:"main,omitempty"`
	Sandbox *GraphQLUpstreamTarget `yaml:"sandbox,omitempty"`
```
plus the matching population in `generateGraphQLAPIDeploymentYAML` and a correction to the type's doc comment. The alternative — rejecting `upstream.sandbox` at Create/Update with a validation error — is also defensible, but silently accepting and dropping it is not.

**Prompt for Claude**
```text
platform-api/internal/dto/graphql_api.go:57-59 defines GraphQLUpstream with only a
Main field, and platform-api/internal/service/graphql_deployment.go:98-108 only
populates Main -- so a GraphQL API's sandbox upstream never reaches the gateway.

1. FIRST confirm the sandbox is genuinely supported elsewhere: check that
   gateway/gateway-controller/pkg/transform/graphql.go:137-170 builds a sandbox
   route from apiData.Upstream.Sandbox, that GraphQLAPIConfigData.Upstream in
   gateway/gateway-controller/pkg/api/management/generated.go has a Sandbox field,
   and that platform-api/internal/service/graphql_api_test.go:602-662 asserts a
   stored sandbox upstream round-trips through the read response. If sandbox is
   NOT actually supported for GraphQL anywhere, STOP and tell me -- then the right
   fix is rejecting it at Create/Update instead.
2. If confirmed, add `Sandbox *GraphQLUpstreamTarget` with yaml tag
   "sandbox,omitempty" to dto.GraphQLUpstream, fix its doc comment (it currently
   claims "no sandbox split, unlike REST"), and populate it in
   generateGraphQLAPIDeploymentYAML from apiModel.Configuration.Upstream.Sandbox,
   carrying URL/Ref/Auth exactly as Main does.
3. Do not change the redaction behavior of mapUpstreamConfigToDTO -- the read
   response must keep redacting auth.value.
4. Confirm afterwards with a unit test in graphql_deployment_test.go asserting the
   marshalled YAML contains spec.upstream.sandbox.url for a model with a sandbox,
   and that gatewaytranslator.Translate still succeeds.
```

platform-api/resources/openapi.yaml
```diff
@@ -6182,6 +7031,394 @@ components:
...
+          type: string
+          format: uri
+          writeOnly: true
+          description: |
+            A URL to a raw SDL document to fetch and use as `sdl` — the write-side
+            counterpart to how an OpenAPI document can be supplied by reference for
+            other artifact kinds (see LlmProviderTemplate's `metadata.openapiSpecUrl`).
+            Distinct from `upstream.main.url`: this is a plain HTTP(S) GET of a static
+            schema file, not a live introspection query against a GraphQL server, and
+            is fetched with the same public-internet-only SSRF hardening as an
```
AnuGayan: **Severity:** 🟠 major
**Subject:** `sdlUrl` documents an SSRF control that the code does not implement

**Description**
This text tells operators `sdlUrl` "is fetched with the same public-internet-only SSRF hardening as an OpenAPI-spec-by-URL fetch (loopback/private/link-local/metadata addresses refused) — it is not meant for a tenant's own in-cluster backend." That is not what happens. `resolveSchema` calls `utils.FetchOpenAPISpecFromURL` (`internal/service/graphql_api.go:280`), which uses `NewUpstreamFetchClient(0)` (`internal/utils/openapi_spec_fetcher.go:78`) — the shared client built under `netguard.PermitPrivateBlockMetadata()`, whose own doc comment at `openapi_spec_fetcher.go:47-51` states it "permits private/loopback/in-cluster upstreams (a Kubernetes ClusterIP, a service-DNS name, localhost) and refuses only link-local/metadata/unspecified/multicast addresses." The stricter `isPublicIP` predicate in that same file is used only by `ValidateExternalURL` for LLM provider endpoints, and its comment says so explicitly. The same false claim appears in code at `internal/service/graphql_introspection.go:220-221`. The permissive behavior may well be the intended platform baseline — the defect is that a *new*, authenticated-user-supplied URL surface ships with published documentation asserting a control that isn't there, which is precisely what an operator or security reviewer would rely on. Confirmed by reading `openapi_spec_fetcher.go` in full and `guarded_http_client.go:41-51`.

**How to verify**
Read `platform-api/internal/utils/openapi_spec_fetcher.go:78` (it calls `NewUpstreamFetchClient(0)`, not a public-only client) against its own doc comment at lines 47-51, and note `isPublicIP` at line 134 has no caller inside `FetchOpenAPISpecFromURL`. Behaviorally: `POST /graphql-apis` with `sdlUrl: http://127.0.0.1:<port>/schema.graphql` succeeds against a locally-served SDL rather than being refused.

**Suggested fix**
Either correct the documentation to match the platform's actual (deliberate) policy, or make the code match the doc. Correcting the doc is the smaller change and keeps `sdlUrl` consistent with every other tenant-configured fetch: state that it is fetched through the shared SSRF-guarded client under the operator-configured policy (default `netguard.PermitPrivateBlockMetadata()`), where the host is resolved and each candidate IP — including every redirect hop — is checked at dial time, refusing link-local/metadata/unspecified/multicast addresses while private and in-cluster addresses remain reachable. If the public-only bar is actually wanted here, that is a code change (validate the resolved host with `isPublicIP` before the fetch) and should be stated as such — pick one, but do not ship the mismatch.

**Prompt for Claude**
```text
platform-api/resources/openapi.yaml around line 7157 documents the GraphQL `sdlUrl`
field as "fetched with the same public-internet-only SSRF hardening ...
(loopback/private/link-local/metadata addresses refused)".

1. FIRST verify which side is wrong. Read
   platform-api/internal/utils/openapi_spec_fetcher.go end to end: confirm
   FetchOpenAPISpecFromURL calls NewUpstreamFetchClient(0), that its doc comment
   (lines ~47-51) says the policy is netguard.PermitPrivateBlockMetadata()
   permitting private/loopback, and that isPublicIP (line ~134) is NOT called from
   FetchOpenAPISpecFromURL. Also read platform-api/internal/utils/
   guarded_http_client.go:41-51. If the fetch DOES enforce a public-only policy,
   STOP and tell me -- the finding is wrong.
2. If confirmed, correct the DOCUMENTATION, not the behavior -- rewrite the sdlUrl
   description so it states the real policy (shared SSRF-guarded client, per-hop
   dial-time IP checks, link-local/metadata refused, private/in-cluster permitted).
   Also fix the same false claim in the comment at
   platform-api/internal/service/graphql_introspection.go:219-221, which asserts
   FetchOpenAPISpecFromURL uses a "stricter public-only policy".
3. Do NOT tighten the fetch to public-only without asking -- that would change
   behavior for existing LlmProviderTemplate openapiSpecUrl callers too.
4. Confirm afterwards that no other spec text or comment in this PR claims a
   public-only bar for sdlUrl: grep -rn "public-only\|public-internet-only" over
   the changed files.
```

platform-api/internal/service/graphql_introspection.go
```diff
@@ -0,0 +1,428 @@
...
+	httpReq.Header.Set("Accept", "application/json")
+
+	// The GraphQL endpoint URL is tenant-supplied — dial through the
+	// SSRF-guarded client (ssrf-prevention.md directive 6: reuse the shared
+	// upstream-fetch helper rather than a one-off client). upstream.main.url
+	// is the tenant's own configured backend (analogous to REST/MCP's
+	// upstream), so NewUpstreamFetchClient's private/in-cluster-permitting
+	// policy is the correct one here — not the stricter public-only policy
+	// FetchOpenAPISpecFromURL uses for fetching a public vendor's OpenAPI doc.
+	client, err := utils.NewUpstreamFetchClient(graphQLIntrospectionTimeout)
```
AnuGayan, line 233: **Severity:** 🟡 minor
**Subject:** Introspection call has no per-call deadline and drops the request context

**Description**
`graphQLIntrospectionTimeout` is documented at lines 36-40 as bounding "the outbound introspection call end to end" and passed here — where it is discarded (`platform-api/internal/utils/guarded_http_client.go:98`: `_ = timeout`). The request is built with `http.NewRequest` at line 208, so there is no context either. Every peer enforces its real budget via the context instead: `CheckURLReachability` (`internal/utils/common.go:483-491`) and `FetchOpenAPISpecFromURL` (`internal/utils/openapi_spec_fetcher.go:75-84`) both do `context.WithTimeout` + `http.NewRequestWithContext`. Net effect: the introspection is bounded only by the shared client's safety-net `Timeouts.Overall`, which `platform-api/config/config-template.toml:390` sets to `60s` (httpkit's built-in default is `30s`) — 4× the documented budget — and it keeps running after the API client disconnects. Same pattern at `graphql_api.go:280`, which passes `context.Background()` rather than the request context. Confirmed against all four files.

**How to verify**
```bash
grep -n "_ = timeout" platform-api/internal/utils/guarded_http_client.go   # line 98: argument is dead
grep -n "overall" platform-api/config/config-template.toml                # 60s
```
Point `upstream.main.url` at a host that accepts the connection and never responds: the request hangs ~60s, not 15s, and cancelling the client request does not stop it.

**Suggested fix**
```suggestion
	client, err := utils.NewUpstreamFetchClient(graphQLIntrospectionTimeout)
	if err != nil {
		return "", fmt.Errorf("failed to create HTTP client: %w", err)
	}
	// The shared client's Timeouts.Overall is a safety net only (see
	// InitSharedHTTPClient) -- enforce this call's real budget via the context,
	// matching CheckURLReachability and FetchOpenAPISpecFromURL.
	ctx, cancel := context.WithTimeout(context.Background(), graphQLIntrospectionTimeout)
	defer cancel()
	httpReq = httpReq.WithContext(ctx)
```
Better still, thread a `ctx context.Context` through `resolveSchema` → `fetchAndConvertGraphQLSchema` from the handler's `r.Context()`, and use it for the `FetchOpenAPISpecFromURL` call at `graphql_api.go:280` too.

**Prompt for Claude**
```text
platform-api/internal/service/graphql_introspection.go:202-226 documents a 15s
introspection budget but does not enforce one.

1. FIRST confirm: check that platform-api/internal/utils/guarded_http_client.go:98
   discards the timeout argument (`_ = timeout`), that graphql_introspection.go:208
   uses http.NewRequest (no context), and that peers
   platform-api/internal/utils/common.go:483-491 and
   platform-api/internal/utils/openapi_spec_fetcher.go:75-84 both use
   context.WithTimeout + http.NewRequestWithContext. If the timeout is actually
   honored somewhere, STOP and say so.
2. If confirmed, add a `ctx context.Context` parameter to
   fetchAndConvertGraphQLSchema and to GraphQLAPIService.resolveSchema, pass
   r.Context() down from the Create/Update handlers, and inside
   fetchAndConvertGraphQLSchema do context.WithTimeout(ctx,
   graphQLIntrospectionTimeout) + http.NewRequestWithContext. Also replace the
   context.Background() at graphql_api.go:280 with the threaded ctx.
3. Do not change the 5 MiB response ceiling, the sterile-error mapping to
   GraphQLAPISchemaResolveFailed, or the choice of NewUpstreamFetchClient.
4. Confirm afterwards: go test ./platform-api/internal/service/... stays green, and
   a test with a hanging httptest server fails in ~15s rather than ~60s.
```

gateway/gateway-controller/pkg/utils/graphql_deployment.go
```diff
@@ -0,0 +1,152 @@
...
+// same generated api.Upstream type), so this intentionally mirrors
+// config.APIValidator's private validateUpstreamUrl in miniature: GraphQLApi does not
+// support upstreamDefinitions references in this pass, so only a direct url is valid.
+func validateGraphQLUpstream(label string, up *api.Upstream) []config.ValidationError {
+	var errors []config.ValidationError
+	if up == nil {
+		return errors
+	}
+
+	if up.Url == nil || strings.TrimSpace(*up.Url) == "" {
```
AnuGayan, line 128: **Severity:** 🟡 minor
**Subject:** `validateGraphQLUpstream` does not reject `upstream.ref`, though the spec says it is rejected

**Description**
The generated spec text for `GraphQLAPIConfigData.Upstream` states: "Only a direct inline url is supported — … `upstream.ref` … cannot be resolved and **is rejected**." This function only checks that `url` is non-empty; it never inspects `Ref`. So `ref` alone produces the misleading error "Upstream URL is required" (nothing tells the user `ref` is unsupported), and `url` + `ref` together is accepted with `ref` silently ignored, since `resolveUpstreamURL` (`gateway/gateway-controller/pkg/transform/restapi.go:587-588`) returns early on `Url`. Confirmed against this file's lines 115-127, `restapi.go:586-609`, and the spec description in `pkg/api/management/generated.go:772`.

**How to verify**
`POST /graphql-apis` with `spec.upstream.main: {url: "http://x:80/graphql", ref: "some-def"}` — accepted, deploys against the `url`, with no mention of the ignored `ref`.

**Suggested fix**
```suggestion
	if up.Ref != nil && strings.TrimSpace(*up.Ref) != "" {
		errors = append(errors, config.ValidationError{
			Field:   "spec.upstream." + label + ".ref",
			Message: "Upstream ref is not supported for GraphQLApi (no upstreamDefinitions list); use a direct url",
		})
	}

	if up.Url == nil || strings.TrimSpace(*up.Url) == "" {
```

**Prompt for Claude**
```text
gateway/gateway-controller/pkg/utils/graphql_deployment.go:115-127
(validateGraphQLUpstream) never inspects up.Ref, but the GraphQLAPIConfigData
upstream description in the management OpenAPI spec says upstream.ref "is rejected".

1. FIRST confirm: read validateGraphQLUpstream and check it only tests up.Url;
   check resolveUpstreamURL in gateway/gateway-controller/pkg/transform/
   restapi.go:586-609 returns early on a non-empty Url so Ref is silently ignored;
   and confirm the spec text in gateway/gateway-controller/api/
   management-openapi.yaml says ref is rejected. If validation already rejects ref,
   STOP and tell me.
2. If confirmed, add an explicit ValidationError for a non-empty up.Ref, before the
   existing "Upstream URL is required" check, naming field
   "spec.upstream.<label>.ref".
3. Do not change validation for RestApi -- config.APIValidator's own
   validateUpstreamUrl must keep supporting refs.
4. Confirm afterwards: go test ./gateway/gateway-controller/... stays green, and add
   a case to the graphql config-validator tests asserting a ref-only and a
   url+ref upstream both produce the new error.
```

gateway/it/features/graphql_deploy.feature
```diff
@@ -0,0 +1,862 @@
...
+            """
+        Then the response should be successful
+        And the response header "X-GraphQL-Test-Marker" should be "graphql-policy-works"
+
+        # Cleanup
+        Given I authenticate using basic auth as "admin"
+        When I delete the GraphQL API "set-headers-graphql-v1"
+        Then the response should be successful
+
+    Scenario: GraphQL API with cors does not handle a preflight request - confirmed limitation, not yet supported
```
AnuGayan, line 707: **Severity:** 🟡 minor
**Subject:** CORS preflight limitation is documented only in this feature file

**Description**
The GraphQL route is a single `Exact`-match `POST`, so an `OPTIONS` preflight never matches and the `cors` policy never runs. Calling this out here is good — but it is the *only* place it is written down. Neither `gateway/gateway-controller/api/management-openapi.yaml` nor `platform-api/resources/openapi.yaml` mentions it (`grep -i preflight` finds nothing in either). A user attaching a `cors` policy to a GraphQL API — which both specs advertise as supported, and which this suite tests — gets a policy that silently does nothing for browser clients. Confirmed by grepping both specs and reading this scenario's own explanation at lines 737-740.

**How to verify**
```bash
grep -ri preflight gateway/gateway-controller/api/management-openapi.yaml platform-api/resources/openapi.yaml   # no output
grep -n preflight gateway/it/features/graphql_deploy.feature                                                    # line 706
```

**Suggested fix**
Add a sentence to the `policies` description of `GraphQLAPIConfigData` (gateway spec) and of `GraphQLAPI` (platform-api spec): a `cors` policy applies to the `POST` route only; browser preflight (`OPTIONS`) is not currently routed for a GraphQL API, so cross-origin browser clients that trigger a preflight will fail. No code change.

**Prompt for Claude**
```text
The CORS-preflight limitation for GraphQL APIs is documented only in
gateway/it/features/graphql_deploy.feature:706-740, not in either OpenAPI spec.

1. FIRST confirm: `grep -ri preflight` over
   gateway/gateway-controller/api/management-openapi.yaml and
   platform-api/resources/openapi.yaml returns nothing, and that
   gateway/gateway-controller/pkg/transform/graphql.go builds only a POST route
   with PathMatchType "Exact". If an OPTIONS route is built anywhere, STOP.
2. If confirmed, add one or two sentences to the `policies` field description of
   GraphQLAPIConfigData in the gateway management spec and of GraphQLAPI in
   platform-api/resources/openapi.yaml, stating that a cors policy applies to the
   single POST route and that OPTIONS preflight is not routed for a GraphQLApi.
3. Documentation only -- do not add an OPTIONS route or change the transformer.
4. Regenerate the server code for both specs the same way this PR did, and confirm
   go build ./gateway/gateway-controller/... ./platform-api/... stays clean.
```

platform-api/internal/service/graphql_api.go
```diff
@@ -0,0 +1,729 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	var subscriptionPlans []string
+	if req.SubscriptionPlans != nil {
+		subscriptionPlans = *req.SubscriptionPlans
+	}
+
+	context := req.Context
```
AnuGayan, line 226: **Severity:** 🔵 nit
**Subject:** `gofmt` fails on three files in this PR

**Description**
`gofmt -l` flags three files: this one (struct field misalignment at lines 226-234), `platform-api/internal/service/graphql_introspection.go` (struct tag misalignment at lines 157-166), and `cli/src/cmd/gateway/apply.go` (a stray trailing blank line added at EOF). No repo-wide gofmt/golangci gate exists for these modules — only `kubernetes/gateway-operator/.golangci.yml` — so CI won't catch it, which is why this is a nit rather than a blocker. It does contradict the PR description's "go build/go vet/go test ./... clean" claim. Confirmed by running `gofmt -l` and `gofmt -d` on the changed Go files (`go build` and `go vet` are genuinely clean on both modules — I verified those).

**How to verify**
```bash
gofmt -l ./cli/src ./platform-api ./gateway/gateway-controller
```

**Suggested fix**
```bash
gofmt -w platform-api/internal/service/graphql_api.go \
         platform-api/internal/service/graphql_introspection.go \
         cli/src/cmd/gateway/apply.go
```

**Prompt for Claude**
```text
gofmt reports three files in this PR as unformatted.

1. FIRST confirm by running:
   gofmt -l ./cli/src ./platform-api ./gateway/gateway-controller
   Expect platform-api/internal/service/graphql_api.go,
   platform-api/internal/service/graphql_introspection.go and
   cli/src/cmd/gateway/apply.go. If the list is empty, STOP -- already fixed.
2. Run `gofmt -w` on exactly the files it lists. Review the resulting diff and
   confirm it is whitespace/alignment only (plus removal of the trailing blank
   line at the end of cli/src/cmd/gateway/apply.go) -- no semantic change.
3. Do not reformat files this PR did not touch.
4. Confirm afterwards: gofmt -l on those paths is empty, and
   go build ./gateway/gateway-controller/... ./platform-api/... is still clean.
```

platform-api/internal/utils/graphql_multipart.go
```diff
@@ -0,0 +1,113 @@
...
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+)
+
+const (
+	// maxGraphQLSDLUploadBytes bounds an uploaded SDL file — mirrors the 5 MiB
+	// ceiling the CLI's standalone-gateway sdlFile path already uses
+	// (cli/src/cmd/gateway/apply.go's maxGraphQLSDLFileBytes), so the limit is
```
AnuGayan: **Severity:** 🔵 nit
**Subject:** Comment cites `maxGraphQLSDLFileBytes`, an identifier that does not exist

**Description**
This comment justifies the 5 MiB ceiling as mirroring "the CLI's standalone-gateway sdlFile path (`cli/src/cmd/gateway/apply.go`'s `maxGraphQLSDLFileBytes`)". Neither that constant nor any `sdlFile` handling exists anywhere in `cli/` — consistent with this PR's CLI diff, which adds no SDL-upload path. A future maintainer will go looking for a precedent that was never written. Confirmed via `grep -rn "maxGraphQLSDLFileBytes\|sdlFile" cli/`, which returns nothing.

**How to verify**
```bash
grep -rn "maxGraphQLSDLFileBytes\|sdlFile" cli/   # no output
```

**Suggested fix**
```suggestion
	// maxGraphQLSDLUploadBytes bounds an uploaded SDL file at 5 MiB, matching
	// defaultOpenAPISpecMaxFetchBytes (openapi_spec_fetcher.go) so the ceiling is
	// the same whether the schema arrives as a file upload or via sdlUrl.
	maxGraphQLSDLUploadBytes = 5 << 20
```

**Prompt for Claude**
```text
platform-api/internal/utils/graphql_multipart.go:29-33 cites
"cli/src/cmd/gateway/apply.go's maxGraphQLSDLFileBytes" as the precedent for its
5 MiB limit.

1. FIRST confirm the identifier does not exist:
   grep -rn "maxGraphQLSDLFileBytes\|sdlFile" cli/
   If it returns a definition, STOP -- the comment is correct.
2. If confirmed absent, rewrite the comment to cite a real precedent
   (defaultOpenAPISpecMaxFetchBytes in
   platform-api/internal/utils/openapi_spec_fetcher.go is the same 5 MiB) or to
   simply state the limit without a cross-reference. Do NOT change the constant's
   value.
3. Leave maxGraphQLMultipartRequestBytes and all parsing logic untouched.
4. Confirm afterwards: go build ./platform-api/... is clean and
   go test ./platform-api/internal/utils/... still passes.
```

gateway/gateway-controller/pkg/transform/graphql.go
```diff
@@ -0,0 +1,204 @@
...
+			Path:            fullPath,
+			PathMatchType:   "Exact",
+			Vhost:           sandboxVhost,
+			AutoHostRewrite: sbAutoHostRewrite,
+			Upstream: models.RouteUpstream{
+				ClusterKey: sbUpstream.ClusterKey,
+				Default:    &sbUpstreamInfo,
+			},
+		}
+		rdc.PolicyChains[sandboxRouteKey] = policyChain
```
AnuGayan: **Severity:** 🔵 nit
**Subject:** Main and sandbox routes share one `*models.PolicyChain` pointer

**Description**
The chain is built once at line 97 and the same pointer is assigned to both route keys (line 133 and here). `RestAPITransformer` instead calls `sdkChainToModel` freshly per route key (`restapi.go:235`). Nothing mutates a chain today — `policyxds/snapshot.go:323-324` only ranges over `chain.Policies`, and `RuntimeDeployConfig.ValidateResolution` only reads — so this is latent rather than broken. But the aliasing means any future per-slot policy adjustment (a sandbox-specific rate limit, say) would silently affect both vhosts. Building the chain twice costs nothing. Confirmed by grepping every consumer of `PolicyChains` in `gateway/gateway-controller/pkg/` for mutation.

**How to verify**
Read line 97 against lines 133 and 169: one `sdkChainToModel` result, two map entries. Compare with `restapi.go:232-235`, which builds a fresh chain inside the per-vhost loop.

**Suggested fix**
```suggestion
		rdc.PolicyChains[sandboxRouteKey] = sdkChainToModel(injected)
```
(and correspondingly at line 133), dropping the shared `policyChain` local.

**Prompt for Claude**
```text
gateway/gateway-controller/pkg/transform/graphql.go builds one *models.PolicyChain
at line 97 and assigns the same pointer to both the main route key (line 133) and
the sandbox route key (line 169).

1. FIRST confirm nothing mutates a chain today: grep every use of PolicyChains and
   of chain.Policies under gateway/gateway-controller/pkg/ (expect only reads --
   policyxds/snapshot.go ranges over it, models/runtime_deploy_config.go only
   validates). If any consumer mutates a chain, say so -- the severity is higher
   than a nit.
2. Whether or not a mutator exists, remove the aliasing: call
   sdkChainToModel(injected) separately for each route key so main and sandbox each
   own their chain, matching RestAPITransformer (restapi.go:235). Keep the single
   collectAPIPolicies/buildPolicyChain/InjectSystemPolicies computation -- only the
   model conversion is duplicated.
3. Do not change the route keys, the policy ordering, or the injected system
   policies.
4. Confirm afterwards: go test ./gateway/gateway-controller/pkg/transform/... is
   green and graphql_test.go's expectations on both route keys still hold.
```

event-gateway/gateway-controller/cmd/controller/main.go
```diff
@@ -430,14 +430,23 @@ func main() {
 	policyVersionResolver := utils.NewLoadedPolicyVersionResolver(policyDefinitions)
 	restTransformer := transform.NewRestAPITransformer(&cfg.Router, cfg, policyDefinitions)
 	llmTransformer := transform.NewLLMTransformer(configStore, db, &cfg.Router, cfg, policyDefinitions, policyVersionResolver)
-	transformerRegistry := transform.NewRegistry(restTransformer, llmTransformer)
+	// GraphQLApi's config validator/deploy parser (pkg/utils/graphql_deployment.go)
+	// self-register via init() and are therefore already active in this binary too
+	// (transitively imported via the shared transform/handlers packages) — the
+	// /graphql-apis CRUD and api-key routes are already reachable here via the
```
AnuGayan: **Severity:** 🔵 nit
**Subject:** Comment overstates that the `/graphql-apis` routes are reachable in this binary

**Description**
This comment says the `/graphql-apis` CRUD and api-key routes "are already reachable here via the shared `*handlers.APIServer`". They are *routable* — this binary does mount the shared `handlers.NewAPIServer` — but this binary's own `generateAuthConfig` (line 741) lists only `websub-*` and `webbroker-*` routes, and `common/authenticators/authz.go:65-69` returns 403 for a matched-but-unlisted route. So with auth enabled they are reachable only in the sense that they 403. This is pre-existing rather than introduced here — `/rest-apis` is in exactly the same position in this binary — so wiring the transformer (which this diff correctly does) is the right change; it is only the comment that reads as a guarantee it isn't. Confirmed against `event-gateway/gateway-controller/cmd/controller/main.go:741-780` and `common/authenticators/authz.go:65-69`.

**How to verify**
```bash
grep -n "graphql-apis\|rest-apis" event-gateway/gateway-controller/cmd/controller/main.go   # neither appears in relativeRoles
```
Then read `common/authenticators/authz.go:66-69`: `if !found { ... StatusForbidden }`.

**Suggested fix**
Soften the claim, e.g. "…are served here by the shared `*handlers.APIServer` (they are not listed in this binary's `generateAuthConfig` role map, the same as `/rest-apis`, so they 403 when auth is enabled — but the transformer must still be wired for the immutable-gateway and DP-local paths)." No code change.

**Prompt for Claude**
```text
event-gateway/gateway-controller/cmd/controller/main.go:433-439 contains a comment
claiming the /graphql-apis CRUD and api-key routes are "already reachable here via
the shared *handlers.APIServer".

1. FIRST confirm: check generateAuthConfig in the same file (line ~741) -- its
   relativeRoles map should contain only websub-*/webbroker-* entries, with no
   /graphql-apis and no /rest-apis. Then check common/authenticators/authz.go:65-69
   returns 403 for a matched route absent from ResourceRoles. If /graphql-apis IS
   in that map, STOP -- the comment is fine.
2. If confirmed, reword the comment only, to say the routes are served by the
   shared APIServer but are not in this binary's auth role map (same as
   /rest-apis) and therefore 403 when auth is enabled.
3. Do NOT add /graphql-apis entries to this binary's relativeRoles map as part of
   this PR -- that would change the event-gateway's auth surface and should be a
   deliberate, separate decision alongside /rest-apis.
4. Confirm afterwards: go build ./event-gateway/... is clean.
```

platform-api/internal/handler/graphql_api.go
```diff
@@ -0,0 +1,349 @@
...
+		return apperror.Unauthorized.New().
+			WithLogMessage("organization claim not found in token")
+	}
+
+	projectId := strings.TrimSpace(r.URL.Query().Get("projectId"))
+	if projectId == "" {
+		return apperror.ValidationFailed.New("projectId query parameter is required")
+	}
+
+	limit, offset := parsePagination(r)
```
AnuGayan: **Severity:** 🟠 major
**Subject:** `GET /graphql-apis` declares `sortBy`, `sortOrder` and `query`, and silently ignores all three

**Description**
The spec declares six query parameters for this endpoint — `projectId-Q`, `limit-Q`, `offset-Q`, `sortBy-Q`, `sortOrder-Q`, `query-Q` (`platform-api/resources/openapi.yaml`, `/graphql-apis` → `get` → `parameters`). This handler reads only `projectId` and `parsePagination(r)`, so `sortBy`, `sortOrder` and `query` are dropped. `GraphQLAPIService.List` takes no sort/filter arguments, and `repository/graphql_api.go:144` hardcodes `ORDER BY created_at DESC` with no search predicate. The infrastructure already exists and REST uses it: `parseListOptions` (`internal/handler/pagination.go:63`) returns a `repository.ListOptions{Limit, Offset, SortBy, SortOrder, Search}`, and `internal/handler/api.go:135` calls it. A client passing `?query=foo` receives a `200` with **unfiltered** results plus correct-looking pagination metadata, so it will page through the whole collection believing it is a filtered set — wrong answers, not a visible error. Confirmed against the spec's parameter list, this handler, `pagination.go:63-88`, `api.go:130-150`, and `repository/graphql_api.go:132-174`.

**How to verify**
```bash
curl "$CP/api/v0.9/graphql-apis?projectId=default-project&query=zzz-no-match&sortBy=name&sortOrder=asc"
```
Returns every GraphQL API in the project, in `created_at DESC` order. Compare with the same query against `/rest-apis`, which honors all three.

**Suggested fix**
```suggestion
	opts := parseListOptions(r)
```
then thread `repository.ListOptions` through `GraphQLAPIService.List` and into the repository (resolving `SortBy` against a column allowlist and applying `Search` as a case-insensitive `handle` `LIKE`), exactly as `APIService.GetAPIsByOrganization` does. If honoring them is out of scope for this PR, the honest alternative is removing the three `$ref`s from the spec so the contract matches the implementation — but shipping them as documented-and-ignored is the one option that misleads.

**Prompt for Claude**
```text
platform-api/internal/handler/graphql_api.go:157 (ListGraphQLAPIs) uses
parsePagination(r), so the sortBy/sortOrder/query params that the OpenAPI spec
declares for GET /graphql-apis are silently ignored.

1. FIRST confirm: check that the /graphql-apis get.parameters list in
   platform-api/resources/openapi.yaml includes sortBy-Q, sortOrder-Q and query-Q;
   that internal/handler/pagination.go defines parseListOptions returning
   repository.ListOptions{Limit,Offset,SortBy,SortOrder,Search}; that
   internal/handler/api.go:135 (ListRESTAPIs) calls parseListOptions; and that
   repository/graphql_api.go List() hardcodes ORDER BY created_at DESC with no
   search predicate. If GraphQL already honors these params, STOP and tell me.
2. If confirmed, switch ListGraphQLAPIs to parseListOptions(r) and thread
   repository.ListOptions through GraphQLAPIService.List and
   GraphQLAPIRepo.List/CountByProject/Count. Resolve SortBy against an explicit
   column allowlist (never interpolate it into SQL) and apply Search as a
   case-insensitive substring match on handle -- copy the approach REST already
   uses so the two behave identically.
3. Keep the projectId-required behavior, the existing parsePagination clamping, and
   the total-count semantics unchanged. Do not interpolate any user value into SQL.
4. Confirm afterwards: a request with query=<no-match> returns an empty list with
   total 0, sortBy=name&sortOrder=asc changes the ordering, and
   go test ./platform-api/... stays green. Add a repository test covering the
   allowlist fallback for an unrecognized sortBy.
```

platform-api/api/generated.go
```diff
@@ -2256,7 +2519,7 @@ type SecretCreateRequest struct {
 	Type *SecretCreateRequestType `json:"type,omitempty" yaml:"type,omitempty"`
 
 	// Value Plaintext secret value — encrypted at rest, never returned in any response
-	Value string `binding:"required" json:"value" yaml:"value"`
+	Value *string `binding:"required" json:"value,omitempty" yaml:"value,omitempty"`
```
AnuGayan, line 2548: **Severity:** 🟡 minor
**Subject:** This regeneration reconciles pre-existing spec drift, so unrelated public-type changes ride along in the diff

**Description**
`SecretCreateRequest.Value` (and `SecretUpdateRequest.Value` at line 2590) change from `string` to `*string`, and `json:"value"` to `json:"value,omitempty"` — a public type change in a GraphQL PR. I traced the cause rather than assuming: the generator is **pinned at v2.5.1** (`platform-api/Makefile:159`) and the file header reports v2.5.1 both before and after, so this is not a generator bump; and this PR makes exactly **one** removal in `openapi.yaml` (a tags description), touching neither the Secret schemas nor the MCP fetch request. Meanwhile `openapi.yaml:8821` already reads "exposed by this **provider**" where the committed `generated.go` said "proxy". So `generated.go` on `main` was simply out of sync with the spec, and this PR is the first to regenerate it. Also swept in: `MCPServerInfoFetchRequest` becoming a union with a new `MCPServerInfoFetchRequest1 = interface{}` (doc comment renders as "defines model for ."), and assorted doc-comment drift. Functional impact is nil — `SecretCreateRequest`/`SecretUpdateRequest` are referenced nowhere outside `generated.go` (the handlers use a hand-written `dto.CreateSecretRequest` parsed from multipart form values), which is why the build stays clean. The cost is attribution: a secrets regression bisected later lands on the GraphQL PR. Confirmed against the Makefile, the file header, the full openapi.yaml diff, and a repo-wide grep for both type names.

Separately, and this part **is** caused by this PR: because `/graphql-apis` reuses `sortBy-Q`/`sortOrder-Q`, oapi-codegen had to disambiguate, renaming `CreatedAt`→`ListRESTAPIsParamsSortByCreatedAt`, `Name`→`…SortByName`, `Asc`/`Desc`→`…SortOrderAsc`/`Desc`, and all six `GetDeploymentsParamsStatus*` constants. Unavoidable and harmless in-repo, but worth knowing it is an exported-symbol rename in a shared generated package.

**How to verify**
```bash
git log -1 --format=%h -- platform-api/resources/openapi.yaml
grep -n "exposed by this provider" platform-api/resources/openapi.yaml   # spec says "provider"
git show HEAD~1:platform-api/api/generated.go | grep -c "exposed by this proxy"  # generated said "proxy"
grep -rn "SecretCreateRequest" platform-api sdk --include='*.go' | grep -v api/generated.go   # no consumers
```
Or on `main` before this branch: run `make generate` (or the `oapi-codegen@v2.5.1` line from `Makefile:159`) and observe a dirty `generated.go` with no source change.

**Suggested fix**
No code change needed. Land the pure-regeneration delta as its own commit (or a small precursor PR) so this PR's generated diff contains only GraphQL additions, and note the `Value` type change in that commit message. At minimum, call it out in this PR's description so it is discoverable later. Adding a CI check that `make generate` leaves the tree clean would stop the drift recurring — that is the durable fix, but out of scope here.

**Prompt for Claude**
```text
platform-api/api/generated.go was stale relative to resources/openapi.yaml before
this PR, so regenerating it here sweeps unrelated public-type changes into a
GraphQL diff.

1. FIRST confirm the drift is pre-existing, not caused by this PR:
   - check platform-api/Makefile pins oapi-codegen at v2.5.1 and that the
     generated.go header says v2.5.1 both before and after this PR;
   - confirm the openapi.yaml diff in this PR has exactly one removed line (a tags
     description) and does not touch SecretCreateRequest, SecretUpdateRequest or
     MCPServerInfoFetchRequest;
   - confirm resources/openapi.yaml says "exposed by this provider" while the
     pre-PR generated.go said "exposed by this proxy".
   If the PR DID edit those schemas, STOP -- the finding is wrong.
2. If confirmed, do NOT hand-edit generated.go. Instead split the diff: create a
   commit containing only the regeneration of generated.go against the unchanged
   base spec, then rebase the GraphQL spec+codegen changes on top, so each commit
   is separately reviewable. Note in the regeneration commit message that
   SecretCreateRequest.Value and SecretUpdateRequest.Value become *string with
   omitempty (a consequence of the spec's existing writeOnly: true).
3. Verify nothing depends on the changed types before splitting: grep for
   SecretCreateRequest and SecretUpdateRequest across platform-api and sdk and
   confirm generated.go is the only hit.
4. Confirm afterwards: go build ./platform-api/... is clean on each commit
   independently, and re-running the Makefile codegen leaves the tree unchanged.
```

platform-api/go.mod
```diff
@@ -17,6 +17,7 @@ require (
 	github.com/microsoft/go-mssqldb v1.10.0
 	github.com/oapi-codegen/runtime v1.5.0
 	github.com/stretchr/testify v1.11.1
+	github.com/vektah/gqlparser/v2 v2.5.36
```
AnuGayan, line 20: **Severity:** 🟡 minor
**Subject:** New dependency not added to the third-party license report

**Description**
This adds `github.com/vektah/gqlparser/v2 v2.5.36` as a direct requirement, pulling three new indirect modules (`agnivade/levenshtein v1.2.1`, `arbovm/levenshtein`, `dgryski/trifles`). `license-reports/go/platform-api-third-party-go-licenses.csv` is not regenerated — `grep -rln gqlparser license-reports/` returns nothing, and `license-reports/` appears nowhere in this PR's changed files. That CSV demonstrably tracks indirect dependencies too: `github.com/apapsch/go-jsonmerge/v2` is marked `// indirect` in this same `go.mod` and is listed among the CSV's 40 rows. There is a dedicated `make go-license-report` target (`Makefile:228` → `scripts/generate-go-license-report.sh`), so keeping the report in sync is an established repo convention — and a WSO2 release-process requirement rather than a nicety. Confirmed against the CSV contents, `go.mod`, the `go.sum` additions in this diff, and the Makefile target.

**How to verify**
```bash
grep -rln gqlparser license-reports/                  # no output
grep -c "" license-reports/go/platform-api-third-party-go-licenses.csv   # 40 rows, none of them gqlparser
grep -n "go-jsonmerge" platform-api/go.mod            # "// indirect", yet present in the CSV
```

**Suggested fix**
```bash
make go-license-report
```
and commit the updated `license-reports/go/*.csv`. Worth eyeballing the result: `gqlparser` is MIT, but confirm the report picks up the three new indirect modules and that none carries a license the release process disallows.

**Prompt for Claude**
```text
This PR adds github.com/vektah/gqlparser/v2 v2.5.36 (direct) plus three indirect
modules to platform-api, without regenerating the third-party license report.

1. FIRST confirm: `grep -rln gqlparser license-reports/` returns nothing; and
   confirm the report does cover indirect deps by checking that
   github.com/apapsch/go-jsonmerge/v2 is marked "// indirect" in
   platform-api/go.mod yet appears in
   license-reports/go/platform-api-third-party-go-licenses.csv. If gqlparser is
   already listed, STOP -- the finding is wrong.
2. If confirmed, run `make go-license-report` (Makefile:228 ->
   scripts/generate-go-license-report.sh) and commit only the resulting changes
   under license-reports/. Do not hand-edit the CSVs.
3. Inspect the diff: confirm gqlparser/v2 is listed with its real license (expected
   MIT) and report back which of agnivade/levenshtein, arbovm/levenshtein and
   dgryski/trifles the generator picked up, with their licenses, so a human can
   confirm none is disallowed.
4. Do not change go.mod or go.sum, and do not attempt to remove the dependency --
   gqlparser is what validateGraphQLSDL and the introspection-to-SDL converter are
   built on.
```

gateway/it/steps_graphql.go
```diff
@@ -0,0 +1,88 @@
...
+
+	"github.com/cucumber/godog"
+	"github.com/wso2/api-platform/gateway/it/steps"
+)
+
+// RegisterGraphQLSteps registers all GraphQL API deployment step definitions.
+// Mirrors RegisterAPISteps (RestApi) / RegisterMCPSteps (Mcp) — GraphQLApi is a
+// core kind on the gateway-controller with the same generic
+// create/list/get/update/delete surface at /graphql-apis, just with no
+// per-operation routes (docs/specs/graphql-api-support.md §6.1).
```
AnuGayan: **Severity:** 🟡 minor
**Subject:** `docs/specs/graphql-api-support.md` is cited five times with section numbers but does not exist

**Description**
This comment cites `docs/specs/graphql-api-support.md §6.1` as the authority for the single-route design. That file does not exist anywhere in the repo, and neither does any GraphQL doc: `ls docs/specs/graphql-api-support.md` fails and `find docs -iname '*graphql*'` returns nothing. The PR description says "Documentation: N/A for now." Five references point at it, four with specific section numbers:

| Location | Cites |
|---|---|
| `gateway/it/steps_graphql.go:34` | §6.1 |
| `gateway/it/features/graphql-api-keys.feature:21` | (file) |
| `gateway/it/features/graphql_deploy.feature:118` | §6.1/§6.2 |
| `gateway/it/features/graphql_deploy.feature:736` | §6.1 |
| `gateway/it/features/graphql_deploy.feature:743` | §7's QoS table |

This compounds the CORS-preflight finding I filed on `graphql_deploy.feature:706`: that scenario says the limitation is "tracked in docs/specs/graphql-api-support.md §7's QoS table", so the only claimed record of a user-visible limitation is a document that was never written. Confirmed by `ls`, `find`, and a repo-wide grep for `graphql-api-support`.

**How to verify**
```bash
ls docs/specs/graphql-api-support.md            # No such file or directory
find docs -iname '*graphql*'                    # no output
grep -rn "graphql-api-support" --include='*.go' --include='*.feature' .   # 5 hits
```

**Suggested fix**
Either add the doc (it is referenced as the design record for the single-route decision and the QoS/limitations table, so it would carry real weight), or drop the section-numbered citations and inline the one-line rationale each site actually needs. A dangling reference with a section number is worse than no reference — it reads as verifiable and is not.

**Prompt for Claude**
```text
Five comments across gateway/it/steps_graphql.go and
gateway/it/features/graphql*.feature cite docs/specs/graphql-api-support.md
(with section numbers) but that file does not exist.

1. FIRST confirm: `ls docs/specs/graphql-api-support.md`,
   `find docs -iname '*graphql*'`, and
   `grep -rn "graphql-api-support" --include='*.go' --include='*.feature' .`
   Expect the file to be absent and exactly 5 citing sites. If the doc exists on
   another branch or path, STOP and tell me where -- then the fix is correcting the
   path, not removing the citations.
2. If confirmed absent, ask the author which they want before editing: (a) write
   docs/specs/graphql-api-support.md, or (b) remove the citations. Do not silently
   delete references to a design record.
3. If (b) is chosen: at each of the 5 sites, replace the citation with the concrete
   one-line rationale that site depends on -- in particular
   graphql_deploy.feature:743 must keep stating that the cors preflight gap is a
   current, known limitation, since that sentence is the only place it is recorded.
4. Change comments only -- do not alter any Given/When/Then step, assertion, or
   Go step definition. Confirm afterwards that the integration suite still compiles
   (go vet ./gateway/it/...).
```

platform-api/internal/handler/graphql_api.go
```diff
@@ -0,0 +1,349 @@
...
+
+	deletedBy, err := resolveActorErr(r, h.identity, "delete GraphQL API")
+	if err != nil {
+		return err
+	}
+	if err := h.graphqlAPIService.Delete(orgId, apiId, deletedBy); err != nil {
+		return serviceError(err, fmt.Sprintf("failed to delete GraphQL API %s in org %s", apiId, orgId))
+	}
+
+	httputil.WriteJSON(w, http.StatusNoContent, nil)
```
AnuGayan: **Severity:** 🔵 nit
**Subject:** 204 response encodes a JSON body, inconsistently with the other two handlers in this PR

**Description**
`httputil.WriteJSON(w, http.StatusNoContent, nil)` encodes `null` into a buffer, sets `Content-Type: application/json`, then writes it after a 204. Go's `net/http` suppresses the body for a status that disallows one, but logs `http: request method or response status code does not allow body` on every delete, so this adds a spurious server-log line per request and a `Content-Type` header on a bodyless response. Two other handlers added in this same PR get it right — `internal/handler/graphql_apikey.go:233` and `internal/handler/graphql_deployment.go:167` both use `w.WriteHeader(http.StatusNoContent)` — which is also the repo's majority pattern (`api_key.go:238`, `api_deployment.go:198`, `gateway.go:274`, `llm_deployment.go:178`). The `WriteJSON`-with-nil form exists in `api.go:216` and `application.go`, so this is not new to the codebase, but it is internally inconsistent within this PR. Confirmed against `httpkit/httputil/response.go:12-21` and the call sites listed.

**How to verify**
`DELETE /api/v0.9/graphql-apis/{id}` and watch the platform-api log for `http: request method or response status code does not allow body`; compare with `DELETE /api/v0.9/graphql-apis/{id}/deployments/{deploymentId}`, which logs nothing.

**Suggested fix**
```suggestion
	w.WriteHeader(http.StatusNoContent)
```

**Prompt for Claude**
```text
platform-api/internal/handler/graphql_api.go:220 writes a 204 via
httputil.WriteJSON(w, http.StatusNoContent, nil), which encodes a JSON "null"
body for a status that disallows one.

1. FIRST confirm: read httpkit/httputil/response.go WriteJSON (it always encodes
   and writes the body), and confirm the sibling handlers added in this PR --
   internal/handler/graphql_apikey.go:233 and
   internal/handler/graphql_deployment.go:167 -- use
   w.WriteHeader(http.StatusNoContent) instead. If graphql_api.go already uses
   WriteHeader, STOP.
2. If confirmed, replace line 220 with w.WriteHeader(http.StatusNoContent) so
   DeleteGraphQLAPI matches its two siblings.
3. Scope: change ONLY the GraphQL handler. Leave api.go:216 and application.go
   alone -- they are pre-existing and outside this PR.
4. Confirm afterwards: go build ./platform-api/... is clean, DELETE still returns
   204, and the "response status code does not allow body" log line is gone.
```

AnuGayan (changes requested):
## Follow-up review — deeper pass, and formal request for changes

This is a second pass over the areas the first review did not reach: the 1,315-line generated-code diff, all three platform-api handler files, `repository/graphql_api.go` 200-389, `model/graphql_api.go`, the ~1,400 lines of CLI, the REST design of all 12 new control-plane endpoints, the integration-suite wiring, every modified pre-existing test, the new dependency, the gateway-translator dispatch, and the event-hub replica-sync path.

Five new findings inline. Combined with [the earlier review](https://github.com/wso2/api-platform/pull/3310#pullrequestreview-5053003576), the tally is:

| Sev | First pass | This pass | Total |
|---|---|---|---|
| 🔴 blocker | 1 | 0 | **1** |
| 🟠 major | 3 | 1 | **4** |
| 🟡 minor | 3 | 3 | **6** |
| 🔵 nit | 4 | 1 | **5** |

### What gates this PR

Only the blocker genuinely gates it: **the gateway-controller has no handler for the `graphqlapi.deployed` / `.undeployed` / `.deleted` events platform-api broadcasts**, so the control-plane deployment path — a stated goal of this PR — is a silent no-op. The deployment returns `201 DEPLOYING`, the gateway logs "unknown event type", and no route is ever programmed. Its mirror image is the missing `GraphQLApi` artifact importer, which rejects every DP→CP push with `Unsupported artifact kind "GraphQLApi"`. Neither surfaces to the user.

The new 🟠 in this pass is of the same character: `GET /graphql-apis` declares `sortBy`, `sortOrder` and `query`, and the handler drops all three, so `?query=foo` returns unfiltered results with correct-looking pagination.

### Verified clean

Recording these so the coverage is legible, and because several were plausible failure sites:

- **The integration suite really runs.** Both feature files are in `getFeaturePaths()`, `RegisterGraphQLSteps` is wired into `InitializeScenario`, and the mock backend has a healthcheck plus a `depends_on: service_healthy` edge.
- **No test was weakened.** Every modified pre-existing test is additive; `sqlite_test.go` bumps 4→5 *and* shifts the unsupported-version case to 6/5, preserving the negative assertion's strength.
- **Multi-replica gateway sync works** — I suspected a gap here and was wrong. GraphQL create/update flows through the shared `DeployAPIConfiguration`, which publishes `EventTypeAPI` for every kind (`api_deployment.go:411`), and `handleAPIDelete` (`eventlistener/api_processor.go:133`) is kind-agnostic.
- **`gatewaytranslator.Translate` has no error path for GraphQL**: `Normalize` no-ops for an unregistered kind, and `DownConvert` needs `deploymentArtifact`, which `dto.GraphQLAPIDeploymentYAML` implements.
- **`description` scanned as a bare `string`** looked like a latent NULL-scan panic, but `repository/mcp.go` and `repository/api.go` do exactly the same — consistent, not a divergence.
- **Repository layer**: parameterized queries throughout, `Rebind` for dialect portability, transactions with deferred rollback, `Update` correctly omits `handle`/`project_uuid`/`origin`, compile-time interface assertion.
- **CLI**: all eight commands `url.PathEscape` their path segments, matching the REST CLI.
- **Deployment handler** mirrors REST route-for-route and status-code-for-status-code, including `paginateDeploymentList`.
- **Unbounded `io.ReadAll(r.Body)`** in the gateway handlers and the unpaginated gateway-side list both match every peer handler — neither is a new defect.

The layer-by-layer work here is strong; what is missing is the seam between the two planes. Every inline comment carries a **Prompt for Claude** that names the evidence to check *first*, so a mistaken finding of mine cannot become a bad commit.

---

PR #3319, opened 2026-08-27
feat(helm): fail on contradictory storage dsn + passwordSecretRef

### LINKED ISSUE #3318: Gateway Helm: fail on contradictory DSN + passwordSecretRef
### Description

The `gateway-helm-chart` lets an operator configure the controller's external database (`postgres`/`sqlserver`) either via a full `dsn` **or** via individual `host`/`port`/`database`/`user` fields with the password injected separately from a Secret (`passwordSecretRef`). These two modes are mutually exclusive, but the chart neither enforces nor clearly documents that — leading to silent misconfiguration.

**Root cause in the controller:** `buildPostgresDSN` (`gateway/gateway-controller/pkg/storage/postgres.go:140`) and `buildSQLServerDSN` (`pkg/storage/sqlserver.go`) return the `dsn` as-is when it is non-empty and never merge the separately-configured `Password`. The password field is only used when `dsn` is empty. At the chart level, `gateway-config.yaml` renders the `password = '{{ env … }}'` line only in the non-DSN branch, so when `dsn` is set the injected `APIP_GW_CONTROLLER_STORAGE_POSTGRES_PASSWORD` env var has nothing to consume it.

**Two footguns this creates:**
- `dsn` set **without** a password + `passwordSecretRef` set → the secret is silently ignored; the controller connects with no password and crashloops on a Postgres `password authentication failed` error that points at the DB, not at the contradictory Helm values.
- `dsn` set **with** an inline password + `passwordSecretRef` set → works (DSN wins), but the mounted secret is dead weight and the password lands in the ConfigMap in plaintext.

**Note — "both set" is sometimes valid:** the secure DSN pattern embeds a `{{ env "APIP_GW_CONTROLLER_STORAGE_POSTGRES_PASSWORD" }}` token inside the `dsn`, which *does* consume `passwordSecretRef` at load time and keeps the password out of the ConfigMap. Any guard must allow this.

### Proposed fix

- **Guarded `fail`:** in `gateway-config.yaml`, for both the `postgres` and `sqlserver`/`database` blocks, `fail` at template render time when `dsn` is set **and** `passwordSecretRef.name` is set **and** the `dsn` does not reference the password env token — with a message directing the operator to either embed the `{{ env … }}` token in the DSN or drop `passwordSecretRef`. Consistent with the chart's existing `fail`/`required` validation (removed `sqlserver` key, missing host/db/user, `encrypt` enum).
- **Docs:** correct the misleading `values.yaml` comments in the `postgres`/`database` blocks, which currently state both "Full DSN takes precedence over individual fields when set" and "Password is injected separately via … passwordSecretRef" — together implying a password-less DSN plus a separate secret works, which it does not.

### Version

_No response_

### Related Issue

_No response_

renuka-fernando:
## Purpose

The `gateway-helm-chart` lets an operator configure the controller's external database (`postgres`/`sqlserver`) either via a full `dsn` or via individual `host`/`port`/`database`/`user` fields with the password injected separately from a Secret (`passwordSecretRef`). These two modes are mutually exclusive, but the chart neither enforced nor clearly documented that, so a common misconfiguration passed silently.

The controller's `buildPostgresDSN` (`gateway/gateway-controller/pkg/storage/postgres.go:140`) and `buildSQLServerDSN` (`pkg/storage/sqlserver.go`) return the `dsn` verbatim when set and never merge the separately-configured password. At the chart level the `password = '{{ env ... }}'` line is only rendered in the non-DSN branch, so when a `dsn` is set the injected `APIP_GW_CONTROLLER_STORAGE_POSTGRES_PASSWORD` env var has nothing to consume it. An operator who set a password-less `dsn` alongside a `passwordSecretRef` got a controller that connected with no password and crashlooped on an opaque Postgres `password authentication failed` error that pointed at the database, not at the contradictory Helm values. The `values.yaml` comments made it worse by stating both "Full DSN takes precedence over individual fields when set" and "Password is injected separately via passwordSecretRef", which read as if a password-less DSN plus a separate secret works.

Resolves #3318.

## Goals

- Turn the silent misconfiguration into a clear, pre-deploy failure.
- Correct the misleading documentation so operators pick the right mode up front.
- Preserve the legitimate secure pattern of embedding an `{{ env ... }}` token inside the DSN so the password stays in a Secret rather than the ConfigMap.

## Approach

- Add a render-time guard in `templates/gateway/gateway-config.yaml` for both the `postgres` and `sqlserver`/`database` blocks: `fail` when a `dsn` is set and the matching `passwordSecretRef.name` is set and the `dsn` does not reference the injected password env token. The error message tells the operator to either embed the token in the DSN or drop the secret ref. This is consistent with the chart's existing `fail`/`required` validation (removed `sqlserver` key, missing host/db/user, `encrypt` enum).
- The token check (`contains "APIP_GW_CONTROLLER_STORAGE_POSTGRES_PASSWORD"` / `..._DATABASE_PASSWORD`) intentionally allows the secure "token inside the DSN" pattern to keep working.
- Rewrite the `values.yaml` comments in the `postgres`, `database`, and both `passwordSecretRef` blocks to state that the two modes are mutually exclusive, that the DSN is used verbatim, that a literal password in the DSN lands in the ConfigMap in plaintext, and to show a copy-pasteable token example.

## User stories

As an operator deploying the gateway against an external database, when I misconfigure the connection by setting a `dsn` alongside a `passwordSecretRef` that the DSN does not use, I get an immediate, actionable Helm render error instead of a controller crashloop with a misleading database error.

## Documentation

N/A. The change is limited to in-chart `values.yaml` comments and a render-time error message; there is no external product documentation impacted.

## Automation tests

- Unit tests
  > N/A. This is a Helm chart templating change with no Go/Java code path; the chart has no unit-test harness. Verification was done via `helm template` and `helm lint`.
- Integration tests
  > Validated with `helm template` across the full matrix for both engines, plus a no-regression check and `helm lint`:
  >
  > | Scenario | Expected | Result |
  > |---|---|---|
  > | postgres: individual fields + `passwordSecretRef` (no dsn) | render | pass |
  > | postgres: `dsn` only, no `passwordSecretRef` | render | pass |
  > | postgres: `dsn` (no token) + `passwordSecretRef` | fail | fails with clear message |
  > | postgres: `dsn` with token + `passwordSecretRef` | render (token preserved) | pass |
  > | sqlserver: individual fields + `passwordSecretRef` (no dsn) | render | pass |
  > | sqlserver: `dsn` (no token) + `passwordSecretRef` | fail | fails with clear message |
  > | sqlserver: `dsn` with token + `passwordSecretRef` | render (token preserved) | pass |
  > | default sqlite path | render | pass (no regression) |
  >
  > `helm lint .` reports 0 charts failed.

## Security checks

- Ran FindSecurityBugs plugin and verified report? N/A (Helm chart change; FindSecurityBugs targets Java bytecode, no Java/Go code changed)

## Samples

N/A

## Related PRs

N/A

## Test environment

Helm v3.18.3, macOS (Darwin 25.5.0). Chart rendered and linted locally; no live database connection required for these tests.

### DIFF
--- a/kubernetes/helm/gateway-helm-chart/templates/gateway/gateway-config.yaml
+++ b/kubernetes/helm/gateway-helm-chart/templates/gateway/gateway-config.yaml
@@ -82,9 +82,12 @@ data:
       is NOT merged into it. Setting both, unless the DSN itself references the
       injected password env token, means the secret is silently ignored and the
       controller connects with no password. Fail loudly instead. The secure way to
-      combine a DSN with the secret is to embed the token in the DSN.
+      combine a DSN with the secret is to embed the token in the DSN. Match the full
+      `{{ env "VAR" }}` interpolation token (tolerating the {{- trim marker, spacing,
+      and the "" default arg), not just the bare variable name, so a DSN that merely
+      mentions the name elsewhere (e.g. in the dbname) still fails the guard.
     */ -}}
-    {{- if and $controller.postgres.passwordSecretRef.name (not (contains "APIP_GW_CONTROLLER_STORAGE_POSTGRES_PASSWORD" $pg.dsn)) }}
+    {{- if and $controller.postgres.passwordSecretRef.name (not (regexMatch `\{\{-? *env +"APIP_GW_CONTROLLER_STORAGE_POSTGRES_PASSWORD"` $pg.dsn)) }}
     {{- fail "gateway.config.controller.storage.postgres.dsn is set together with gateway.controller.postgres.passwordSecretRef, but the DSN does not reference the injected password. A full DSN is used verbatim; the separately-injected password is ignored. Either embed the token in the DSN, e.g. dsn: 'postgres://user:{{ env \"APIP_GW_CONTROLLER_STORAGE_POSTGRES_PASSWORD\" }}@host:5432/db?sslmode=require', or unset gateway.controller.postgres.passwordSecretRef and put the password directly in the DSN." }}
     {{- end }}
     dsn = {{ $pg.dsn | quote }}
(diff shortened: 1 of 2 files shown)


renuka-fernando:
@coderabbitai review

renuka-fernando:
@coderabbitai approve

Thushani-Jayasekera (approved):

PR #3324, opened 2026-08-28
Add a generic deployment override and a Deployments capability on pdk.Deps

dakshina99:
## Purpose
The deployment API can only override a fixed set of metadata keys
(`endpointUrl`, `vhostMain`, `vhostSandbox`), promoting an existing deployment
onto a gateway reuses the base artifact without adapting it to the target
gateway's schema version, and external plugins have no typed access to the
deployment lifecycle. This change makes per-deployment customization open-ended,
makes cross-version promotion safe, and exposes deploy/read/undeploy to plugins
alongside the existing Gateways and Projects capabilities.

## Goals
- Let a caller customize any field of an API's config for a single deployment,
  without adding a new field per customization.
- Make promoting an existing deployment onto a gateway of a different data
  version produce a valid artifact, without re-reading the API definition.
- Expose the deployment lifecycle on `pdk.Deps` for external plugins.

## Approach
- `DeployRequest.overrides`: an optional structured override document
  deep-merged onto the resolved deployment definition before it is sent to the
  gateway, and persisted with the deployment so it can be read back and carried
  forward when the deployment is used as a promotion base. The existing
  `endpointUrl`/`vhost` metadata overrides are unchanged.
- Promotion (`base` = an existing deployment) now re-translates the base
  deployment's already-rendered artifact to the target gateway's data version,
  so promoting across gateways on different versions yields a valid artifact.
  The base API definition is never re-read: the source data version is computed
  from the base artifact's own `apiVersion`, and only the immutable artifact
  `Kind` is read from the API record.
- `pdk.Deps.Deployments` capability (`DeployAPIByHandle`,
  `GetDeploymentsByHandle`, `GetDeploymentByHandle`,
  `UndeployDeploymentByHandle`), satisfied verbatim by `DeploymentService`,
  mirroring the existing Gateways/Projects capabilities.

## User stories
N/A

## Documentation
N/A — an internal capability and deploy-request extension; no product doc impact.

## Automation tests
 - Unit tests
   > Generic override deep-merge (`TestMergeGenericOverrides`, `TestDeepMergeMap`)
   > and the base-flow override applier (`TestApplyDeploymentOverrides`). Existing
   > deployment and pdk suites pass.
 - Integration tests
   > Existing deployment integration suite is unaffected by this change.

## Security checks
 - Ran FindSecurityBugs plugin and verified report? no

## Samples
N/A

## Related PRs
Builds on #3300 (Projects capability + RouteOverride on `pdk.Deps`), now merged.

## Test environment
Go unit test suite (sqlite); local build.

### DIFF
--- a/platform-api/api/generated.go
+++ b/platform-api/api/generated.go
@@ -1031,6 +1031,9 @@ type DeployRequest struct {
 
 	// Name Name/label for this deployment (e.g., "v1.0-prod", "hotfix-2024-01-15")
 	Name string `binding:"required" json:"name" yaml:"name"`
+
+	// Overrides Optional structured override document deep-merged onto the deployment definition before it is sent to the gateway, letting a caller customize any field of the API config for this deployment. The applied overrides are persisted with the deployment.
+	Overrides *map[string]interface{} `json:"overrides,omitempty" yaml:"overrides,omitempty"`
 }
 
 // DeploymentListResponse defines model for DeploymentListResponse.
--- a/platform-api/internal/constants/constants.go
+++ b/platform-api/internal/constants/constants.go
@@ -251,6 +251,10 @@ const (
 	MetadataKeyVhostMain = "vhostMain"
 	// MetadataKeyVhostSandbox is the metadata key for the per-deployment sandbox vhost value.
 	MetadataKeyVhostSandbox = "vhostSandbox"
+	// MetadataKeyOverrides is the metadata key under which the applied generic
+	// override document is persisted, so it can be read back (e.g. to prefill a
+	// re-deployment from the same environment).
+	MetadataKeyOverrides = "overrides"
 	// VhostGatewayDefault is the sentinel value that instructs the gateway-controller to resolve
 	// and persist the current gateway default vhosts, ensuring deployments are immune to future
 	// gateway config changes.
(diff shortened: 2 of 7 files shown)


dakshina99:
Updated (force-push): the override now **rejects an override that targets an immutable identity field** (`apiVersion`, `kind`, `metadata.name`, `spec.context`, `spec.version`, `spec.operations`, `spec.channels`), so a customization can only customize a deployment of an API — never repoint or redefine it. Covered by `TestMergeGenericOverrides_ProtectsImmutableFields`.

dakshina99:
Superseded by #3364: same change rebased onto current `main` (this branch was ~157 commits behind), plus a fix for the review finding on override metadata during promotion — a promoted deployment now inherits the base's override document instead of reporting none while its content carries them — with tests (`TestEffectiveOverrideDocument`).

On the other two review findings: the `generated.go` one concerns the MCP server-info fetch path, which this change does not touch. The outbound-URL one is a pre-existing property of the deployment API — `metadata.endpointUrl` already lets an authenticated caller set an arbitrary upstream URL — so the override document does not widen it in kind; hardening outbound calls (dialer protection, address checks) is worth doing but belongs in its own change rather than being bolted onto this one.

PR #3338, opened 2026-08-31
feat(APICP): add wizard-based REST API creation flow

ShavinAnjithaAlpha:
## Purpose

The API Control Plane portal needs a guided way to create an API. Creating one involves several unrelated decisions. What kind of API it is, where its definition comes from, and how it is configured and asking for all of them on one screen makes the flow hard to follow and easy to get wrong.

Resolves https://github.com/wso2-enterprise/apim-saas/issues/2897, https://github.com/wso2-enterprise/apim-saas/issues/2931

## Goals

- A stepped wizard that asks one thing at a time: API type → how to define it → configure and create.
- Support bringing an existing OpenAPI contract from wherever it already lives, and tell the user up front when that contract won't work.
- Show what is happening while the API is created, and let a failed attempt be retried without re-entering the form.

## Approach

**`apis/create/` - the creation wizard**

`ApiCreationWizard` owns the three steps and the draft state each step contributes to. Steps are navigable, so going back doesn't lose what was already filled in.

- **Step 1 - API type.** REST is selectable; WebSocket, GraphQL, WebSub and gRPC are listed but disabled, so the shape of the product is visible rather than hidden.
- **Step 2 - define the API.** Two approaches sharing one preview pane:
  - *Start with a contract* — import from a URL, a file upload, a GitHub repository (with a directory browser for picking the spec out of a tree), or SwaggerHub (org/API/version lookup).
  - *Design from scratch* — starts from a default OpenAPI skeleton to edit.

Imported documents are validated before the wizard moves on: dialect support (OpenAPI 3.0/3.1, Swagger 2.0), missing `paths`/operations/title/version, bad path keys, external `$ref`s, and a size ceiling. Errors block the import, warnings let it through. Both surfaced as a readable issue list. Name, version, context, upstream and operations are read off the accepted document and pre-fill the next step.
- **Step 3 - configure and create.** A validated form over the pre-filled draft (display name, identifier with availability check, version, base path, description, upstream URL), submitting through the `useCreateRestApi` hook.
- **Progress screen.** Replaces the form on submit with staged progress (validating → creating → finalizing → done), then redirects to the created API. On failure it offers retry or back-to-form, with the submitted values still intact.

image

image

image

image

image

**Also in this PR (supporting refactors)**

- APIs pages organized into `create/`, `listing/`, `overview/`, `develop/` with grid/list views and shared display helpers, on the new data access layer hooks.
- Gateways pages reworked: grid + list views, simplified gateway card, detail page made consistent with the other portals, and a create page with form controls and field validation.
- Projects/organization pages restructured; `@` absolute-import alias configured in TypeScript and Vite.
- Deployments, Policies, Routing, System pages integrated with the new data access layer and add i18n support for related components.
- Refactored the Insight Menu Panels to route users to the external Moesif portal.
- Removed the old mvp data access layer entirely.

## User stories

- As an API developer, I want to create an API by importing an existing OpenAPI contract from a URL, file, GitHub or SwaggerHub.
- As an API developer, I want to be told why a contract was rejected before I get to the end of the flow.
- As an API developer, I want to start from a blank skeleton when I have no contract yet.
- As an API developer, I want to see that creation is in progress, and retry without losing what I entered if it fails.

## Documentation

N/A -> portal UI change; no new configuration or API surface that needs product documentation.


Induwara04:
@coderabbitai review

portals/api-control-plane/src/pages/appShell/appShellPages/apis/overview/ApiKeysPanel.tsx
```diff
@@ -35,38 +35,173 @@ import {
...
  */
-export function ApiKeysPanel({ api }: { api: Api }) {
+export function ApiKeysPanel({ restApiId }: { restApiId: string }) {
+  const intl = useIntl();
   const { notify } = useNotifications();
-  const keysQuery = useApiKeys(api);
+  // The spec has no per-API key listing — the only read is the caller's own
+  // inventory across artifacts, so this narrows to REST API keys server-side
+  // and to this API here. It therefore shows the signed-in user's keys only.
+  const keysQuery = useMyApiKeys({ type: ['RestApi'] });
```
AnuGayan: **Severity:** 🟠 major
**Subject:** `useMyApiKeys({ type: ['RestApi'] })` takes the server's default page of 20 and filters it in the browser

**Description**
This reads the caller's key inventory across *all* REST APIs and then narrows to this one with `.filter((key) => key.artifactId === restApiId)` just below. No `limit` is passed, so the server applies its default. Once a user holds more than 20 REST-API keys org-wide, this API's keys fall off the first page and vanish from the table — the user cannot see or revoke them, and nothing indicates anything is missing. Confirmed against `platform-api/resources/openapi.yaml` `components.parameters.limit-Q` (`default: 20`, `maximum: 100`) and the `/me/api-keys` operation, which takes `limit`/`offset`. `apis/listing/ApiList.tsx` in this same PR does server-side `limit`/`offset` correctly, so this is a deviation from the pattern the PR itself establishes.

**How to verify**
In devtools, watch the request to `/me/api-keys?type=RestApi` — it carries no `limit`, and the response `list` caps at 20. To see it end-to-end: create 21+ REST-API keys in one org as one user, then open the overview of the API whose key was created first.

**Suggested fix**
Raising the page to the spec maximum fixes the common case:

```suggestion
  const keysQuery = useMyApiKeys({ type: ['RestApi'], limit: 100 });
```

That still breaks past 100. The correct fix is to page until the matches are exhausted (or `list.length < limit`), or to add a per-API listing endpoint — the spec has none today, which this hook's own doc comment already notes.

**Prompt for Claude**
```text
In portals/api-control-plane/src/pages/appShell/appShellPages/apis/overview/ApiKeysPanel.tsx,
the panel calls useMyApiKeys({ type: ['RestApi'] }) and then filters client-side on
key.artifactId === restApiId.
1. First confirm the paging default: read platform-api/resources/openapi.yaml, find
   components.parameters.limit-Q and the /me/api-keys operation (listUserAPIKeys). Confirm limit
   defaults to 20 and maxes at 100, and confirm there is NO per-REST-API key LISTING operation
   (only POST /rest-apis/{restApiId}/api-keys and PUT/DELETE on {apiKeyId}). If a per-API list
   endpoint exists, stop and tell me — use that instead and the rest of this is wrong.
2. If confirmed, make the panel exhaustive rather than first-page-only: pass limit: 100 and page
   through offsets until a page returns fewer than limit items, accumulating only the keys whose
   artifactId matches restApiId. Keep the existing useMemo shape so the table's identity is stable.
3. Do not change the create/revoke mutations or the invalidation in useInvalidateApiKeys.
4. Confirm by asserting in a test that a two-page response yields matching keys from BOTH pages.
```

portals/api-control-plane/src/pages/appShell/appShellPages/apis/create/components/ContractSourceForm.tsx
```diff
@@ -0,0 +1,1717 @@
...
+    setFileError(null);
+    setFetchError(null);
+    setFile(next);
+  };
+
+  /** Rejects the current file, whether it was unsupported or simply removed. */
+  const handleFileReject = () => {
+    const unsupported = file === null;
+    setFile(null);
+    setFileError(unsupported ? 'unsupported' : null);
```
AnuGayan: **Severity:** 🟠 major
**Subject:** `handleFileReject` infers "unsupported" from `file === null`, but `onReject` serves two different events

**Description**
One `onReject` callback serves both "the user clicked the X to remove the file" and "the picked/dropped file has the wrong extension" — `ContractFileControl.take()` calls `onReject()` when `accepts()` fails, and the remove `IconButton` (rendered only when `file !== null`) calls the same thing. This handler then guesses which happened. With a valid file already selected, dropping `notes.txt` takes the `unsupported = false` branch: the valid selection is discarded and the helper text reverts to the neutral "Accepted: …" line. The user loses their spec and is told nothing. Confirmed by reading both call sites in this file.

**How to verify**
Step 2 → Upload a file → pick a valid `openapi.yaml` (the name appears under the drop zone) → now pick `README.md` through the same picker. The `openapi.yaml` row disappears and no error is shown.

**Suggested fix**
Make the reason explicit rather than inferred. This half is one-clickable:

```suggestion
  const handleFileReject = (reason: 'unsupported' | 'removed') => {
    setFile(null);
    setFileError(reason === 'unsupported' ? 'unsupported' : null);
```

It needs the other half in the same commit: change `ContractFileControlProps.onReject` to `(reason: 'unsupported' | 'removed') => void`, pass `'unsupported'` from `take()`'s failure branch, and `'removed'` from the remove `IconButton`'s `onClick`.

**Prompt for Claude**
```text
In portals/api-control-plane/src/pages/appShell/appShellPages/apis/create/components/ContractSourceForm.tsx:
1. First confirm the two call sites of ContractFileControl's onReject: (a) inside `take()`, when
   `accepts(candidate)` is false, and (b) the remove IconButton rendered only when `file !== null`.
   If onReject has only ONE call site, stop and tell me — the finding is wrong.
2. If both exist, change ContractFileControlProps.onReject to
   `(reason: 'unsupported' | 'removed') => void`; pass 'unsupported' from take() and 'removed' from
   the IconButton. Update handleFileReject to switch on the reason instead of on `file === null`.
3. Do not change `accepts()`, the extension list, or the drag-and-drop handlers.
4. Confirm with a test: select a valid .yaml, then reject an unsupported one, and assert the
   'unsupported' helper text is shown (currently it is not).
```

portals/api-control-plane/src/pages/appShell/appShellPages/apis/create/ApiCreationWizard.tsx
```diff
@@ -0,0 +1,193 @@
...
+  };
+
+  const getSubtitleForStep = (step: ApiCreationStepKey) => {
+    switch (step) {
+      case 'apiType':
+        return 'This decides how the gateway exposes your backend. Only REST is available today.';
+      case 'source':
+        return 'Bring an existing contract, or start from a blank slate and fill in the details yourself.';
+      case 'configure':
+        return `${API_TYPES.find((type) => type === apiType)?.rawTitle || 'API'}.DEFAULT SKELETON - EDIT IT OR ASK AI TO REFINE IT`;
```
AnuGayan: **Severity:** 🟠 major
**Subject:** The "Configure and create" subtitle renders `REST API.DEFAULT SKELETON - EDIT IT OR ASK AI TO REFINE IT`

**Description**
This string is rendered verbatim as the page subtitle under the "Configure and create" heading. It reads as leftover working notes: shouty caps, a stray `.` joining two unrelated clauses, and it describes the *skeleton editor* rather than the configure step it sits above. Confirmed by reading `getSubtitleForStep` and its single call site further down this file. It ships on every pass through step 3, whichever source the user came from.

**How to verify**
Open the wizard, pick REST, import any contract, click Next. Read the subtitle under "Configure and create".

**Suggested fix**
```suggestion
        return 'Name it, set where it routes, and create it.';
```

Better still, route it through `defineMessages` like every other component in this directory — see the companion comment on `getTitleForStep`.

**Prompt for Claude**
```text
In portals/api-control-plane/src/pages/appShell/appShellPages/apis/create/ApiCreationWizard.tsx,
getSubtitleForStep('configure') returns a template string ending
".DEFAULT SKELETON - EDIT IT OR ASK AI TO REFINE IT".
1. Confirm it is rendered: find where getSubtitleForStep is called in the same file and check the
   result goes into a <Typography>. If it is dead code, stop and tell me.
2. If rendered, replace all three of getTitleForStep/getSubtitleForStep's hardcoded strings with
   react-intl messages defined via defineMessages in this file, following the pattern in
   ./components/ApiCreationSteps.tsx. The 'configure' subtitle must describe the configure step,
   not a skeleton editor.
3. Where the copy interpolates the API type, use `apiType?.rawTitle` directly rather than
   API_TYPES.find((type) => type === apiType)?.rawTitle — the identity search is redundant.
4. Confirm with `npm run i18n:extract` then `npm run lint:i18n`, and check the new ids appear in
   src/i18n/messages/en.json.
```

portals/api-control-plane/src/pages/appShell/appShellPages/apis/create/ApiCreationWizard.tsx
```diff
@@ -0,0 +1,193 @@
...
+              onDataFetched={handleDataExtracted}
+              onBack={() => setStep('apiType')}
+              onRefreshSwaggerHubOrganizations={() => {}}
+            />
+          )}
+
+          {step === 'configure' && (
+            <Box sx={{ maxWidth: '80%', mt: 1 }}>
+              <GeneralCreateApiForm
+                initialValues={prefilledData}
```
AnuGayan: **Severity:** 🟠 major
**Subject:** "Back to configuration" after a failed create discards everything the user typed

**Description**
The PR description promises "On failure it offers retry or back-to-form, with the submitted values still intact." Retry does hold, but Back does not. This component early-returns `<ApiCreationProgress/>` whenever `submittedValues` is set, which unmounts the form entirely. `onBack` clears `submittedValues`, the form remounts, and its state comes from `useState(() => getInitialValues(props.initialValues || {}, projectHandler))` — a mount-only lazy initialiser over `prefilledData`, i.e. the draft read off the *imported spec*, not what the user edited. Every hand-typed change (identifier, base path, description, target URL) is silently reverted. Confirmed against `GeneralCreateApiForm.tsx` (lazy init) and the conditional return above.

**How to verify**
Import a contract, on step 3 change the identifier and target URL, block the create request in devtools so it fails, then click "Back to configuration". The form shows the spec-derived values again, not your edits.

**Suggested fix**
`initialValues` is `Partial<GeneralApiCreationFormState>` and `submittedValues` is the full state, so this type-checks — and it works because the form remounts, re-running the lazy initialiser:

```suggestion
                initialValues={submittedValues ?? prefilledData}
```

**Prompt for Claude**
```text
In portals/api-control-plane/src/pages/appShell/appShellPages/apis/create/ApiCreationWizard.tsx and
./components/GeneralCreateApiForm.tsx:
1. First confirm the state loss. In GeneralCreateApiForm, confirm formState comes from
   useState(() => getInitialValues(props.initialValues || {}, projectHandler)) — a lazy initialiser
   that runs only on mount — and that the wizard unmounts the form by early-returning
   <ApiCreationProgress/> while submittedValues is set. If the form is kept mounted (e.g. hidden
   rather than replaced), stop and tell me — the finding is wrong.
2. If confirmed, pass `initialValues={submittedValues ?? prefilledData}` so the remount restores what
   the user submitted. Keep onBack clearing submittedValues and resetting the mutation.
3. Do not change the retry path, and do not make the form re-sync from props on every render — that
   would fight the user's typing.
4. Confirm manually: force a create failure, edit a field, go Back, check the edited value survives.
```

AnuGayan (commented):
## Summary

A three-step API creation wizard (type → source → configure) plus a large supporting reorganisation of the APIs, Gateways and Projects pages onto the new data-access layer. Reviewed at commit `823cc37` against merge-base `ae8fb36` — `gh pr diff` refuses this one (20K+ lines over GitHub's cap), so this was read from a checkout of the branch.

The new pure-logic modules are the strongest part of the PR: `specValidation.ts` gets the error/warning split right (block only what leaves nothing to show or nothing to create), the `$ref` walk is both node-capped and `WeakSet`-cycle-aware with a comment explaining why YAML anchors demand it, and `createRestApiBody.ts` documents every form↔wire divergence including the `RestApis`→`RestApi` trap. The wizard shell that stitches them together is noticeably less finished than the pieces.

## Verdict

🔧 **Request changes** — five user-facing defects in the new flow plus one silent correctness bug in `ApiKeysPanel`. None are hard to fix; all are visible to the first person who uses the wizard. (Posted as a comment rather than a formal Request Changes.)

**Blocking-ish, in rough priority order:**

1. `ApiKeysPanel` filters one 20-item server page client-side, so a user with >20 REST-API keys silently cannot see or revoke this API's keys.
2. Dropping an unsupported file over an existing selection clears the good file and shows no error.
3. Step 3's subtitle renders placeholder copy: `REST API.DEFAULT SKELETON - EDIT IT OR ASK AI TO REFINE IT`.
4. "Back to configuration" after a failed create discards everything the user typed — the PR description promises the opposite.
5. "Authorize GitHub" and the SwaggerHub refresh button are wired to `() => {}` and render enabled.

Each is an inline comment below with a repro and a paste-ready prompt. Eight minors and six nits follow the same format.

## One finding with no line to anchor to: the new pure logic has no tests

`find src/pages/appShell/appShellPages/apis/create -name "*.test.*"` returns exactly one file — `components/ApiCreationProgress.test.tsx`, which is good and pins the "never claims 100% before the server answers" invariant properly. Untested: `specValidation.ts` (236 lines), `specDetails.ts` (197), `github.ts` (307), `swaggerHub.ts` (184), `createRestApiBody.ts` (143). All five are pure, dependency-free, and encode decisions that are cheap to get wrong and expensive to notice: the `RestApis` → `RestApi` kind rename, dropping blank optionals so `upstream.sandbox` doesn't fail the spec's exactly-one-of rule, the `$ref` walk's cycle-awareness, Swagger 2.0's `host`/`basePath`/`schemes` assembly, the `X-Versions` star-stripping.

The repo's convention is co-located tests — `gatewaySetup.test.ts`, `developEdit.test.ts` and `policyHub.endpoints.test.ts` are all in this PR — so this reads as a gap rather than a change of practice. Highest-value cases if you want a minimum: `toCreateRestApiBody` emits `kind: 'RestApi'` and omits blank `id`/`description`/`sandbox`; `validateApiSpec` returns `invalid` for a doc with `paths` but no supported methods, and `valid` with a `noServers` warning for one with no `servers`; `extractApiDetails` skips a templated server URL; `findExternalRefs` terminates on a self-referential structure; `readVersions` strips the `*`.

## What's good

- **The comments explain decisions, not mechanics.** `readServerUrl` says *why* a templated URL is dropped; `useRestApiIdAvailability` says why it's built on the list query rather than a 404-probe; `keepPreviousData` in `useRestApis` carries the pagination rationale. I checked the load-bearing claim — `query` really is a case-insensitive substring filter on the resource id — against `openapi.yaml` `components.parameters.query-Q`, and it holds.
- **`ApiList.tsx` is a model of the pattern**: server-side `limit`/`offset`, debounced search, page reset on filter change. `ApiKeysPanel` is the one place that doesn't follow it.
- **`GeneralCreateApiForm`'s availability probe** handles the states most implementations miss: a pending debounce, an in-flight request, and a *failed* probe are all distinguished, and a failed probe falls back to plain helper text rather than claiming a handle is free.
- **`ApiCreationProgress`'s honesty invariant** — the bar eases toward a ceiling and only snaps to 100 when the server answers — is both a good decision and pinned by a test that would catch a regression.
- **`createRestApiBody.ts` builds the body field-by-field** specifically so `readonly` server-owned fields can't leak in, and says so.

*(Automated review. Every finding names how to verify it — if one is wrong, say so and I'll correct it in place.)*

portals/api-control-plane/src/pages/appShell/appShellPages/apis/create/ApiCreationWizard.tsx
```diff
@@ -0,0 +1,193 @@
...
+              onChange={(apiType) => {
+                setApiType(apiType);
+                setStep('source');
+              }}
+            />
+          )}
+
+          {step === 'source' && (
+            <DefineApiPanel
+              onAuthorizeGitHub={() => {}}
```
AnuGayan: **Severity:** 🟠 major
**Subject:** "Authorize GitHub" and the SwaggerHub refresh control are wired to `() => {}`

**Description**
Both no-op stubs reach real, enabled controls: a full-width outlined "Connect GitHub" button occupying its own card beside the repository field, and a tooltipped refresh `IconButton` next to the SwaggerHub organization field (`ContractSourceForm.tsx`, `onClick={onAuthorizeGitHub}` and `onClick={onRefreshSwaggerHubOrganizations}`). Clicking either produces nothing — no spinner, no message, no state change. The component doc comments acknowledge this ("inert until wired"), but neither control is gated on that, so they ship as dead ends in a first-run flow.

**How to verify**
Step 2 → Start with a contract → GitHub repository: the "Connect GitHub" card renders and its button is enabled. Click it — nothing happens. Same for the ↻ button on the SwaggerHub tab.

**Suggested fix**
Don't render a control that has no handler. No one-click suggestion here, because the fix spans two files — committing only this side would leave the buttons rendering with an undefined `onClick`, which is what they effectively do today.

In `ContractSourceForm.tsx`, gate each control on its prop:

```tsx
{repositorySettled || onAuthorizeGitHub === undefined ? null : (
  // … the "or" divider + Connect GitHub card
)}
```

and likewise wrap the refresh `IconButton` in `onRefreshSwaggerHubOrganizations === undefined ? null : (…)`. Then delete both `() => {}` props from this call site — they are already optional on `DefineApiPanelProps`.

**Prompt for Claude**
```text
In portals/api-control-plane/src/pages/appShell/appShellPages/apis/create/:
1. Confirm the dead controls: ApiCreationWizard.tsx passes onAuthorizeGitHub={() => {}} and
   onRefreshSwaggerHubOrganizations={() => {}} to DefineApiPanel, and
   components/ContractSourceForm.tsx renders a Button with onClick={onAuthorizeGitHub} and an
   IconButton with onClick={onRefreshSwaggerHubOrganizations}. If either handler is actually
   implemented somewhere, stop and tell me which.
2. If both are no-ops, hide each control when its handler prop is undefined, and remove the two
   `() => {}` stubs from ApiCreationWizard so nothing dead renders. Keep the props optional so the
   flows can be wired later without touching the render.
3. Do not remove the props or the surrounding "or" divider layout wholesale — the GitHub card's grid
   sizing is shared with the repository card.
4. Confirm by opening step 2 → GitHub (Connect card gone) and step 2 → SwaggerHub (refresh icon
   gone).
```

portals/api-control-plane/src/pages/appShell/appShellPages/apis/create/ApiCreationWizard.tsx
```diff
@@ -0,0 +1,193 @@
...
+  type ApiCreationProgressStatus,
+} from './components/ApiCreationProgress';
+
+export const ApiCreationWizard = () => {
+  const [step, setStep] = useState<ApiCreationStepKey>('apiType');
+  const [apiType, setApiType] = useState<ApiType | null>(null);
+
+  const [prefilledData, setPrefilledData] = useState<Partial<GeneralApiCreationFormState>>({});
+
+  const getTitleForStep = (step: ApiCreationStepKey) => {
```
AnuGayan, line 90: **Severity:** 🟡 minor
**Subject:** `getTitleForStep`/`getSubtitleForStep` return raw English strings, bypassing react-intl

**Description**
Every other new component in this PR uses `defineMessages` (`ApiCreationSteps.tsx`, `ApiCreationProgress.tsx`, `GeneralCreateApiForm.tsx` — the PR adds 1,817 lines to `en.json`). These two functions return six hardcoded English strings instead. They will never be extracted by `npm run i18n:extract` and never translated. And because `eslint.i18n.config.js` promotes only `formatjs/no-literal-string-in-jsx`, which fires on literals *written inside JSX*, a string returned from a function and interpolated as `{getTitleForStep(step)}` slips through — so `lint:i18n` stays green. Confirmed against `eslint.i18n.config.js`.

**How to verify**
`npm run lint:i18n` — clean. Then `grep 'What kind of API are you exposing' src/i18n/messages/en.json` — no match.

**Suggested fix**
Move all six strings into a `defineMessages` block in this file and return `MessageDescriptor`s, rendering them with `<FormattedMessage>`; mirror the shape of `components/ApiCreationSteps.tsx`. Same change as the companion comment on the step-3 subtitle — do them together.

portals/api-control-plane/src/pages/appShell/appShellPages/apis/create/components/GeneralCreateApiForm.tsx
```diff
@@ -0,0 +1,633 @@
...
+      ...(draftData.upstream || {}),
+    },
+  };
+
+  const id = merged.id.trim() === '' ? toHandle(merged.displayName) : merged.id;
+
+  return {
+    ...merged,
+    id,
+    context: toBasePath(projectHandler, id, merged.version),
```
AnuGayan: **Severity:** 🟡 minor
**Subject:** `getInitialValues` overwrites the imported `context` unconditionally, making the extractor's work dead

**Description**
`utils/specDetails.ts` derives a `context` from the document's first non-templated server URL, falling back to a slug of `info.title`, with a deliberate guard against a bare `"/"`. This line discards all of it: `toBasePath(...)` runs regardless of whether `draftData.context` was set. So `readContextFromUrl`, `slugify` and the `context === '/'` guard are all dead from the form's perspective, and the PR description's claim that "context … [is] read off the accepted document and pre-fill the next step" doesn't hold. Confirmed by reading both functions.

**How to verify**
Import a spec whose `servers[0].url` is `https://api.example.com/v2/orders`. The Base Path field shows `/{project}/{id}/v{version}`, never `/v2/orders`.

**Suggested fix**
Decide which wins and delete the loser. If the document should win when it names one:

```suggestion
    context: draftData.context ?? toBasePath(projectHandler, id, merged.version),
```

If the platform's base path always wins (which the helper text "Built from the project, identifier and version" implies), keep this line and drop `context` from `extractApiDetails`'s return — `readContextFromUrl` and `slugify` then have no callers and should go too.

**Prompt for Claude**
```text
In portals/api-control-plane/src/pages/appShell/appShellPages/apis/create/:
1. Confirm the dead path: utils/specDetails.ts extractApiDetails returns a `context` key derived from
   the server URL (readContextFromUrl) or a title slug, and components/GeneralCreateApiForm.tsx
   getInitialValues sets `context: toBasePath(projectHandler, id, merged.version)` unconditionally,
   ignoring merged.context. If getInitialValues already honours draftData.context, stop — I misread.
2. If confirmed, pick ONE and apply it:
   (a) platform base path always wins → delete `context` from extractApiDetails's return value and
       delete readContextFromUrl and slugify, which then have no callers; or
   (b) the document wins when it names a context → `draftData.context ?? toBasePath(...)`.
   Default to (a) unless the base path helper text is wrong.
3. Either way, add a unit test for extractApiDetails covering a server URL with a path, a relative
   server URL ("/v2"), and a templated URL.
```

portals/api-control-plane/src/pages/appShell/appShellPages/apis/create/ApiCreationWizard.tsx
```diff
@@ -0,0 +1,193 @@
...
+        return 'Bring an existing contract, or start from a blank slate and fill in the details yourself.';
+      case 'configure':
+        return `${API_TYPES.find((type) => type === apiType)?.rawTitle || 'API'}.DEFAULT SKELETON - EDIT IT OR ASK AI TO REFINE IT`;
+      default:
+        return '';
+    }
+  };
+
+  const handleDataExtracted = (data: ApiCreationWizardDraftState) => {
+    setPrefilledData((prev) => ({
```
AnuGayan: **Severity:** 🟡 minor
**Subject:** `prefilledData` merges across imports, so a second spec inherits the first one's fields

**Description**
`extractApiDetails` deliberately *omits* keys the document doesn't answer for — a documented design choice, so that merging "can only ever fill blanks". Spreading over `prev` here turns that omission into a leak across imports: the steps are navigable, so a user can import spec A (with a `description` and `operations`), click back to "Source", import spec B (with neither), and carry A's description and A's operation list into B's create request. Confirmed against the conditional spreads in `utils/specDetails.ts`.

**How to verify**
Import a spec with `info.description` set → Next → click the "Source" step in the stepper → import a spec with no description → Next. The Description field still holds the first spec's text.

**Suggested fix**
```suggestion
    setPrefilledData(data);
```

(dropping the updater form and the spread).

portals/api-control-plane/src/pages/appShell/appShellPages/apis/create/components/ContractSourceForm.tsx
```diff
@@ -0,0 +1,1717 @@
...
+    if (!response.ok) {
+      return { status: 'unreachable' };
+    }
+    text = await response.text();
+  } catch {
+    // Network failure, or the host refused the cross-origin read. The reason
+    // is developer-facing, so it stays in the console, not the UI.
+    return { status: 'unreachable' };
+  }
+  if (text.length > MAX_CONTRACT_BYTES) {
```
AnuGayan, line 808: **Severity:** 🟡 minor
**Subject:** The 10 MB contract ceiling is checked after the whole body is already buffered in memory

**Description**
`await response.text()` on the line above has already materialised the entire response, so this guard cannot prevent the tab freeze its constant's comment cites ("a huge document would freeze the tab"). Two secondary issues in the same function: `text.length` counts UTF-16 code units, not bytes, so `MAX_CONTRACT_BYTES` overstates what is enforced; and there is no `AbortSignal` and no timeout, so a spec host that accepts the connection and never answers leaves the Fetch button spinning with no way out. The file-upload path (`fetchContractForPreview`, case `'file'`) checks `file.size` *before* reading — that's the right shape.

**How to verify**
Point the URL source at a few-hundred-MB file on a CORS-open host and watch memory in the browser task manager: it climbs to the full size before the "too large" error appears. For the hang, point it at a host that accepts and stalls.

**Suggested fix**
Reject on `Content-Length` before reading, keep this check as a backstop for chunked responses, and give the request a deadline:

```ts
const response = await fetch(url, {
  headers: { Accept: 'application/json, application/yaml, text/yaml, text/plain' },
  signal: AbortSignal.timeout(CONTRACT_FETCH_TIMEOUT_MS),
});
if (!response.ok) return { status: 'unreachable' };
const declared = Number(response.headers.get('content-length'));
if (Number.isFinite(declared) && declared > MAX_CONTRACT_BYTES) {
  return { status: 'oversized' };
}
```

**Prompt for Claude**
```text
In portals/api-control-plane/src/pages/appShell/appShellPages/apis/create/components/ContractSourceForm.tsx,
fetchDocumentFrom checks `text.length > MAX_CONTRACT_BYTES` only after `await response.text()`.
1. Confirm that ordering by reading the function, and confirm the file path
   (fetchContractForPreview, case 'file') checks `values.file.size` BEFORE reading — that is the
   shape to match. If the URL path already guards before reading, stop and tell me.
2. If confirmed: add a Content-Length pre-check returning { status: 'oversized' } before the body is
   read, keep the existing post-read check as a backstop for responses with no Content-Length, and
   pass signal: AbortSignal.timeout(...) with a named constant so a stalled host cannot leave the
   Fetch button spinning. Verify a timeout maps to 'unreachable' via the existing catch.
3. Do not change the ContractFetchResult union or the 'oversized'/'unreachable' mapping in
   fetchErrorText.
4. Confirm with a unit test over fetchContractForPreview using a mocked fetch that returns a
   Content-Length above the ceiling, asserting no body read occurred.
```

portals/api-control-plane/tsconfig.json
```diff
@@ -16,12 +16,12 @@
...
     "noEmit": true,
     "jsx": "react-jsx",
-    "baseUrl": "./src",
     "paths": {
       "react": ["../node_modules/@types/react"],
       "react/jsx-runtime": ["../node_modules/@types/react/jsx-runtime"],
       "react-dom": ["../node_modules/@types/react-dom"],
-      "react-dom/client": ["../node_modules/@types/react-dom/client"]
+      "react-dom/client": ["../node_modules/@types/react-dom/client"],
+      "@/*": ["./src/*"]
```
AnuGayan, line 24: **Severity:** 🟡 minor
**Subject:** Removing `baseUrl: "./src"` silently broke the `react` type-dedup path mappings above

**Description**
The four `react`/`react-dom` entries were written against `baseUrl: "./src"`, resolving to `portals/api-control-plane/node_modules/@types/react`. TypeScript 5.8 (`package.json` pins `typescript: 5.8.3`) resolves `paths` without `baseUrl` **relative to the tsconfig's own directory**, so `../node_modules/...` now points at `portals/node_modules/@types/react` — and `portals/` has no `package.json` and no `node_modules` (it holds only `ai-workspace/ api-control-plane/ api-portal/ cloud-plugins/ scripts/`). The mappings therefore no longer map anything; TS falls back to ordinary resolution. Minor because it fails quietly today — but it re-surfaces as confusing duplicate-`@types/react` errors the next time a dependency ships its own copy. The `@/*` migration itself is complete: no `src`-rooted imports remain under `src/`.

**How to verify**
`cd portals/api-control-plane && npx tsc --noEmit --traceResolution 2>&1 | grep -B2 -A4 "Module name 'react'"` — the substitution names a `portals/node_modules` path that does not exist.

**Suggested fix**
Rewrite the values against the tsconfig directory rather than restoring `baseUrl` (which would re-enable the src-rooted imports this PR just removed). Replacing the whole `paths` body:

```jsonc
    "paths": {
      "react": ["./node_modules/@types/react"],
      "react/jsx-runtime": ["./node_modules/@types/react/jsx-runtime"],
      "react-dom": ["./node_modules/@types/react-dom"],
      "react-dom/client": ["./node_modules/@types/react-dom/client"],
      "@/*": ["./src/*"]
    },
```

**Prompt for Claude**
```text
In portals/api-control-plane/tsconfig.json, this PR removed "baseUrl": "./src" and added
"@/*": ["./src/*"], but left the four react/react-dom mappings as "../node_modules/...".
1. First confirm the breakage empirically. From portals/api-control-plane run:
   npx tsc --noEmit --traceResolution 2>&1 | grep -B2 -A4 "Module name 'react'"
   and check which path the `paths` substitution produces. If it resolves into
   portals/api-control-plane/node_modules (still correct), stop and tell me — nothing should change.
2. If it resolves into portals/node_modules (no package.json there), rewrite the four values as
   "./node_modules/@types/react" etc. — relative to the tsconfig dir, matching how TS 5.x resolves
   paths without baseUrl. Do NOT re-add baseUrl.
3. Verify no src-rooted imports crept back:
   grep -rE "from '(api|components|pages|hooks|scope|routes|config|theme|types|i18n|navigation|assets|test)/" src
   must return nothing.
4. Confirm with `npm run typecheck` and by re-running the traceResolution grep.
```

portals/api-control-plane/src/pages/appShell/appShellPages/gateways/utils/gatewaySetup.ts
```diff
@@ -0,0 +1,137 @@
...
+ */
+export const helmInstallCommand = (
+  target: GatewaySetupTarget,
+  releaseName: string,
+  token: string,
+): string =>
+  `helm install ${releaseName} ${HELM_CHART_REF} --version ${target.version} \\\n` +
+  `  --set gateway.controller.controlPlane.host="${target.controlPlaneHost}" \\\n` +
+  `  --set gateway.controller.controlPlane.port=${CONTROL_PLANE_PORT} \\\n` +
+  `  --set gateway.controller.controlPlane.token.value="${token}"`;
```
AnuGayan, line 137: **Severity:** 🟡 minor
**Subject:** The Helm command puts the gateway registration token in the user's shell history

**Description**
This embeds the registration token inline in the command the console tells the user to copy and paste. The docstring on `configureCommand` two functions above states the rationale for *not* doing this — "a heredoc rather than two `echo` lines … so the token never lands in the user's shell history as a bare argument" — so the Docker path takes the care the Helm path doesn't. An inline `--set` value is recorded in `~/.bash_history`/`~/.zsh_history`, visible in `ps` for the life of the command, and stored in the Helm release's values. `gatewaySetup.test.ts` asserts the command shape but not how the token travels.

**How to verify**
Read the two functions side by side in this file, or open a Kubernetes gateway's setup panel and read the generated Helm command — the token is in the copied text.

**Suggested fix**
Write the token to a values file with the same heredoc technique the Docker path already uses, then `helm install … -f gateway-values.yaml`:

```ts
export const helmValuesCommand = (target: GatewaySetupTarget, token: string): string =>
  `cat > gateway-values.yaml << 'VALUES'\n` +
  `gateway:\n  controller:\n    controlPlane:\n      host: ${target.controlPlaneHost}\n` +
  `      port: ${CONTROL_PLANE_PORT}\n      token:\n        value: ${token}\n` +
  `VALUES`;
```

**Prompt for Claude**
```text
In portals/api-control-plane/src/pages/appShell/appShellPages/gateways/utils/gatewaySetup.ts,
helmInstallCommand embeds the registration token as an inline --set argument.
1. Confirm the inconsistency: read configureCommand's docstring in the same file, which states the
   heredoc exists so "the token never lands in the user's shell history as a bare argument". Confirm
   helmInstallCommand does exactly that. Also check whether the gateway Helm chart accepts a values
   file (-f) or --set-file for controlPlane.token.value — if it only accepts --set, say so and stop.
2. If a values file works, split the Helm flow into a heredoc that writes gateway-values.yaml and a
   `helm install <release> <chart> --version <v> -f gateway-values.yaml`, mirroring the Docker
   configure/start split. Keep the release name = gateway handle behaviour.
3. Update gatewaySetup.test.ts and any GatewaySetupBanner / GatewayGetStartedPanel step list that
   renders the command sequence — the step count changes.
4. Confirm with `npm test -- gatewaySetup` and by reading the rendered Kubernetes setup panel.
```

portals/api-control-plane/src/pages/appShell/appShellPages/apis/create/ApiCreationWizard.tsx
```diff
@@ -0,0 +1,193 @@
...
+            <ApiTypeSelector
+              onChange={(apiType) => {
+                setApiType(apiType);
+                setStep('source');
+              }}
+            />
+          )}
+
+          {step === 'source' && (
+            <DefineApiPanel
```
AnuGayan, line 235: **Severity:** 🟡 minor
**Subject:** Step 1's API-type choice never reaches step 2

**Description**
`DefineApiPanel` accepts `initialApiTypeKey` and threads it into `ContractSourceForm`, which uses it to pick the available sources and accepted file extensions (`sourcesFor`/`extensionsFor`). It isn't passed here, so the contract step always falls back to `CONTRACT_API_TYPES[0]?.key`. Harmless today — REST is the only enabled type — but it means the `apiType` state captured in step 1 feeds nothing except a title lookup, so the wiring is inert and the bug stays invisible until a second type is enabled.

**How to verify**
Read `DefineApiPanelProps.initialApiTypeKey` and its default in `ContractSourceForm`, then compare with the props passed at this call site.

**Suggested fix**
```suggestion
            <DefineApiPanel
              initialApiTypeKey={apiType?.key}
```

portals/api-control-plane/src/pages/appShell/appShellPages/apis/create/components/ContractSourceForm.tsx
```diff
@@ -0,0 +1,1717 @@
...
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [repositoryKey, sourceKey]);
+
+  /**
+   * The branch's tree, in one request. Loaded as soon as a branch is known
+   * rather than when the dialog opens, so the root directory can be checked for
+   * a contract — and its file pre-selected — without anyone having to browse.
+   */
+  useEffect(() => {
+    if (sourceKey !== 'github' || repositoryRef === null || branch === '') {
```
AnuGayan, line 1035: **Severity:** 🔵 nit
**Subject:** The GitHub effects don't reset derived state when the repository changes

**Description**
Swapping repo A for repo B leaves `branch`, `tree`, `directory` and `contractFile` from A in place, so this effect fires one request for `repo B @ branch A` before the lookup effect corrects `branch` — a wasted call against the 60-per-hour anonymous budget `utils/github.ts` itself flags as scarce, plus a browse dialog briefly showing A's tree under B's name. Separately, this early-return branch has no cleanup, so `treeLoading` can stick at `true` when the guard trips mid-load (the previous run's `controller.abort()` makes the `.then` return before `setTreeLoading(false)`). Not observable today because the spinner only renders while `repoLookup.status === 'found'`, but it's a missing reset either way.

**How to verify**
Type a valid repo URL, wait for the branch to resolve, then paste a different repo whose default branch differs — the network panel shows a tree request against the *old* branch before the corrected one.

**Suggested fix**
Reset `branch`, `tree`, `directory` and `contractFile` when `repositoryKey` changes (a small effect keyed on it alone, or fold the reset into the lookup effect), and add `setTreeLoading(false)` to this early-return path.

portals/api-control-plane/vite.config.ts
```diff
@@ -50,6 +50,11 @@ export default ({ mode }: { mode: string }) => {
       }),
       basicSsl(),
     ],
+    resolve: {
+    alias: {
+      '@': '/src',
```
AnuGayan: **Severity:** 🔵 nit
**Subject:** The new `resolve.alias` block is mis-indented, has trailing whitespace, and is outside CI's reach

**Description**
The block's indentation doesn't match the surrounding object, and there is trailing whitespace after `'@': '/src', `. Neither `lint` (`files: ['src/**/*']` in `eslint.config.js`) nor `format:check` (`"src/**/*.{ts,tsx,js,jsx,json,css,md}"`) covers `vite.config.ts`, so CI won't flag either. The root-relative string does work in both dev and build, but `path.resolve` is the form that doesn't depend on Vite's root handling.

**How to verify**
`npx prettier --check vite.config.ts` — fails, while `npm run format:check` passes.

**Suggested fix**
```suggestion
      '@': path.resolve(__dirname, 'src'),
```

with `import path from 'node:path'` at the top, and re-indent the enclosing `resolve`/`alias` braces to match the neighbouring keys. Worth widening `format:check`/`lint` to the config files while you're here.

portals/api-control-plane/src/pages/appShell/appShellPages/apis/create/uiConfig.tsx
```diff
@@ -0,0 +1,151 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import { Box } from '@wso2/oxygen-ui';
+import {} from '@wso2/oxygen-ui-icons-react';
```
AnuGayan: **Severity:** 🔵 nit
**Subject:** Empty import, and `rawTitle`/`rawDescription` duplicate the message descriptors below

**Description**
`import {} from '@wso2/oxygen-ui-icons-react';` imports nothing and can go. Separately, each `API_TYPES` entry carries `rawTitle`/`rawDescription` alongside the `defineMessages` `defaultMessage`s two lines above — two sources of truth for the same copy, and they have already drifted: `grpcDescription` reads "Expose high-performance gRPC services." while `rawDescription` reads "Expose gRPC services through the gateway."

**How to verify**
Compare `messages.grpcDescription.defaultMessage` with the `grpc` entry's `rawDescription` in this file.

**Suggested fix**
```suggestion
```

(delete the line). For the duplication, derive the raw strings from the descriptors (`messages.grpcTitle.defaultMessage`) or drop them entirely once the wizard's copy goes through `intl` — see the comment on `ApiCreationWizard.getTitleForStep`.

portals/api-control-plane/src/pages/appShell/appShellPages/apis/create/utils/github.ts
```diff
@@ -0,0 +1,307 @@
...
+  const segments = url.pathname.split('/').filter((segment) => segment !== '');
+  const [owner, rawRepo, kind, branch, ...rest] = segments;
+  if (owner === undefined || rawRepo === undefined) {
+    return null;
+  }
+
+  const repo = rawRepo.replace(/\.git$/, '');
+  // Only `tree` carries a branch and directory; `blob`, `commit` and the rest
+  // name something this step can't start from, so they are ignored.
+  if (kind !== 'tree' || branch === undefined) {
```
AnuGayan: **Severity:** 🔵 nit
**Subject:** `parseGitHubRepoUrl` mis-splits branch names containing a slash

**Description**
A URL like `github.com/org/repo/tree/feature/x/apis` is read as branch `feature` with path `x/apis`, because the segment after `tree` is taken as the whole branch name. GitHub's URL format is genuinely ambiguous here — `tree/<ref>/<path>` gives no delimiter — so this isn't solvable from the string alone. The cost is a failed tree fetch (and one spent request from the 60/hour anonymous budget) for anyone pasting a deep link on a slash-named branch, which is a common convention.

**How to verify**
Paste `https://github.com/<org>/<repo>/tree/feature/some-branch` for a repo that has such a branch — the branch select lands on the default branch instead, and the directory pre-selection is wrong.

**Suggested fix**
The lookup already fetches the branch list. Resolve the ambiguity against it: try progressively longer prefixes of `[branch, ...rest]` joined by `/` and keep the longest that appears in `repository.branches`, treating the remainder as the path. That needs the ref resolved after `lookupGitHubRepository` rather than at parse time, so it's a change to how the effect consumes this — worth a follow-up issue rather than an inline fix.

portals/api-control-plane/src/pages/appShell/appShellPages/apis/create/components/ApiResourcesPreview.tsx
```diff
@@ -0,0 +1,167 @@
...
+  components: { Topbar: () => null },
+});
+
+const HideInfoPlugin = () => ({
+  components: { info: () => null },
+});
+
+export type ApiResourcesPreviewProps = {
+  /**
+   * The fetched definition. Swagger UI renders it as-is and fetches nothing,
```
AnuGayan: **Severity:** 🔵 nit
**Subject:** "renders it as-is and fetches nothing" may not hold for specs with external `$ref`s

**Description**
`utils/specValidation.ts` classifies external `$ref`s as a *warning*, so a document containing `$ref: https://…` is accepted and handed straight to `<SwaggerUI spec={spec} />`. swagger-client's resolver generally does fetch remote refs, which would contradict this comment and mean the preview makes network requests to whatever host the spec names. Unverified — I read the consumer, not swagger-client's behaviour under the `spec` prop at this version — so this is a claim to check, not a defect I've confirmed.

**How to verify**
Feed the wizard a spec whose `components.schemas` uses `$ref: https://example.com/schema.json` and watch the network panel while the preview renders.

**Suggested fix**
Either confirm the claim and leave the comment, or soften it to say what is actually guaranteed (Swagger UI is given a parsed object rather than a URL, and try-it-out is disabled) and note that remote `$ref`s may still be resolved. If they are and that's unwanted, promote `externalRefs` from warning to error, or strip them before handing the spec over.

portals/api-control-plane/src/pages/appShell/appShellPages/apis/create/components/ApiCreationProgress.tsx
```diff
@@ -0,0 +1,381 @@
...
+  onBack,
+  onComplete,
+  onRetry,
+  status,
+}: ApiCreationProgressProps) => {
+  const intl = useIntl();
+  const [percent, setPercent] = useState(0);
+
+  useEffect(() => {
+    if (status !== 'creating') return undefined;
```
AnuGayan, line 278: **Severity:** 🔵 nit
**Subject:** `percent` isn't reset when a retry restarts the flow

**Description**
`percent` is only ever raised — toward `PENDING_CEILING` while creating, then snapped to 100 on success. On the failure → retry path `status` goes `failed` → `creating` again with `percent` still at ~92, so the second attempt opens the ring near-full and immediately reads "Finalizing setup". Cosmetic, but it undercuts the honesty the module's own comment is built around ("it never claims 'done' before it is").

**How to verify**
Force a create failure, click "Try again", and read the ring: it starts at ~92%, not 0%.

**Suggested fix**
```suggestion
    if (status !== 'creating') return undefined;
    setPercent(0);
```

That resets on entry to every `creating` phase, including the first, where it's already 0.

ShavinAnjithaAlpha: Pushed a fix in commit [4703281](https://github.com/wso2/api-platform/pull/3338/commits/47032810c905b8bc698887ecbca8c2f7ab4fa14e).

portals/api-control-plane/src/pages/appShell/appShellPages/apis/create/components/ApiResourcesPreview.tsx
ShavinAnjithaAlpha: Pushed a fix in commit [4703281](https://github.com/wso2/api-platform/pull/3338/commits/47032810c905b8bc698887ecbca8c2f7ab4fa14e).

portals/api-control-plane/src/pages/appShell/appShellPages/apis/create/uiConfig.tsx
ShavinAnjithaAlpha: Pushed a fix in commit [4703281](https://github.com/wso2/api-platform/pull/3338/commits/47032810c905b8bc698887ecbca8c2f7ab4fa14e).

portals/api-control-plane/src/pages/appShell/appShellPages/apis/create/ApiCreationWizard.tsx
ShavinAnjithaAlpha: Pushed a fix in commit [4703281](https://github.com/wso2/api-platform/pull/3338/commits/47032810c905b8bc698887ecbca8c2f7ab4fa14e).

portals/api-control-plane/tsconfig.json
ShavinAnjithaAlpha: Pushed a fix in commit [4703281](https://github.com/wso2/api-platform/pull/3338/commits/47032810c905b8bc698887ecbca8c2f7ab4fa14e).

portals/api-control-plane/src/pages/appShell/appShellPages/apis/create/components/ContractSourceForm.tsx
ShavinAnjithaAlpha: Pushed a fix in commit [4703281](https://github.com/wso2/api-platform/pull/3338/commits/47032810c905b8bc698887ecbca8c2f7ab4fa14e).

portals/api-control-plane/src/pages/appShell/appShellPages/apis/create/ApiCreationWizard.tsx
ShavinAnjithaAlpha: Pushed a fix in commit [4703281](https://github.com/wso2/api-platform/pull/3338/commits/47032810c905b8bc698887ecbca8c2f7ab4fa14e).

portals/api-control-plane/src/pages/appShell/appShellPages/apis/create/components/GeneralCreateApiForm.tsx
ShavinAnjithaAlpha: Pushed a fix in commit [4703281](https://github.com/wso2/api-platform/pull/3338/commits/47032810c905b8bc698887ecbca8c2f7ab4fa14e).

portals/api-control-plane/src/pages/appShell/appShellPages/apis/create/ApiCreationWizard.tsx
ShavinAnjithaAlpha: Pushed a fix in commit [4703281](https://github.com/wso2/api-platform/pull/3338/commits/47032810c905b8bc698887ecbca8c2f7ab4fa14e).

portals/api-control-plane/src/pages/appShell/appShellPages/apis/create/ApiCreationWizard.tsx
ShavinAnjithaAlpha: Pushed a fix in commit [4703281](https://github.com/wso2/api-platform/pull/3338/commits/47032810c905b8bc698887ecbca8c2f7ab4fa14e).

portals/api-control-plane/src/pages/appShell/appShellPages/apis/create/ApiCreationWizard.tsx
ShavinAnjithaAlpha: Pushed a fix in commit [4703281](https://github.com/wso2/api-platform/pull/3338/commits/47032810c905b8bc698887ecbca8c2f7ab4fa14e).

portals/api-control-plane/src/pages/appShell/appShellPages/apis/create/ApiCreationWizard.tsx
ShavinAnjithaAlpha: Pushed a fix in commit [4703281](https://github.com/wso2/api-platform/pull/3338/commits/47032810c905b8bc698887ecbca8c2f7ab4fa14e).

portals/api-control-plane/src/pages/appShell/appShellPages/apis/create/components/ContractSourceForm.tsx
ShavinAnjithaAlpha: Pushed a fix in commit [4703281](https://github.com/wso2/api-platform/pull/3338/commits/47032810c905b8bc698887ecbca8c2f7ab4fa14e).

portals/api-control-plane/src/pages/appShell/appShellPages/apis/overview/ApiKeysPanel.tsx
ShavinAnjithaAlpha: Pushed a fix in commit [4703281](https://github.com/wso2/api-platform/pull/3338/commits/47032810c905b8bc698887ecbca8c2f7ab4fa14e).

portals/api-control-plane/src/pages/appShell/appShellPages/apis/create/components/ContractSourceForm.tsx
ShavinAnjithaAlpha: Pushed a fix in commit [4703281](https://github.com/wso2/api-platform/pull/3338/commits/47032810c905b8bc698887ecbca8c2f7ab4fa14e).

portals/api-control-plane/src/pages/appShell/appShellPages/apis/create/utils/github.ts
ShavinAnjithaAlpha: Pushed a fix in commit [4703281](https://github.com/wso2/api-platform/pull/3338/commits/47032810c905b8bc698887ecbca8c2f7ab4fa14e).

portals/api-control-plane/vite.config.ts
ShavinAnjithaAlpha: Pushed a fix in commit [4703281](https://github.com/wso2/api-platform/pull/3338/commits/47032810c905b8bc698887ecbca8c2f7ab4fa14e).

portals/api-control-plane/src/pages/appShell/appShellPages/gateways/utils/gatewaySetup.ts
ShavinAnjithaAlpha: Not taking this here. These Helm commands are ported verbatim from Bijira and must stay consistent across both portals. The token exposure concern is valid, but any change should happen upstream and land in both.

Separately, `--set gateway.controller.controlPlane.port=443` appears to be a no-op; the chart only uses host and token.*.

Induwara04 (dismissed):

lasanthaS (approved):

PR #3341, opened 2026-09-01
Enhance role-based access control by adding 'ap_developer' role with scopes

Thushani-Jayasekera:
Enhance role-based access control by adding 'ap_developer' role with scopes comprehensive API management scopes. Update documentation to reflect the new role and its permissions in the role-to-scope mapping.


platform-api/resources/role-to-scope-mapping.yaml
```diff
@@ -103,7 +103,63 @@ roles:
       - dp:label:manage
       - dp:webhook_subscriber:manage
       - dp:event:read
-
+  - name: ap_developer
```
renuka-fernando, line 110: Shall we add a comment on what role this is, the same as other roles?

platform-api/resources/role-to-scope-mapping.yaml
```diff
@@ -104,6 +104,44 @@ roles:
...
+      - ap:mcp_proxy:manage
+      - ap:llm_proxy:manage
+      - ap:llm_provider:read
+      - ap:llm_template:read
+      - ap:gateway:read
+      - ap:gateway:manifest:read
+      - ap:gateway_custom_policy:read
+      - ap:secret:create
+      - ap:secret:read
+      - ap:secret:update
```
renuka-fernando, line 125: Shall we just provide `ap:secret:read` only?

PR #3342, opened 2026-09-01
Add cloud Moesif Insights plugin for API Control Plane

Milanka00:
## Purpose

- Add `@wso2-enterprise/apip-cloud-ui-insights`, a shared cloud plugin that embeds Moesif Insights via wrap/basic iframes for **organization** and **project** scope in API Control Plane.
- Register org/project Insights sidebar extensions in `apip-cloud-ui` (`organization-insights`, `project-insights`) with Moesif token fetch via BFF (`GET /proxy/cloud/analytics/id-token`).
- Hide the built-in Insights nav item outside API scope when cloud Insights extensions are loaded, while keeping the built-in API Insights submenu at API scope (Coming soon overlay).
- Same approach has been used for ai-workspace

## Related issues

- https://github.com/wso2-enterprise/apim-saas/issues/2893


Induwara04 (dismissed):

Milanka00:
@coderabbitai approve

portals/cloud-plugins/apip-cloud-ui-insights/src/api/analyticsApi.ts
```diff
@@ -0,0 +1,154 @@
...
+      accept: 'application/json',
+      ...(init?.headers ?? {}),
+    },
+    ...init,
+  });
+  return readJson<T>(response);
+};
+
+/** GET /cloud/analytics/id-token — Moesif dashboard-viewer token for the caller org. */
+export async function fetchViewerToken(): Promise<string> {
```
AnuGayan, line 74: **Severity:** 🔴 blocker
**Subject:** `GET /cloud/analytics/id-token` exists in neither repo, and `/proxy/cloud/*` isn't routed

**Description**
This is the single gate on the whole embed — no token, no iframe. The README says it maps to `wso2cloud/backend/core/internal/moesifmapping/handler/handler.go`. That package doesn't exist at `wso2-enterprise/wso2cloud@bb2b7530`, and no file in either repo contains the string `analytics/id-token` (the only Moesif code in wso2cloud is `backend/billing/internal/integration/moesif`, a metering client). Separately, the BFF mounts `/proxy` → platform-api and `/proxy/<name>` only for *configured* named upstreams (`bff/internal/server/server.go:94,105-117`); this PR adds no `cloud` upstream, so the call falls through to platform-api, which has no such route either. Every org/project Insights page lands on `ErrorState`. Confirmed against `bff/internal/server/server.go:89-118` and greps over both repos at those refs.

**How to verify**
```
cd wso2cloud && git fetch upstream main && git grep -In "moesifmapping" upstream/main   # no hits
git grep -In "analytics/id-token" 54f394d4                                              # no hits
```

**Suggested fix**
Land (or link) the endpoint and the BFF upstream config before merging. If the backend is in flight elsewhere, gate registration on a runtime flag so the sidebar entry doesn't appear against a deployment that can't serve it — `billingProxyEnabled` in `bff/internal/config/runtime_config.go:39-47` is the existing precedent. Not validated against a running deployment; this is a "referenced thing is absent" finding, not a proposed code change.

**Prompt for Claude**
```text
In portals/cloud-plugins/apip-cloud-ui-insights/src/api/analyticsApi.ts (fetchViewerToken)
and its README.md:
1. FIRST verify the finding. Search wso2-enterprise/wso2cloud@main for
   `backend/core/internal/moesifmapping` and for the string "analytics/id-token", and search
   api-platform for the same string. If EITHER produces a real handler serving
   GET /cloud/analytics/id-token, stop and tell me - the finding is wrong.
2. Also check portals/api-control-plane/bff/internal/server/server.go and bff/internal/config
   for whether a named upstream called "cloud" is configured or documented anywhere.
   /proxy/<name> routing only exists for configured upstreams.
3. If the endpoint really is absent: do NOT invent it. Instead make the plugin's registration
   conditional, mirroring the existing billingProxyEnabled pattern in
   bff/internal/config/runtime_config.go:39-47, so the Insights sidebar entry is only
   registered when the deployment advertises the analytics upstream.
4. Must not break: the ai-workspace host registration, or the behaviour where built-in
   Insights stays visible at API scope.
5. Confirm with a test: with the flag absent, cloudExtensions contains no
   'organization-insights'/'project-insights' entry and useNavigationItems still returns the
   built-in 'insights' item at org scope.
```

portals/cloud-plugins/apip-cloud-ui-insights/src/config/runtimeConfig.ts
```diff
@@ -0,0 +1,97 @@
...
+    config?: LegacyWindowConfig;
+    __RUNTIME_CONFIG__?: LegacyWindowConfig;
+  };
+  return {
+    ...(globalWindow.__RUNTIME_CONFIG__ ?? {}),
+    ...(globalWindow.config ?? {}),
+  };
+};
+
+const isProductionEnvironment = () => {
```
AnuGayan: **Severity:** 🔴 blocker
**Subject:** Production silently resolves to `web-dev.moesif.com`

**Description**
`isProductionEnvironment()` reads `environmentName` off `window.__RUNTIME_CONFIG__`, and `configuredMoesifAppUrl()` reads four Moesif aliases off the same object. The ACP BFF emits **exactly three** runtime-config keys — `authMode`, `platformApiBaseUrl`, and conditionally `billingProxyEnabled` (`bff/internal/config/runtime_config.go:31-48`). Neither `environmentName` nor any Moesif key is among them, and this PR adds none. So in any BFF-served deployment `configuredMoesifAppUrl()` is `''` and this falls to the `'local'` default → `https://web-dev.moesif.com`: production analytics rendered from Moesif's dev environment, with the org's viewer token posted to it. The only escape is the build-time `VITE_ENVIRONMENT_NAME`, which no build config here sets. It fails silently — the allowlist accepts web-dev, so nothing logs or errors. Confirmed against `bff/internal/config/runtime_config.go:31-48`.

**How to verify**
Read `buildRuntimeConfig` in `bff/internal/config/runtime_config.go` — the returned map has no `environmentName`. On a BFF-served build, evaluate `window.__RUNTIME_CONFIG__` in the console; the iframe `src` will be `https://web-dev.moesif.com/wrap/basic#auth=post`.

**Suggested fix**
Both halves are needed: have the BFF emit the Moesif host explicitly (same way `platformApiBaseUrl` is force-set there), and make the client fail loudly rather than guess — render `ErrorState` when no allowlisted host is configured. An environment-sniffing default can only be wrong in the direction that points production at the wrong Moesif tenant.

**Prompt for Claude**
```text
In portals/cloud-plugins/apip-cloud-ui-insights/src/config/runtimeConfig.ts:
1. FIRST verify: read portals/api-control-plane/bff/internal/config/runtime_config.go and list
   every key buildRuntimeConfig puts into the map. If "environmentName" or any moesif* key IS
   emitted there, stop and tell me - the finding is wrong.
2. If they are absent, make a missing Moesif host an explicit error rather than a silent
   fallback: let insightsRuntimeConfig.moesifAppUrl be `string | undefined` when nothing
   allowlisted is configured, and have InsightsEmbed render ErrorState ("Insights is not
   configured for this deployment") instead of building an iframe src. Keep
   resolveTrustedMoesifAppUrl's https + allowlist check exactly as it is.
3. Separately add the runtime-config key on the BFF side in buildRuntimeConfig so a real
   deployment can set it, following the platformApiBaseUrl precedent in that same function.
4. Must not break: the moesifEmbed.test.ts allowlist tests, or the InsightsEmbed test that
   mocks insightsRuntimeConfig with an explicit moesifAppUrl.
5. Confirm with a new test: with no window config and no VITE_* env, the embed renders the
   not-configured error and never produces a web-dev.moesif.com iframe src.
```

AnuGayan (commented):
Reviewed against PR head `54f394d4`, plus `wso2-enterprise/wso2cloud@bb2b7530` (upstream main, freshly fetched) for the backend contract the README cites.

The cross-origin handshake is done carefully — https+allowlist on the configured origin, serialized `targetOrigin` on send, and **both** `event.origin` and `event.source === iframe.contentWindow` on receive. Server-side token minting is the right call. The "stale project metadata when switching projects" test targets exactly the bug this component shape invites.

My concern is that the two external contracts the embed depends on don't resolve against anything that exists today — see the two 🔴 inline comments. Everything else is small.

**Also worth a look (not inlined):**
- `readJson` turns *any* 404 into "Insights is not available for this organization", but it's shared by the token call and both project lookups — a missing project reports an org entitlement problem.
- `resolveProjectScope`'s fallback reads only `response.list`, but that response is `{ list, count, pagination }` (`projectClient.ts:57-64`) — breaks past page one. Its `catch {}` also discards 401/403/500 identically to 404.
- ~6,800 lines of `package-lock.json` across two new files. No sibling cloud plugin commits one, and `apip-cloud-ui` had none before this PR — looks like incidental `npm install` output.
- All user-facing strings are inline English, while this PR's own `overlays/api-control-plane/InsightsPage.tsx` uses `FormattedMessage`.
- The body's 4th bullet (`ApiCreationWizard` → `ApiCreatePage`) isn't in this diff — `AppRoutes.tsx` isn't among the 27 files and the identifier doesn't exist at head. The README also documents an `ai-workspace.tsx` registration this PR doesn't add.
- Nits: `embedAttempt` is listed as a dep of the handshake-timeout effect but never read (`handshakeError` already re-triggers it); `retryHandshake` remounts with the *same* token, so a token-rejection stall can only fail again; the two new nav tests shadow the file-level `atOrg`/`atApi` helpers at `useNavigationItems.test.tsx:52-65` and duplicate the extension fixture.

Happy to be wrong on the two blockers if there's a companion backend PR or a cloud-specific BFF build I haven't seen — point me at it and I'll re-check.

portals/cloud-plugins/apip-cloud-ui-insights/src/api/analyticsApi.ts
```diff
@@ -0,0 +1,154 @@
...
+  handler?: string;
+  handle?: string;
+  name?: string;
+  displayName?: string;
+};
+
+const asRecord = (value: unknown): Record<string, unknown> =>
+  value && typeof value === 'object' ? (value as Record<string, unknown>) : {};
+
+const pickProjectHandle = (project: ProjectRecord) =>
```
AnuGayan, line 96: **Severity:** 🟠 major
**Subject:** Matches on `handler`, but the URL segment is the project `id`

**Description**
`resolveProjectScope` re-validates its own successful lookup with `pickProjectHandle(project) === trimmedHandle`, and this prefers `handler → handle → id`. The core treats that same URL segment as the **id**: `ConsoleScopeProvider.tsx:131` matches `item.id === params.projectHandler`, and `projectClient.ts:104` documents "projects are UUID-keyed and the URL handler is that UUID". `toProject` (`api/adapters.ts:164-192`) explicitly models `handler` as a field distinct from `id`, and the GraphQL selection at `projectClient.ts:80-97` selects both separately. So when the payload carries a `handler` different from `id`, the direct lookup is discarded, the list-scan fallback misses the same way, and the user gets `Project "<uuid>" was not found`. Both new tests use payloads with no `handler` — exactly the shape that passes.

**How to verify**
Add a case to the `analyticsApi` block in `moesifEmbed.test.ts` returning `{ id: 'proj-uuid', handler: 'orders', uuid: 'proj-uuid', displayName: 'Orders' }` for `/projects/proj-uuid`; `resolveProjectScope('default', 'proj-uuid')` rejects with "was not found".

**Suggested fix**
```suggestion
const pickProjectHandle = (project: ProjectRecord) =>
  project.id?.trim() ||
  project.handler?.trim() ||
  project.handle?.trim() ||
  '';
```
Not validated against a live platform-api payload — what's confirmed is the divergence from the core's own matching rule, not the exact field set the server returns.

**Prompt for Claude**
```text
In portals/cloud-plugins/apip-cloud-ui-insights/src/api/analyticsApi.ts, pickProjectHandle /
resolveProjectScope:
1. FIRST confirm what the URL's :projectHandler segment actually is. Read
   portals/api-control-plane/src/scope/ConsoleScopeProvider.tsx:131 and
   portals/api-control-plane/src/api/projects/projectClient.ts:100-118. If the core matches the
   route segment against `handler` rather than `id`, stop and tell me - the finding is inverted.
2. If the core matches on `id` (expected), reorder pickProjectHandle to prefer project.id
   before handler/handle so the guard agrees with the core's convention. Leave pickProjectId
   (uuid first) alone - that is the Moesif project_id, a different concern.
3. Must not break: the two existing resolveProjectScope tests in src/utils/moesifEmbed.test.ts.
4. Add a regression test with a payload where handler and id differ
   ({ id: 'proj-uuid', handler: 'orders', uuid: 'proj-uuid', displayName: 'Orders' }) and assert
   resolveProjectScope('default', 'proj-uuid') resolves rather than throwing.
```

portals/cloud-plugins/apip-cloud-ui-insights/src/api/analyticsApi.ts
```diff
@@ -0,0 +1,154 @@
...
+  url: string,
+  init?: RequestInit
+): Promise<T> => {
+  const response = await fetch(url, {
+    credentials: 'include',
+    headers: {
+      accept: 'application/json',
+      ...(init?.headers ?? {}),
+    },
+    ...init,
```
AnuGayan: **Severity:** 🟡 minor
**Subject:** `...init` spread after `headers` discards the merged defaults

**Description**
`...init` comes last and `init` carries `headers`, so the carefully merged `{ accept, ...init.headers }` object is replaced wholesale. `resolveProjectScope`'s two calls therefore go out with only `X-Org-Id` and no `Accept: application/json`. The core's own client sets `Accept` on every request (`api/core/http.ts:87`). Confirmed by reading `fetchJson`.

**How to verify**
In the `resolveProjectScope` tests, assert on the second `fetch` argument: `headers` is `{ 'X-Org-Id': 'default' }` with no `accept` key.

**Suggested fix**
```suggestion
    ...init,
    credentials: 'include',
    headers: {
      accept: 'application/json',
      ...(init?.headers ?? {}),
    },
```
(and drop the now-duplicated `credentials`/`headers` entries above)

**Prompt for Claude**
```text
In portals/cloud-plugins/apip-cloud-ui-insights/src/api/analyticsApi.ts, fetchJson (~line 58):
1. FIRST confirm the bug: the object literal passed to fetch spreads ...init AFTER the headers
   key, so a caller-supplied headers object replaces the merged one instead of extending it. If
   ...init is already spread first, stop - the finding is wrong.
2. If confirmed, reorder so ...init is spread first and credentials/headers are set after it,
   keeping the accept default merged under any caller headers.
3. Must not break: the fetchViewerToken test asserting fetch is called with
   objectContaining({ credentials: 'include' }).
4. Add an assertion that a resolveProjectScope call sends BOTH X-Org-Id and accept.
```

portals/api-control-plane/src/navigation/useNavigationItems.ts
```diff
@@ -86,7 +115,11 @@ export const useNavigationItems = (): NavigationItem[] => {
             })
           : undefined;
         return {
-          group: extension.group,
+          group:
+            CLOUD_INSIGHTS_SIDEBAR_IDS.has(extension.id) &&
```
AnuGayan: **Severity:** 🟡 minor
**Subject:** The `group`/`order` override is a no-op

**Description**
This substitutes `builtinInsightsDefinition.group`/`.order` for the extension's own values. But built-in `insights` is `group: CLUSTER.api` (= `'api'`) and `order: 60` (`navigationRegistry.tsx:60-66,367-370`), and both plugin extensions already declare `group: 'api', order: 60` in `hosts/api-control-plane.tsx`. Neither branch can change any value it computes. What it does do is hardcode two cloud plugin ids into the core nav pipeline — the thing the extension architecture exists to avoid — for zero effect. `isBuiltinInsightsHiddenByCloudPlugin` below does real work and should stay.

**How to verify**
Compare `navigationRegistry.tsx:367-370` against the two `defineCloudPlugin` entries in `portals/cloud-plugins/apip-cloud-ui/src/hosts/api-control-plane.tsx` — `group` and `order` are identical on both sides. Deleting both ternaries leaves every existing test green.

**Suggested fix**
Drop the two ternaries (use `extension.group` / `extension.order` directly) and `builtinInsightsDefinition`. If core should really own this, a declarative extension property (`replacesBuiltin: 'insights'`) would let core stop knowing plugin ids entirely — including in `CLOUD_INSIGHTS_SIDEBAR_IDS`.

**Prompt for Claude**
```text
In portals/api-control-plane/src/navigation/useNavigationItems.ts, the group (~line 119) and
order (~line 135) ternaries keyed on CLOUD_INSIGHTS_SIDEBAR_IDS:
1. FIRST verify they are no-ops. Read the `insights` entry in navigationRegistry.tsx (its group
   and order) and the two extension entries in
   portals/cloud-plugins/apip-cloud-ui/src/hosts/api-control-plane.tsx. If group or order differ
   between the two sides, stop and tell me - the override is load-bearing and the finding is wrong.
2. If they are identical, replace both ternaries with plain extension.group / extension.order and
   delete the now-unused builtinInsightsDefinition lookup.
3. Do NOT touch isBuiltinInsightsHiddenByCloudPlugin or CLOUD_INSIGHTS_SIDEBAR_IDS - the
   suppression is real behaviour covered by tests.
4. Confirm by running the api-control-plane nav tests; both new Insights tests in
   useNavigationItems.test.tsx (including the insights-before-observability ordering assertion)
   must still pass.
```

portals/cloud-plugins/apip-cloud-ui-insights/src/api/analyticsApi.ts
Milanka00: A PR is opened at WSO2 Cloud repository where these services are implemented.

portals/cloud-plugins/apip-cloud-ui-insights/src/InsightsEmbed.tsx
```diff
@@ -0,0 +1,345 @@
...
+      embeddingOrigin
+    );
+  }, [embeddingOrigin, viewerToken]);
+
+  useEffect(() => {
+    if (!isIframeDomLoaded || !viewerToken || isEmbedReady) return;
+    sendTokenToChild();
+    const intervalId = window.setInterval(sendTokenToChild, SET_TOKEN_RETRY_MS);
+    return () => window.clearInterval(intervalId);
+  }, [isEmbedReady, isIframeDomLoaded, sendTokenToChild, viewerToken]);
```
AnuGayan, line 166: **Severity:** 🟠 major
**Subject:** A refreshed viewer token is never posted to the Moesif iframe once the handshake completes

**Description**
This effect is the only code path that calls `sendTokenToChild()` on a schedule, and it returns early when `isEmbedReady` is true. `isEmbedReady` is set by the handshake and never cleared again except on `iframeSrc`/`scope.level` change or an explicit retry. So both renewal paths — the 50-minute `setInterval` (L136-146) and the `REFRESH_TOKEN` handler (L207-211) — mint a fresh token, call `setViewerToken`, and then nothing delivers it to the child frame. `onLoad` is the only other caller and does not fire again. A dashboard left open past the ~1h viewer-token TTL silently stops working inside the iframe, with no error in the console shell. Confirmed by reading this file: `child.postMessage` appears exactly once, inside `sendTokenToChild`.

**How to verify**
Set `VIEWER_TOKEN_REFRESH_INTERVAL_MS` to `5000` locally, open Insights, wait for the iframe to appear, and watch the frame's message traffic in devtools — no `SET_TOKEN` is posted after the first handshake even though `fetchViewerToken` is re-called. Or add a test: render, dispatch `SCHEMA_GEN_FINISHED`, dispatch a `REFRESH_TOKEN` message, and assert `postMessage` was called again — it is not.

**Suggested fix**
```suggestion
  useEffect(() => {
    if (!isIframeDomLoaded || !viewerToken) return;
    // Deliver the current token immediately — this also covers the periodic
    // refresh and REFRESH_TOKEN re-mints, which land after isEmbedReady.
    sendTokenToChild();
    if (isEmbedReady) return;
    const intervalId = window.setInterval(sendTokenToChild, SET_TOKEN_RETRY_MS);
    return () => window.clearInterval(intervalId);
  }, [isEmbedReady, isIframeDomLoaded, sendTokenToChild, viewerToken]);
```
This also stops the 400ms retry timer leaking forever once a handshake error unmounts the iframe.

**Prompt for Claude**
```text
In portals/cloud-plugins/apip-cloud-ui-insights/src/InsightsEmbed.tsx around lines 161-166.

1. FIRST verify the finding: grep the file for `postMessage`. Confirm it appears only inside
   `sendTokenToChild`, and that `sendTokenToChild` is called only (a) in the effect at ~161 and
   (b) in the iframe's `onLoad`. Confirm the effect at ~161 early-returns when `isEmbedReady`
   is true, and that `isEmbedReady` is only reset by the effect at ~98 (deps iframeSrc,
   scope.level) and by `retryHandshake`. If any other code path posts SET_TOKEN after the
   handshake, STOP and tell me — the finding is wrong.
2. If it holds, change the effect so `sendTokenToChild()` runs on every change of
   `viewerToken` / `isIframeDomLoaded`, and only the 400ms `setInterval` is gated on
   `!isEmbedReady`.
3. Do not change: the `embeddingOrigin` targetOrigin argument, the ALLOWED_MOESIF_ORIGINS
   check in the message listener, or the display:none-until-ready behaviour asserted by
   src/InsightsEmbed.test.tsx.
4. Add a test in src/InsightsEmbed.test.tsx: render, dispatch SCHEMA_GEN_FINISHED from
   https://web-dev.moesif.com, then dispatch a REFRESH_TOKEN message, and assert a second
   SET_TOKEN postMessage reaches the iframe's contentWindow.
5. Confirm with `npm test` in portals/cloud-plugins/apip-cloud-ui-insights.
```

portals/cloud-plugins/apip-cloud-ui-insights/tsconfig.json
```diff
@@ -0,0 +1,25 @@
...
+      ]
+    },
+    "jsx": "react-jsx",
+    "strict": true,
+    "noEmit": true,
+    "skipLibCheck": true,
+    "resolveJsonModule": true,
+    "types": ["vitest/globals", "@testing-library/jest-dom"]
+  },
+  "include": ["src"]
```
AnuGayan, line 24: **Severity:** 🟠 major
**Subject:** Dual-host plugin ships no `tsconfig.console.json`, so nothing typechecks it under API Control Plane's dependency versions

**Description**
`apip-cloud-ui-gateways` — the other plugin registered in both hosts — carries a `tsconfig.console.json` whose own comment states the reason: *"the console's image build is the only thing that ever typechecks this package … Run BOTH configs — clean under one host's dependency versions is not clean under the other's."* This package has only `tsconfig.json`, with `paths` pointed exclusively at `../../ai-workspace/node_modules`, yet it is now registered in `hosts/api-control-plane.tsx` as well. Nothing typechecks it against `portals/api-control-plane/node_modules`, so an oxygen-ui or `@types/react` skew between the two portals surfaces first in the console image build. Separately, `include: ["src"]` has no `exclude` for `*.test.*`, unlike gateways, which excludes them precisely because "The AI Workspace has no vitest installed, so the tests cannot resolve it under these paths". Confirmed against `portals/cloud-plugins/apip-cloud-ui-gateways/tsconfig.json` and `tsconfig.console.json`. The exact `tsc` diagnostics are unverified — no `node_modules` were installed in my checkout.

**How to verify**
`ls portals/cloud-plugins/*/tsconfig.console.json` — only `apip-cloud-ui-gateways` has one, and it is the only other plugin imported by both `hosts/ai-workspace.tsx` and `hosts/api-control-plane.tsx`. Then run `npx tsc --noEmit -p portals/cloud-plugins/apip-cloud-ui-insights/tsconfig.json` after a Rush install and see whether the test files resolve `vitest`/`@testing-library/*` under the ai-workspace-pointed `react` path.

**Suggested fix**
Copy `apip-cloud-ui-gateways/tsconfig.console.json` into this package (adding `vitest` and `@testing-library/*` path entries, since unlike gateways this package's tests need them), add `"exclude": ["src/**/*.test.ts", "src/**/*.test.tsx"]` here, and run both configs from `_phase:build` the way gateways does.

**Prompt for Claude**
```text
In portals/cloud-plugins/apip-cloud-ui-insights.

1. FIRST verify: read portals/cloud-plugins/apip-cloud-ui-gateways/tsconfig.json and
   tsconfig.console.json and confirm (a) the console config exists because the plugin is
   registered in BOTH hosts, and (b) the base config excludes src/**/*.test.ts because AI
   Workspace has no vitest. Then confirm apip-cloud-ui-insights is imported by BOTH
   portals/cloud-plugins/apip-cloud-ui/src/hosts/api-control-plane.tsx and
   .../hosts/ai-workspace.tsx. If insights is registered in only one host, STOP — the
   finding is wrong.
2. If it holds, add portals/cloud-plugins/apip-cloud-ui-insights/tsconfig.console.json
   mirroring the gateways one (paths under ../../api-control-plane/node_modules), including
   entries for vitest and @testing-library types that this package's tests need, and add
   "exclude" for test files to the base tsconfig.json.
3. Update the package's "build"/"_phase:build" scripts to run both configs, matching how
   apip-cloud-ui-gateways does it.
4. Do not change src/ behaviour. Confirm with `npx tsc --noEmit -p tsconfig.json` and
   `-p tsconfig.console.json` after a Rush install; report any errors rather than silencing
   them with skipLibCheck or any-casts.
```

portals/cloud-plugins/apip-cloud-ui-insights/src/api/analyticsApi.ts
```diff
@@ -0,0 +1,167 @@
...
+export async function resolveProjectScope(
+  orgHandle: string,
+  projectHandle: string
+): Promise<{ projectId: string; projectName: string }> {
+  const trimmedHandle = projectHandle.trim();
+  if (!trimmedHandle) {
+    throw new Error('Project scope is required for project insights.');
+  }
+
+  const headers = { 'X-Org-Id': orgHandle };
```
AnuGayan, line 127: **Severity:** 🟠 major
**Subject:** The plugin bypasses the host Port's `apiFetch` and sends `X-Org-Id`, which the Port contract says must not be sent

**Description**
`portals/api-control-plane/src/hostPort.tsx:64-70` documents the extension transport contract explicitly: *"The org is resolved by platform-api from that token, so no `X-Org-Id` is sent here."* Every other cloud plugin (`apip-cloud-ui-gateways`, `apip-cloud-ui-pipelines`) takes `apiFetch` off the Port and uses it. This plugin instead defines an `InsightsHostPort` (`src/hostPort.ts`) *without* `apiFetch` and rolls its own `fetch` with a hand-built base URL and this `X-Org-Id` header. Two consequences: the plugin re-derives `platformApiBaseUrl` with a *different* precedence than the host (`src/config/runtimeConfig.ts` puts `window` ahead of `VITE_*`; `api-control-plane/src/config/runtime.ts:166` puts `VITE_*` first), so a dev override targeting a local BML diverges; and the org the viewer token is minted for comes from the BFF session while the project lookup comes from a header — two sources of org truth in one feature. Confirmed by reading `hostPort.tsx`, `apip-cloud-ui-gateways/src/hostPort.ts`, and this file.

**How to verify**
`grep -rn "apiFetch" portals/cloud-plugins/*/src/hostPort.ts` — insights is the only plugin whose Port omits it. Then read the comment at `portals/api-control-plane/src/hostPort.tsx:68`.

**Suggested fix**
Add `apiFetch: ApiFetch` to `InsightsHostPort` (mirroring `apip-cloud-ui-gateways/src/hostPort.ts`) and route `resolveProjectScope` through it, dropping this header and the local base-URL derivation. `fetchViewerToken` targets `/proxy/cloud/...`, which is not under the platform-api base, so it legitimately stays a direct fetch — but it should reuse the base prefix the host publishes rather than recomputing it.

**Prompt for Claude**
```text
In portals/cloud-plugins/apip-cloud-ui-insights.

1. FIRST verify: read portals/api-control-plane/src/hostPort.tsx around lines 60-72 and confirm
   the comment says the org is resolved from the session token and no X-Org-Id is sent. Read
   portals/cloud-plugins/apip-cloud-ui-gateways/src/hostPort.ts and confirm its Port exposes
   `apiFetch`. If the ACP Port does NOT expose apiFetch, or if platform-api actually requires
   X-Org-Id on GET /projects when called through the BFF proxy, STOP and tell me.
2. If it holds: add `apiFetch: ApiFetch` to src/hostPort.ts (copy the ApiFetch type and doc
   comment from the gateways plugin), thread `port` into `resolveProjectScope`, and replace the
   raw fetch + X-Org-Id header in src/api/analyticsApi.ts with
   `port.apiFetch('GET', '/projects/...')`.
3. Leave fetchViewerToken's /proxy/cloud/analytics/id-token call as a direct fetch — it is not
   under the platform-api base — but keep it reading the base prefix from one place.
4. Do not change the user-facing error copy in userFacingRequestError, which has tests.
5. Update src/InsightsFeature.test.tsx and the analyticsApi describe block in
   src/utils/moesifEmbed.test.ts to stub apiFetch instead of global fetch, and run `npm test`.
```

portals/cloud-plugins/apip-cloud-ui-insights/src/api/analyticsApi.ts
```diff
@@ -0,0 +1,167 @@
...
+  value && typeof value === 'object' ? (value as Record<string, unknown>) : {};
+
+const pickProjectHandle = (project: ProjectRecord) =>
+  project.id?.trim() ||
+  project.handler?.trim() ||
+  project.handle?.trim() ||
+  '';
+
+const pickProjectId = (project: ProjectRecord) =>
+  project.uuid?.trim() || pickProjectHandle(project);
```
AnuGayan, line 112: **Severity:** 🟡 minor
**Subject:** `project_id` sent to Moesif is the project handle; the `uuid`/`handler`/`handle`/`name` branches are dead against the API contract

**Description**
`ProjectRecord` (L93-101) declares optional `uuid`, `handler`, `handle`, `name` and this line prefers `uuid` for the Moesif `project_id`. The platform-api `Project` schema has none of those: only `id` (the URL-friendly handle), `displayName`, `organizationId`, and audit fields. So `pickProjectId` always falls through to `project.id`, and `buildBasicProjectIframeSrc` always sends `project_id=<handle>`. The "prefers uuid when present" test at `src/utils/moesifEmbed.test.ts:210` passes only because it feeds a hand-written fixture the API never produces. Whether Moesif filters correctly on a handle is the load-bearing question — if it does not, the embed shows unfiltered or empty data with no error. Confirmed against `platform-api/resources/openapi.yaml` (`Project:` schema) and `portals/api-control-plane/src/api/generated/platform.d.ts`. **Unverified:** what Moesif actually accepts as `project_id`.

**How to verify**
`grep -n -A30 '^    Project:' platform-api/resources/openapi.yaml` — no `uuid`, `handler`, `handle`, or `name` property. Then open a project-scope Insights page and read the iframe `src` in devtools: it carries the URL handle, not a UUID.

**Suggested fix**
Confirm with the Moesif integration what `project_id` is keyed on. If it is the handle, delete the speculative fields and collapse the pickers to `project.id` / `project.displayName`, typed against `Schema<'Project'>`. If it is a UUID, this needs a platform-api field to carry it — don't ship the handle. I have not validated either shape against Moesif, so no diff here.

**Prompt for Claude**
```text
In portals/cloud-plugins/apip-cloud-ui-insights/src/api/analyticsApi.ts lines 93-115.

1. FIRST verify the contract: read the `Project:` schema in platform-api/resources/openapi.yaml
   and the generated `Project` type in portals/api-control-plane/src/api/generated/platform.d.ts.
   Confirm there is no `uuid`, `handler`, `handle`, or `name` property. If any exists, STOP —
   the finding is wrong.
2. Do NOT change the value sent to Moesif until someone confirms what Moesif's `project_id`
   is keyed on — ask me before touching buildBasicProjectIframeSrc.
3. If the contract confirms, remove the dead `uuid`/`handler`/`handle`/`name` branches from
   ProjectRecord, pickProjectHandle, pickProjectId and pickProjectName, and type the response
   against the generated Project type instead of a hand-written record.
4. Update the now-stale tests in src/utils/moesifEmbed.test.ts ("prefers uuid when present",
   "matches project id when handler differs") so they exercise the real response shape.
5. Confirm with `npm test`.
```

portals/cloud-plugins/apip-cloud-ui-insights/src/InsightsFeature.tsx
```diff
@@ -0,0 +1,206 @@
...
+        '|'
+      ),
+    [embedProfile, orgHandle, projectHandle, requestedScopeLevel]
+  );
+
+  const [embedScopeLevel, setEmbedScopeLevel] =
+    useState<InsightsScopeLevel>(requestedScopeLevel);
+  const [projectId, setProjectId] = useState<string | null>(null);
+  const [projectName, setProjectName] = useState<string | null>(null);
+  const [scopeError, setScopeError] = useState<string | null>(null);
```
AnuGayan, line 95: **Severity:** 🟡 minor
**Subject:** Dead code across the new package — unreachable error branch, test-only exports, unused Port members

**Description**
Five instances, all against the "no unused identifiers / no dead code" standard. `scopeError` here is assigned `null` at three sites and never anything else, making the `if (scopeError)` branch at L179-183 unreachable. `src/utils/moesifEmbed.ts:58` exports `resolveTrustedMoesifAppUrl`, referenced only by its own test. `src/index.ts:27` exports `isInsightsMoesifConfigured`, which no host imports — `portals/ai-workspace/src/App.tsx:112` re-implements it instead. `src/utils/routeParams.ts:28` returns `apiHandler`, unused. `src/hostPort.ts` marks `navigate` and `notify` required, and neither is called anywhere in the package. Confirmed by grepping each identifier across `portals/`.

**How to verify**
```
grep -rn "resolveTrustedMoesifAppUrl\|isInsightsMoesifConfigured\|apiHandler\|port\.navigate\|port\.notify" portals/ | grep -v node_modules
```
Every production hit is the declaration itself.

**Suggested fix**
Delete `scopeError`/`setScopeError` and the branch it guards; delete `resolveTrustedMoesifAppUrl` and its four tests; either wire `isInsightsMoesifConfigured` into `App.tsx` (see the separate comment there) or drop the export; drop `apiHandler`; make `navigate`/`notify` optional or remove them.

**Prompt for Claude**
```text
In portals/cloud-plugins/apip-cloud-ui-insights.

1. FIRST verify each item by grepping across portals/ (excluding node_modules):
   - src/InsightsFeature.tsx: is setScopeError ever called with a non-null value?
   - src/utils/moesifEmbed.ts: is resolveTrustedMoesifAppUrl referenced outside its own test?
   - src/index.ts: does any host import isInsightsMoesifConfigured?
   - src/utils/routeParams.ts: is apiHandler read anywhere?
   - src/hostPort.ts: are navigate/notify called anywhere in this package?
   Skip any item whose grep finds a real consumer.
2. For the confirmed ones: remove scopeError and its unreachable render branch; remove
   resolveTrustedMoesifAppUrl and its tests; remove apiHandler; make navigate/notify optional
   on InsightsHostPort. For isInsightsMoesifConfigured, do NOT delete it yet — it is the
   intended replacement for the duplicated helper in portals/ai-workspace/src/App.tsx; flag it
   to me instead.
3. Do not remove pickAllowlistedMoesifAppUrl, which runtimeConfig.ts uses.
4. Confirm with `npx tsc --noEmit` and `npm test` in the package.
```

portals/ai-workspace/src/App.tsx
```diff
@@ -97,6 +98,43 @@ import {
...
+  'https://www.moesif.com',
+  'https://web-dev.moesif.com',
+]);
+
+/**
+ * Whether cloud Insights can embed Moesif. Kept in App (not imported from the
+ * plugin) so the OSS portal builds without `@wso2-enterprise/apip-cloud-ui-insights`.
+ * When false, InsightsRoute keeps the built-in page even if a cloud override is registered.
+ */
+function isCloudInsightsMoesifConfigured(): boolean {
```
AnuGayan, line 112: **Severity:** 🟡 minor
**Subject:** Moesif config resolution is duplicated here with a different key set and different precedence from the plugin

**Description**
Hand-copying `ALLOWED_MOESIF_ORIGINS` and the config reader so the OSS portal builds without the plugin is a reasonable goal, but the two readers disagree. The plugin's `configuredMoesifAppUrl()` also accepts `MOESIF_BASIC_INSIGHTS_URL` / `moesifBasicInsightsUrl` / `VITE_MOESIF_BASIC_INSIGHTS_URL`; this one does not. And the plugin merges `{...__RUNTIME_CONFIG__, ...config}` so `window.config` wins, while this chains `__RUNTIME_CONFIG__?.X || config?.X` so `__RUNTIME_CONFIG__` wins. A deployment that sets only a `BASIC_INSIGHTS` key, or sets both objects with different values, gets `InsightsRoute` and the plugin disagreeing about whether Moesif is configured — the built-in page renders while the plugin would have embedded fine. Confirmed by reading this file and `portals/cloud-plugins/apip-cloud-ui-insights/src/config/runtimeConfig.ts`. The plugin already exports `isInsightsMoesifConfigured` for exactly this, and nothing uses it.

**How to verify**
Set `window.config = { moesifBasicInsightsUrl: 'https://www.moesif.com' }` with `__RUNTIME_CONFIG__` empty and load `/ai-workspace/insights`: the built-in page renders even though the plugin's own `insightsRuntimeConfig.moesifAppUrl` resolves.

**Suggested fix**
Keep one reader. Either import `isInsightsMoesifConfigured` in the cloud host file (`apip-cloud-ui/src/hosts/ai-workspace.tsx`) and let the registered override decide, so `App.tsx` needs no plugin knowledge at all; or, if the check must stay here, make the key list and the `config`/`__RUNTIME_CONFIG__` precedence identical to the plugin's and cross-reference both in comments.

**Prompt for Claude**
```text
In portals/ai-workspace/src/App.tsx lines 99-136 and
portals/cloud-plugins/apip-cloud-ui-insights/src/config/runtimeConfig.ts.

1. FIRST verify the divergence: list the keys each reader accepts, and compare the
   window.config vs window.__RUNTIME_CONFIG__ precedence in each. If the key sets and
   precedence already match, STOP — the finding is wrong.
2. Preferred fix: move the "is Moesif configured" decision out of App.tsx into the cloud host
   registration (portals/cloud-plugins/apip-cloud-ui/src/hosts/ai-workspace.tsx), using the
   plugin's exported isInsightsMoesifConfigured, so InsightsRoute only checks whether an
   override is registered. Verify the OSS ai-workspace build still compiles with no import of
   @wso2-enterprise/apip-cloud-ui-insights in portals/ai-workspace/src — that constraint is
   the whole reason the helper was inlined.
3. If step 2 is not possible, instead align App.tsx's key list and precedence exactly with the
   plugin's and cross-reference both in comments.
4. Do not change the fallback behaviour: when Moesif is unconfigured, InsightsRoute must keep
   rendering the built-in <Insights /> page.
5. Confirm with `npm run typecheck` and `npm test` in portals/ai-workspace.
```

portals/api-control-plane/src/pages/appShell/appShellPages/insights/InsightsPage.tsx
```diff
@@ -29,6 +32,10 @@ const messages = defineMessages({
     description:
       'Button that opens the Moesif analytics console in a new tab. Moesif is a product name — leave it untranslated.',
   },
+  cloudFeature: {
+    id: 'appShell.insightsPage.feature',
```
AnuGayan, line 36: **Severity:** 🟡 minor
**Subject:** New i18n message `appShell.insightsPage.feature` is not in `src/i18n/messages/en.json`

**Description**
This id is absent from the extracted catalog, so `npm run i18n:check` (`i18n:extract` then `git diff --exit-code src/i18n/messages`) fails. Runtime is unaffected — the babel plugin runs with `ast: true` and no `removeDefaultMessage`, so `defaultMessage` still renders — but the catalog drifts and translators never see the string. The id prefix itself is correct: `appShell.compliancePage.feature` is the established shape for `ComingSoon` features. Confirmed: `grep '"appShell.insightsPage' portals/api-control-plane/src/i18n/messages/en.json` returns nothing, while `appShell.compliancePage.feature` is present at line 2184.

**How to verify**
`cd portals/api-control-plane && npm run i18n:check` — exits non-zero with `en.json` modified.

**Suggested fix**
Run `npm run i18n` in `portals/api-control-plane` and commit the regenerated `src/i18n/messages/en.json`. Also consider adding a `description`, as the sibling messages in this file do.

**Prompt for Claude**
```text
In portals/api-control-plane.

1. FIRST verify: grep '"appShell.insightsPage.feature"' src/i18n/messages/en.json. If it is
   present, STOP — the finding is wrong.
2. If absent, add a `description` field to the `cloudFeature` message in
   src/pages/appShell/appShellPages/insights/InsightsPage.tsx matching the style of its
   neighbours, then run `npm run i18n` and commit the regenerated src/i18n/messages/en.json.
3. Do not hand-edit en.json and do not touch src/i18n/compiled.
4. Confirm with `npm run i18n:check` exiting 0.
```

portals/api-control-plane/bff/internal/server/middleware.go
```diff
@@ -47,7 +51,7 @@ func (s *Server) securityHeaders(next http.Handler) http.Handler {
 		h.Set("X-Frame-Options", "DENY")
 		h.Set("Content-Security-Policy", "frame-ancestors 'self'")
 		h.Set("X-Content-Type-Options", "nosniff")
-		h.Set("Referrer-Policy", "no-referrer")
+		h.Set("Referrer-Policy", "strict-origin-when-cross-origin")
```
AnuGayan, line 54: **Severity:** 🟡 minor
**Subject:** `Referrer-Policy` relaxed globally on both BFFs, though the iframe's own `referrerpolicy` attribute already governs that request

**Description**
This line and `portals/ai-workspace/bff/internal/server/middleware.go:49` change the document-wide header from `no-referrer` to `strict-origin-when-cross-origin` for *every* response, justified by Moesif embed allowlisting. But `InsightsEmbed.tsx` already sets `referrerPolicy="strict-origin-when-cross-origin"` on the `<iframe>`, and per the Referrer Policy spec an element's `referrerpolicy` attribute overrides the document policy for that fetch — so the header change is very likely not what makes the embed work. Meanwhile it applies to OSS standalone deployments that never load Moesif, weakening a documented security header for them. Exposure is modest (only the origin leaks cross-origin; console URLs carrying org/project/API handles are not sent), which is why this is minor rather than major. Confirmed by reading both middleware files and the iframe attribute; **unverified** against a live Moesif embed — I could not test whether Moesif's allowlist check sees a `Referer` with only the attribute set.

**How to verify**
Load the Insights page with the BFF still emitting `Referrer-Policy: no-referrer` and inspect the `Referer` request header on the iframe's document request in devtools — the attribute should still produce the origin.

**Suggested fix**
Confirm in a browser first. If the attribute suffices, revert both header changes. If a document-level policy really is required, scope it to the SPA document response rather than every BFF response, and gate it on the deployment actually configuring Moesif. Either way this deserves a test — `grep -rn "Referrer-Policy" portals/*/bff` currently returns only the two `middleware.go` lines, so no test covers `securityHeaders` at all.

**Prompt for Claude**
```text
In portals/api-control-plane/bff/internal/server/middleware.go:54 and
portals/ai-workspace/bff/internal/server/middleware.go:49.

1. DO NOT change anything yet. First establish the fact: with the header reverted to
   "no-referrer", load the Insights page in a browser and inspect the Referer header sent on
   the Moesif iframe's document request. The iframe already carries
   referrerPolicy="strict-origin-when-cross-origin" (see
   portals/cloud-plugins/apip-cloud-ui-insights/src/InsightsEmbed.tsx). If the origin is still
   sent, the global header change is unnecessary — report that and revert both files. If it is
   NOT sent, the change is justified — report that instead and leave the code alone.
2. Either way, add a Go test in each BFF asserting the exact Referrer-Policy value the team
   settles on.
3. Do not touch X-Frame-Options, the frame-ancestors CSP, or X-Content-Type-Options.
```

portals/cloud-plugins/apip-cloud-ui-insights/overlays/api-control-plane/InsightsPage.tsx
```diff
@@ -0,0 +1,46 @@
...
+ * when assembling a cloud deployment image.
+ */
+
+import { FormattedMessage } from 'react-intl';
+
+import { ComingSoon } from '../../../../components/ComingSoon';
+import { routes } from '../../../../routes/paths';
+import { ScopeGate } from '../../../../scope/ScopeGate';
+
+export function InsightsPage() {
```
AnuGayan, line 29: **Severity:** 🟡 minor
**Subject:** This overlay now duplicates the inline `cloudProxyEnabled` branch it was meant to replace

**Description**
This file exists to be copied over the built-in page during cloud image assembly, and renders exactly the `ScopeGate` + `ComingSoon` tree that this same PR added *inline* to the built-in page behind `if (runtimeConfig.cloudProxyEnabled)`. Two mechanisms now produce the same behaviour, and this file is outside every `tsconfig` `include` (they cover only `src`), so it is never typechecked and will drift the moment `ScopeGate`'s props or `routes.apiInsightsApi` change. Nothing in this repo copies it. Confirmed by reading both files and `portals/cloud-plugins/apip-cloud-ui-insights/tsconfig.json`.

**How to verify**
Diff the two render bodies — identical apart from import style. Then `grep -rn "overlays/api-control-plane" --include=Dockerfile --include=*.mjs --include=*.yml .` finds no consumer in this repo.

**Suggested fix**
Drop the `overlays/` directory and the README row describing it, since the runtime flag already covers the cloud case. Keep it only if an out-of-repo Dockerfile genuinely copies it — in which case the inline `cloudProxyEnabled` branch is the redundant half and should go instead.

**Prompt for Claude**
```text
In portals/cloud-plugins/apip-cloud-ui-insights/overlays/api-control-plane/InsightsPage.tsx and
portals/api-control-plane/src/pages/appShell/appShellPages/insights/InsightsPage.tsx.

1. FIRST verify: confirm the two render exactly the same ScopeGate + ComingSoon tree, and
   search the repo for anything that copies the overlay file (Dockerfiles, *.mjs build scripts,
   .github/workflows). If a consumer exists, STOP and tell me which one — then the inline
   runtimeConfig.cloudProxyEnabled branch is the redundant half, not the overlay.
2. If no consumer exists, delete the overlays/ directory and the overlay rows in the package
   README that describe it.
3. Do not change the inline `if (runtimeConfig.cloudProxyEnabled)` branch or the OSS
   ExternalToolPanel fallback below it.
4. Confirm with `npm run typecheck` in portals/api-control-plane and `npm test`.
```

AnuGayan (commented):
## Review — cloud Moesif Insights plugin

The plumbing here is careful: the postMessage `targetOrigin` is pinned to an allowlisted origin rather than `'*'`, the viewer token never leaves the BFF boundary unchecked, `buildRuntimeConfig` keeps the SPA from ever learning the real cloud upstream URL, and `TestReverseProxy_CloudPrefixJoinsTargetPath` covers the one thing most likely to break silently in the new proxy hop. The "shows loading instead of stale project metadata when switching projects" test catches a genuinely subtle race, and `resolvedScopeKey === scopeKey` is a clean way to express it.

**Verdict: 🔧 Request changes** — one confirmed silent failure in the token-refresh path (the embed dies ~1h after the page is opened), plus a missing `tsconfig.console.json` that leaves a dual-host plugin unverified against one of its two hosts.

### Findings (details inline)

| | Finding | Where |
|---|---|---|
| 🟠 | Refreshed viewer token is never posted to the iframe once the handshake completes | `InsightsEmbed.tsx:161-166` |
| 🟠 | Dual-host plugin ships no `tsconfig.console.json`; nothing typechecks it under ACP's deps | `apip-cloud-ui-insights/tsconfig.json` |
| 🟠 | Plugin bypasses the host Port's `apiFetch` and sends `X-Org-Id`, which the Port contract forbids | `analyticsApi.ts:127` |
| 🟡 | `project_id` sent to Moesif is the handle; `uuid`/`handler`/`handle`/`name` are dead against the OpenAPI contract | `analyticsApi.ts:93-115` |
| 🟡 | Dead code: unreachable `scopeError` branch, test-only exports, unused Port members | `InsightsFeature.tsx:95` + 4 others |
| 🟡 | Moesif config resolution duplicated with a different key set and precedence | `ai-workspace/src/App.tsx:99-136` |
| 🟡 | `appShell.insightsPage.feature` missing from `en.json` — `npm run i18n:check` fails | `InsightsPage.tsx:36` |
| 🟡 | `Referrer-Policy` relaxed globally, though the iframe attribute already governs that request | `middleware.go:54` (both BFFs) |
| 🟡 | Cloud overlay now duplicates the inline `cloudProxyEnabled` branch | `overlays/api-control-plane/InsightsPage.tsx` |

### Nits

- **`ai-workspace/bff/internal/server/routes.go:72`** — "must be registered before the catch-all `/proxy/`" is not why it works: Go's `ServeMux` picks the most specific pattern regardless of registration order. Code is correct; the comment will mislead.
- **`ai-workspace/bff/internal/server/server.go:101`** — `return nil, err` from `url.Parse` is unwrapped, unlike the primary target three lines above (`fmt.Errorf("parse control_plane.url: %w", err)`). A bad `cloud_url` gives an operator no key name.
- **`ai-workspace/bff/internal/server/handlers.go:196-200`** — `serveProxy`'s `rp == nil` branch is unreachable: `routes.go` only registers `handleCloudProxy` when `cloudProxy != nil`, and `s.proxy` is always set.
- **ai-workspace misconfiguration path** — if `moesif_web_url` is set but `[control_plane] cloud_url` is not, `/ai-workspace/proxy/cloud/analytics/id-token` falls through to the primary proxy, reaches platform-api as `/cloud/analytics/id-token`, and 404s — which `analyticsApi.ts:38` renders as *"Insights are not available for this organization. Contact your administrator…"*. Consider validating the pair at startup, or emitting a `cloudProxyEnabled` runtime flag for ai-workspace the way ACP does.
- **`src/utils/moesifEmbed.test.ts:107-262`** — the whole `describe('analyticsApi')` block tests `src/api/analyticsApi.ts`, not `moesifEmbed`. It belongs in `src/api/analyticsApi.test.ts`.
- **`src/utils/moesifEmbed.test.ts:95-101`** — `it('rejects non-https moesif hosts')` passes whether or not the https check exists, since `http://www.moesif.com` is not in the allowlist either. Assert `pickAllowlistedMoesifAppUrl('http://www.moesif.com')` is `undefined` directly.
- **`analyticsApi.ts:151`** — the `GET /projects` fallback iterates an unpaginated list. Fine at current scale, but it is an unbounded result set on the hot path for every project-scope page load; the primary `GET /projects/{handle}` lookup already covers the normal case.
- **`analyticsApi.ts:56,74`** — `credentials: 'include'` is a no-op for these same-origin calls.
- **Hardcoded origin allowlist** — `ALLOWED_MOESIF_ORIGINS` admits only `www.moesif.com` and `web-dev.moesif.com` in three places (plugin util, plugin config, `App.tsx`). Any other Moesif deployment silently disables Insights with no diagnostic; a one-line `console.warn` on rejection would save a support round-trip.
- **New tests won't run in CI** — `package.json` adds `"test": "vitest run"` but no `_phase:test`, and no workflow under `.github/workflows` covers `portals/cloud-plugins`. The 8 new test files are currently documentation.

---

AnuGayan (approved):

PR #3353, opened 2026-09-02
Adding Project handle as a metadata for analytics events

Milanka00:
## Purpose

This PR fixes an inconsistency in how the project ID is propagated from platform-api to the gateway and Moesif.

Previously, REST API deployments used the internal `project_uuid` when populating the `gateway.api-platform.wso2.com/project-id` annotation, while Project Insights used the project handle (e.g. new-project) when filtering Moesif traffic. This caused Project Insights to show 0 traffic for APIs with existing traffic.

**Changes**:

- Adds a second deployment annotation `gateway.api-platform.wso2.com/project-handle` and teaches the gateway analytics path to prefer it for Moesif, while keeping `gateway.api-platform.wso2.com/project-id` as the internal project UUID.
- Leaves non-analytics UUID references unchanged (DB FK, `project-id` annotation, `Metadata.ProjectID`, import contract).

**_After this change,_**

1. platform-api BuildAPIDeploymentYAML: stamps both annotations on deployment YAML

2. gateway-controller: reads both → Metadata.ProjectID (UUID) + Metadata.ProjectHandle (handle)

3. analytics emit (policy xDS / Envoy route metadata): project_id = handle (prefer project-handle)

4. policy-engine → Moesif: metadata.projectId = handle  ← matches Insights


gateway/gateway-controller/pkg/models/runtime_deploy_config.go
```diff
@@ -47,13 +48,23 @@ type RuntimeDeployConfig struct {
...
+	Version       string
+	DisplayName   string
+	ProjectID     string // from gateway.api-platform.wso2.com/project-id (UUID for CP deploys)
+	ProjectHandle string // from gateway.api-platform.wso2.com/project-handle (analytics-facing)
+	LLM           *LLMMetadata // nil for non-LLM kinds
+}
+
+// AnalyticsProjectRef returns the project identity to publish for analytics
+// (Moesif metadata.projectId). Prefer the user-facing handle when present.
+func (m Metadata) AnalyticsProjectRef() string {
```
O-sura, line 63: The requirement is to send both to Moesif. AnalyticsProjectRef() currently replaces the UUID with the handle, which changes the meaning of the existing projectId attribute and could break existing dashboards/alerts and other consumers of that value.

Can we keep the existing projectId as the UUID and add the handle as a separate attribute instead? That would satisfy the requirement without changing existing semantics.

common/constants/constants.go
```diff
@@ -2,7 +2,12 @@ package constants
 
 const (
 	// AnnotationProjectID is the domain-prefixed annotation key for project identity on API resources.
+	// Control-plane deployments stamp the internal project UUID here. Gateway import still
+	// accepts a project handle in this key for DP-originated artifacts.
 	AnnotationProjectID = "gateway.api-platform.wso2.com/project-id"
+	// AnnotationProjectHandle is the user-facing project handle used for analytics
+	// (e.g. Moesif metadata.projectId). Prefer this over AnnotationProjectID when both are set.
+	AnnotationProjectHandle = "gateway.api-platform.wso2.com/project-handle"
```
O-sura, line 10: ResolveImportProject still reads the dual-typed project-id annotation and passes it to a handle lookup, so control-plane artifacts with UUIDs would fail if they reach this path.

Since we now have a dedicated handle annotation, can the importer prefer the new key and fall back to project-id for older artifacts? This would also let us eventually make project-id consistently UUID-based and remove the current ambiguity.

gateway/gateway-controller/pkg/xds/translator.go
```diff
@@ -1508,6 +1508,24 @@ func extractProjectIDFromConfig(cfg *models.StoredConfig) string {
 	return ""
 }
 
+// extractProjectHandleFromConfig reads the analytics-facing project handle annotation.
```
O-sura, line 1511: This change is a no-op here: the value is only passed to the REST route builder, which doesn’t use the project-id parameter. The actual Envoy metadata is stamped by the per-topic route builder, which still uses the original UUID-only extraction. WebSub would keep publishing the UUID while REST publishes the handle, making the same Moesif attribute mean different things by API type.

follow-up: MCP/LLM APIs currently get no project attribute, and MCP would drop annotations during conversion anyway. We should confirm whether those API types are in scope.

PR #3360, opened 2026-09-03
Add A2A support to AI Gateway

### LINKED ISSUE #2844: [Task]: Implement A2A Support in API Platform

Gateway

Aspect/AI (AI/LLM integration, MCP, AI readiness)

### Description

This issue is created to track the progress of A2A implementation in API Platform.

The discussion can be found in https://github.com/wso2/api-platform/discussions/2743

RakhithaRR:
## Purpose

The platform had no first-class way to front an A2A (Agent2Agent) agent. The only option was to hand-roll a REST API proxy per agent, which cannot express A2A's actual shape: two protocol bindings address the *same* eleven operations, and one of them (JSON-RPC) is a single endpoint that names its operation in the request body. Because a REST API's route and its policy chain are the same thing, there was no way to attach per-operation policy, authentication, rate limiting or analytics to an agent, and no way to serve or protect an Agent Card.

This PR introduces `Agent` as a first-class artifact kind so an A2A agent is managed, secured, observed and deployed the same way REST APIs, MCP servers and LLM providers already are.

Resolves https://github.com/wso2/api-platform/issues/2844

## Goals
- A first-class `Agent` artifact kind implementing A2A protocol version 1.0.
- Both protocol bindings — `JSONRPC` and `HTTP+JSON` — over the same eleven operations, addressable independently.
- Agent-wide and per-operation policy attachment, with correct chain ordering.
- Agent Card handling: public and protected cards, each in `managed` (served by the gateway) or `passthrough` (proxied from the upstream) mode.
- Per-agent API keys, plus the existing authentication policies (JWT/OAuth2) usable on agent operations.
- Analytics and tracing that understand A2A, with operation and transport as first-class dimensions.
- Kubernetes-native declaration via an `Agent` CRD reconciled by the gateway operator.
- Multi-replica correctness through the existing event-sync mechanism.

## Approach

**Management REST API** — `POST/GET /agents`, `GET/PUT/DELETE /agents/{id}`, plus `/agents/{id}/api-keys` (create, list, delete, regenerate). Models are generated from `management-openapi.yaml` via `make generate-server-code`; REST API docs are regenerated alongside.

**Storage** — one new `agents` table per dialect (SQLite, PostgreSQL, SQL Server). Purely additive: guarded `CREATE TABLE IF NOT EXISTS` / `IF OBJECT_ID(...) IS NULL`, no `ALTER` against any shipped table, so it applies to fresh and already-provisioned databases alike.

**Transformation** — `AgentTransformer` builds the `RuntimeDeployConfig` directly rather than desugaring an Agent into a synthetic `RestAPI` the way MCP does. A REST-shaped intermediate cannot express "eleven chains, one route", so building one and patching it would mean maintaining the patch rather than reusing the model. Everything that decides a value another component also computes — upstream cluster construction, resilience parsing, policy version resolution, system-policy injection, route naming, card path arithmetic — is reused rather than reimplemented.

**Policy engine** — a new A2A chain resolver selects the operation chain from the path (`HTTP+JSON`) or from the request body (`JSON-RPC`, plus the two message-sending HTTP+JSON routes). Body-resolved routes carry an explicit 1 MiB acceptance ceiling, because on those routes the body is parsed before any policy — authentication included — has run, so that ceiling is what bounds the unauthenticated parsing work an anonymous caller can request. The resolver interface gained protocol-specific attributes, forwarded to both the Go and Python policy engines.

**Agent Card** — a public card is either served locally from configured content (with ETag and conditional-GET support) or proxied byte-identically from the upstream. A protected (extended) card is served locally in `managed` mode or proxied in `passthrough` mode, and is **always** guarded: the guard instance is unconditional, so "no auth policy attached" cannot silently mean "public extended card". Omitting `agentCard.protected` reads as `passthrough`, not as unprotected — see Reviewer notes.

**Envoy/xDS** — per-operation routes for both bindings, `OPTIONS` preflight routes generated beside the real ones so CORS preflights have something to match before any policy runs, and per-operation resiliency configuration.

**Operator** — an `Agent` CRD in both `v1` and `v1alpha1` with a conversion path, a reconciler, RBAC, and a sample CR. Helm chart updated to install it.

**Analytics and tracing** — A2A invocations report operation and transport dimensions; the same operation over both bindings shares an operation dimension and differs by transport. Card fetches and CORS preflights are reported as themselves rather than shaped like invocations, and a JSON-RPC error inside a `200` is reported as a failed invocation. Policy denials are attributed to the gateway rather than to the agent.

## Automation tests

  66 new BDD scenarios across six feature files:

  | Feature | Scenarios | Covers |
  |---|---|---|
  | `agent_deploy` | 13 | CRUD, deploy/undeploy, transport declaration, all eleven operations over both bindings via the official A2A SDK |
  | `agent_card` | 23 | managed/passthrough public and protected cards, ETag and conditional GET, custom card path, guard behaviour, card validation rejections |
  | `agent_policies` | 8 | agent-wide vs per-operation attachment and ordering, CORS preflight |
  | `agent_streaming` | 5 | `SendStreamingMessage` and `SubscribeToTask` over both bindings, terminal-state close, buffered error responses |
  | `agent_analytics` | 8 | operation/transport dimensions, discovery vs invocation, JSON-RPC-error-in-200, policy-denial attribution |
  | `agent-api-keys` | 9 | key lifecycle, runtime credential replacement, undeploy/redeploy, 404s |

## Related PRs

https://github.com/wso2/api-platform/pull/3198
https://github.com/wso2/api-platform/pull/3317


RakhithaRR:
@coderabbitai approve

RakhithaRR:
@coderabbitai approve

Krishanx92 (approved):

Krishanx92 (approved):

Issue #3381, opened 2026-09-06
[Bug]: Gateway Controller does not validate policy params for LLM `operationPolicies` and the deprecated `policies` list


thivindu:
### Please select the area the issue is related to

Gateway

### Please select the aspect the issue is related to

Aspect/API (API backends, definitions, contracts, interfaces, OpenAPI)

### Description

`PolicyValidator` validates policy `params` against the policy definition's declared JSON Schema for most artifact types, but silently skips it for two of the five policy collections an `LlmProvider`/`LlmProxy` can declare.

[`validateLLMPolicyRefs`](https://github.com/wso2/api-platform/blob/main/gateway/gateway-controller/pkg/config/policy_validator.go#L120-L150) resolves the policy reference for the operation-level and deprecated lists and then **discards the resolved definition**, so the `params` are never schema-checked:

```go
for i, policy := range *operationPolicies {
    _, errs := pv.validatePolicyRef(policy.Name, policy.Version, fmt.Sprintf("spec.operationPolicies[%d]", i))
    errors = append(errors, errs...)
}
```

- [`operationPolicies`](https://github.com/wso2/api-platform/blob/main/gateway/gateway-controller/pkg/config/policy_validator.go#L133-L139) and [deprecated `policies`](https://github.com/wso2/api-platform/blob/main/gateway/gateway-controller/pkg/config/policy_validator.go#L141-L147).

Only the global (api-level) list routes through [`validatePolicy`](https://github.com/wso2/api-platform/blob/main/gateway/gateway-controller/pkg/config/policy_validator.go#L152-L174), which performs the schema check.

### Coverage today

| Artifact kind | Policy collection | Name + version | Params vs. schema |
|---|---|---|---|
| `RestApi` | `spec.policies` | ✅ | ✅ |
| `RestApi` | `spec.operations[].policies` | ✅ | ✅ |
| `Mcp` | `spec.policies` | ✅ | ✅ |
| `LlmProvider` / `LlmProxy` | `spec.globalPolicies` | ✅ | ✅ |
| `LlmProvider` / `LlmProxy` | `spec.operationPolicies[].paths[].params` | ✅ | ❌ **not validated** |
| `LlmProvider` / `LlmProxy` | `spec.policies[].paths[].params` (deprecated) | ✅ | ❌ **not validated** |

[...]

PR #3383, opened 2026-09-07
Validate policy params for LLM operationPolicies and the policies list

thivindu:
## Purpose

`PolicyValidator` validates policy `params` against the policy definition's declared JSON Schema for most artifact types, but silently skips it for two of the five policy collections an `LlmProvider`/`LlmProxy` can declare. [`validateLLMPolicyRefs`](https://github.com/wso2/api-platform/blob/main/gateway/gateway-controller/pkg/config/policy_validator.go#L120-L150) resolved the policy reference for the operation-level and deprecated lists and then discarded the resolved definition:

```go
// Operation-level policies: validate name + version existence.
_, errs := pv.validatePolicyRef(policy.Name, policy.Version, ...)   // ← the *models.PolicyDefinition is dropped
```

Only the global (api-level) list routed through `validatePolicy`, which performs the schema check.

**Coverage before this PR:**

| Artifact kind | Policy collection | Name + version | Params vs. schema |
|---|---|---|---|
| `RestApi` | `spec.policies`, `spec.operations[].policies` | ✅ | ✅ |
| `Mcp` | `spec.policies` | ✅ | ✅ |
| `LlmProvider` / `LlmProxy` | `spec.globalPolicies` | ✅ | ✅ |
| `LlmProvider` / `LlmProxy` | `spec.operationPolicies[].paths[].params` | ✅ | ❌ |
| `LlmProvider` / `LlmProxy` | `spec.policies[].paths[].params` (deprecated) | ✅ | ❌ |

A misconfigured operation-level LLM policy therefore deployed successfully instead of being rejected: missing required params, out-of-range values and unknown properties all passed. The failure surfaced later at runtime, where the policy is dropped or misbehaves with no deploy-time signal — and the *same* params on the *same* policy were correctly rejected when attached as a `globalPolicy`, which made the behavior look arbitrary.

Resolves:
- https://github.com/wso2/api-platform/issues/3381

## Goals

Schema-validate `params` for the two remaining LLM policy collections, so an invalid operation-level policy fails at deploy time with a field path naming the offending param — matching what `RestApi`, `Mcp` and LLM `globalPolicies` already do.

## Approach

`validateLLMPolicyRefs` now uses the `*models.PolicyDefinition` that `validatePolicyRef` was already returning, and validates each path attachment's params:

```go
policyDef, errs := pv.validatePolicyRef(policy.Name, policy.Version, fieldPath)
if len(errs) > 0 { errors = append(errors, errs...); continue }
for j := range policy.Paths {
    errors = append(errors, pv.validateAttachedPolicyParams(policyDef, policy.Paths[j].Params,
        fmt.Sprintf("%s.paths[%d]", fieldPath, j))...)
}
```

A new `validateAttachedPolicyParams` helper coerces then schema-checks one params map. It handles the two things that differ from the api-level path:

- **A nil params map is still validated**, not skipped — otherwise omitting `params:` entirely would bypass a schema's `required` list.
- **Coercion runs first**, since template rendering always yields strings (`{{ env "LIMIT" }}` → `"100"` for an integer param), mirroring `validatePolicy`'s handling of api-level params. Both call sites validate rendered config (`RenderSpec` at `llm_deployment.go:270`/`:457`, validation at `:312`/`:498`), so this is the correct order.

An unresolvable name/version reports once and skips param validation, rather than repeating the same error per path.

## User stories

As an API platform user deploying an `LlmProvider`/`LlmProxy`, when I attach an operation-level policy with invalid params, the deploy is rejected with an error naming the param — instead of succeeding and silently misbehaving at runtime.

## Automation tests

- **Unit tests** — 8 new tests in `policy_validator_llm_test.go` (+188 lines), covering: valid params; missing required / out-of-range / unknown-property; absent `params` map; string→int coercion; a definition with no parameter schema; unresolvable ref not re-reporting per path; the deprecated `policies` list; and the template-merge rationale. Full `gateway-controller` module passes (`go build ./...`, `go vet`, `go test ./...`).
- **Integration tests** — none added. The change is confined to a pure validation function with no I/O; the behavior is fully covered at unit level.

### Regression check against real policy definitions

Because this turns previously-accepted config into rejected config, verified it doesn't reject anything valid: loaded all 36 real policy definitions from `wso2/gateway-controllers` and validated every operation-level/deprecated `params` block in `gateway/examples/*.yaml` (`api-key-auth`, `content-length-guardrail`, `llm-header-router`, `llm-cost-based-ratelimit`, `openai-to-bedrock-transformer`, …) — all clean, no false positives.

**Behavior change to be aware of:** an `LlmProvider`/`LlmProxy` already deployed with invalid operation-level policy params will now fail validation on its next deploy/update. That is the intended fix, but it can surface as a new failure on config that previously "worked".

## Test environment

Go 1.26.2, macOS (darwin 24.6.0). Validation logic is platform- and DB-independent; no browser or database involvement.


thivindu:
@coderabbitai review

renuka-fernando (approved):

PR #3398, opened 2026-09-07
Add token exchange functionality and related configuration

Thushani-Jayasekera:
# Token exchange for the AI Workspace BFF

Trade the user's login token for one minted specifically for the Platform API, so the
credential the BFF sends upstream is audience- and scope-scoped to that one API.

**Off by default.** A deployment that omits the new
`[ai_workspace.auth.oidc.token_exchange]` table behaves exactly as it does today: the
login token is forwarded upstream unchanged. Nothing in this PR changes the behaviour
of an existing config.

---

## Why

Two problems, one cause — the token the BFF forwards was minted for *logging into the
AI Workspace*, not for *calling the Platform API*.

**Some enterprise IDPs cannot mint the platform's `ap:*` scopes.** Microsoft Entra ID
is the documented case: it has no way to register them, and requesting the full set
exceeds its authorize-URL length limit. The existing workaround is
`[auth.authorization] mode = "role"`, where the BFF and the Platform API each load the
*same* `role-to-scope-mapping.yaml` and expand roles locally. That works, but it is one
grant table hand-mirrored across two services, and a drift shows up as a UI offering
actions that then 403.

**No least privilege on the upstream hop.** The forwarded token is the full-privilege
session token, carrying whatever audience the login client happens to have.

An exchange fixes both. The corporate IDP proves *who the user is*; an STS that *can*
mint `ap:*` scopes (WSO2 IS, Asgardeo) issues the token the Platform API actually
authorizes against. Both sides then read scopes off one token, and
`[auth.authorization] mode` can go back to `"scope"` on both — retiring the mirror.

---

## What the flow looks like

```
Browser                BFF                      IDP / STS              Platform API
   │                    │                          │                        │
   │─ login ───────────▶│─ authorization_code ────▶│                        │
   │                    │◀──── login token ────────│                        │
   │                    │                          │                        │
   │                    │─ TOKEN EXCHANGE ────────▶│   ← the new step       │
   │                    │   subject_token=login    │                        │
   │                    │   audience=platform-api  │                        │
   │                    │◀── exchanged token ──────│                        │
   │◀── session cookie ─│   (aud=platform-api,     │                        │
   │                    │    scope=ap:*)           │                        │
   │─ GET /proxy/... ──▶│─ Bearer <exchanged> ─────────────────────────────▶│
```

- The **login token stays the session anchor** — it is what the cookie carries and what
  the session store is keyed by. Only the upstream `Authorization` header changes.
- The **exchanged token is never sent to the browser.**
- The exchange runs **eagerly at login**, so a misconfiguration surfaces as a failed
  login an operator sees rather than a mystery 502 on the SPA's first call, and is
  **cached on the session** and renewed on the proxy path afterwards.

---

## Two grant types, because the IDPs disagree

This shaped the design and is the least obvious part of the PR.

**Entra ID does not implement RFC 8693.** It rejects
`grant_type=urn:ietf:params:oauth:grant-type:token-exchange` outright with
`AADSTS70003`. Its on-behalf-of flow is a *different specification* (RFC 7523), not a
dialect of it:

| | RFC 8693 (`token_exchange`) | Entra OBO (`jwt_bearer`) |
|---|---|---|
| `grant_type` | `…:token-exchange` | `…:jwt-bearer` |
| Subject travels as | `subject_token` + `subject_token_type` | `assertion` + `requested_token_use=on_behalf_of` |
| Target named by | `audience` / `resource` | `scope` (`api://<app-id>/.default`) |
| `issued_token_type` in response | REQUIRED | absent |

Since Entra is *the* IDP the platform documents as unable to mint `ap:*` scopes,
supporting only RFC 8693 would have missed the motivating deployment. So `grant_type`
is a config key with both protocols behind one interface, validated against a closed
set at startup.

The key selects the protocol; the request carries the registered URI either way. Because
that URI is the form an IDP's own documentation shows, it is accepted as an equivalent
spelling of the short name. The set stays closed regardless — a value with no branch
behind it has no implementation, so it fails startup listing every accepted spelling
rather than being forwarded to the IDP as-is.

Confirmed RFC 8693 support (official docs): WSO2 IS / Asgardeo, Okta custom
authorization servers, Keycloak v2 (confidential clients only), PingFederate, PingOne,
Curity, PingAM. **Not** Entra ID.

---

## Configuration

All keys live in `[ai_workspace.auth.oidc.token_exchange]`, a child of the login table.
The minimal WSO2 / Asgardeo case is two keys:

```toml
[ai_workspace.auth.oidc]
# ... authority / client_id / client_secret / redirect_url as usual ...

[ai_workspace.auth.oidc.token_exchange]
enabled  = true
audience = "platform-api"
```

`client_id`, `client_secret` and `scope` fall back to the login client's, and
`token_endpoint` to the one discovered from `authority` — set them only when the
exchange differs from login.

### Why a child table rather than more keys on `[auth.oidc]`

The two calls are genuinely two OAuth clients. Login posts `grant_type=authorization_code`
with the login `client_id`; the exchange posts `grant_type=…:token-exchange` with a
`client_id` an STS commonly registers separately. A table expresses that. It is a
*child* of `[auth.oidc]` rather than a sibling because both calls go to the same issuer
and four keys inherit from the parent, which is what keeps the common single-application
deployment down to two lines.

Full per-key operator documentation is in `configs/config-template.toml`; the design
notes are in `bff/TOKEN_EXCHANGE.md`.

### Platform API side

**No code change needed.** Its IDP authenticator already validates issuer and audience
when configured, so two values must agree:

```toml
[platform_api.auth.idp]
issuer   = ["https://iam.example.com/oauth2/token"]   # who issued the EXCHANGED token
audience = ["platform-api"]                            # must match the exchange audience
```

A mismatch 401s every request and looks like a broken exchange when the fault is
upstream — the single most likely misconfiguration, called out in the docs.

---

## Security properties

These are deliberate, not incidental.

- **Fail-closed, with no fallback.** No path returns the unexchanged subject token. A
  failed exchange fails the request; forwarding the login token instead would reach the
  Platform API with the wrong audience and, on a role-mode IDP, no platform
  authorization at all. A misconfiguration takes the UI down rather than silently
  downgrading it, and is validated at startup wherever it can be.
- **Rejection vs. unavailability are distinguished, and transients are retried.** An
  IDP *rejection* destroys the session and returns 401 — it can never produce an
  upstream token. An *unavailable* IDP keeps the session and returns 502, since logging
  the user out over a transient blip would be self-inflicted. Neither response carries
  the IDP's reason: whether the subject or the target was refused maps out the
  deployment's trust configuration. The split cannot be made on the status code alone —
  Asgardeo answers `invalid_grant` (a 4xx) when *its own* fetch of the trusted issuer's
  JWKS fails, which is a fault that clears by itself. That case, and `408`/`429`, are
  classified as unavailable, and the exchange retries up to three times within the
  existing 15s budget before giving up.
- **The exchange never follows a redirect.** Its POST body carries the client secret and
  the subject token, and Go replays a POST body on `307`/`308` — so a redirect would hand
  both to whatever host the response names. The exchanger uses a copy of the shared
  client with `CheckRedirect` disabled, leaving that client's SSRF-guarded transport
  intact (`ssrf-prevention.md` directive 2).
- **`token_endpoint` must be HTTPS.** The same body makes plaintext a credential
  disclosure, so a non-loopback `http://` endpoint now fails startup rather than emitting
  a warning. Loopback stays permitted — the request never reaches a network there.
- **No credential reaches a log or an error string.** The subject token, the exchanged
  token, and the client secret are never logged on any path, and
  `TestExchangeErrorsDoNotContainTokens` asserts no error value carries one — errors
  propagate outward, so they are the easiest way for a token to end up in someone's
  log aggregator. The IDP's `error_description` *is* logged, because it is the only
  field that separates an untrusted issuer from a failed key fetch from a rejected
  audience — but it is redacted first, by exact value (the subject token and client
  secret this call sent, which covers opaque tokens that no pattern could match) with a
  JWT-shaped regex as a backstop (GO-AUTH-003).
- **The exchanged token never reaches the browser.** `/api/session` reports the
  exchanged *scopes* (they are what the Platform API authorizes) but not the token.
- **A cached token cannot outlive its inputs.** The cache entry carries a fingerprint
  of the settings that determine what the IDP mints, so a config change invalidates it;
  an unknown expiry is treated as uncacheable rather than as valid; and a rotated login
  token drops the cached exchange rather than carrying it forward.
- **No refresh token is requested for the exchanged token.** RFC 8693 §2.2.1 advises
  against one when trading temporary credentials, and it would outlive the login session
  it derives from — revoking the upstream session would stop revoking API access. The
  BFF re-exchanges from the session's own subject token instead.
- **Single-flight per session.** The burst of parallel calls the SPA makes on page load
  hits the IDP once, not once per request.

---

## Testing

`go test ./...` in `portals/ai-workspace/bff` — all packages pass.

| File | Covers |
|---|---|
| `internal/auth/tokenexchange_test.go` | Wire format pinned per grant, expiry/scope resolution, error classification, no-token-leak guard, transient-vs-rejection split (`408`/`429`/5xx/JWKS-fetch), retry behaviour, redaction of JWT *and* opaque credentials from the logged description, redirects refused. |
| `internal/config/token_exchange_test.go` | Defaults-off compatibility, credential inheritance, every validation rule, keys pinned to the right table, plaintext `token_endpoint` refused with loopback exempt. |
| `internal/config/debug_overlay_test.go` | `configs/config-debug.toml` stays inert on an empty environment and wires up correctly once the variables are exported. |
| `internal/server/token_exchange_test.go` | End-to-end: fail-closed, 401-vs-502 classification, caching, single-flight, exchanged scopes on `/api/session`. |

The IDP is stubbed throughout, so the wire format is pinned against the *specifications
and vendor documentation*, not against a live server.

### Trying it locally

`configs/config.toml` carries a ready-made, entirely `{{ env }}`-driven `[auth.oidc]` +
`[auth.oidc.token_exchange]` pair with inert defaults, so no credential is committed and
the quickstart stays in basic mode until the variables are set. The tables live in the
shipped file rather than the debug overlay because that file is the one the container
mounts (`docker-compose.yaml`) and the overlay is never mounted — a compose deployment
could not otherwise enable the feature without editing a mounted file. For
docker-compose, set them in the gitignored `api-platform.env` the services already load;
for `make bff-run`, export them:

```bash
export APIP_AIW_AUTH_MODE=oidc
export APIP_AIW_AUTH_OIDC_AUTHORITY=...          # discovery base
export APIP_AIW_AUTH_OIDC_CLIENT_ID=...          # the authorization_code client
export APIP_AIW_AUTH_OIDC_CLIENT_SECRET=...
export APIP_AIW_AUTH_OIDC_TOKEN_EXCHANGE_ENABLED=true
export APIP_AIW_AUTH_OIDC_TOKEN_EXCHANGE_AUDIENCE=platform-api
make bff-run
```

`configs/config-debug.toml` keeps only genuinely debug-shaped overrides (the BFF's own
port, the local Platform API URL). No credential lives in any tracked file:
`TestShippedConfig_QuickstartLoadsWithNoEnv` and `TestDebugOverlay` both assert the
shipped defaults stay in basic mode with the exchange off on an empty environment.

---

## Not done in this PR

**Entra ID (`jwt_bearer`) is still unverified against a live tenant.** The OBO flow and
the `api://<app-id>/.default` scope need a round trip before that grant is trusted in
production. The RFC 8693 path no longer does — see below.

**Not verified: switching an existing role-mode deployment to `mode = "scope"` on both
sides.** That is the payoff this PR is for, and it needs a deployment that is currently
running the mirror.

**The Helm chart does not render the new table yet.** `ai-workspace-ui-helm-chart`'s
configmap template emits `[ai_workspace.auth.oidc]` but has no token-exchange values, so
a Kubernetes deployment cannot enable the feature from `values.yaml`. Config-file and
local deployments are unaffected. Worth a follow-up once the live-IDP verification above
settles the key set.

---

## Verified end to end against Asgardeo

The RFC 8693 path has been run against a live two-organization Asgardeo setup — Org A
authenticates the user, Org B trusts Org A as a trusted token issuer and mints the
`ap:*` token — with the Platform API validating the result via JWKS. The walkthrough is
`bff/ASGARDEO_TOKEN_EXCHANGE_DEMO.md`.

Five things had to be true that the code cannot check for you, each of which failed
first and is now documented in that file's troubleshooting table:

1. The login application must issue **JWT** access tokens, not opaque ones — an opaque
   subject token gets `Error while parsing the JWT` from the exchange.
2. The login application's **Audience** list must contain the exchange organization's
   token endpoint, or the exchange answers `invalid_target` about the *subject token's*
   `aud` claim — not about the `audience` parameter, which should stay empty for
   Asgardeo.
3. The trusted token issuer's **Issuer** must equal the subject token's `iss` exactly,
   including the `/t/` segment Asgardeo's organization endpoints carry.
4. Scopes are granted to a **user through a role**, not to the application: authorizing
   the exchanging app against the API resource only makes them requestable. Without the
   role the exchange succeeds and returns an empty `scope`, which surfaces as a 403 from
   the Platform API rather than as an exchange failure.
5. The subject must resolve to a **local user** in the exchange organization (implicit
   account linking), or the exchange answers `Use mapped local subject is mandatory but
   a local user couldn'''t be found`.

**One deployment caveat this exposed.** The exchanged token's `org_id` is the *exchange*
organization's, not the user's home org, because Asgardeo issues it in the context of
the linked local user. The Platform API scopes every query by that claim, so on this
topology all users land in one organization. That is an artifact of the two-org demo,
not of the exchange: an STS that propagates the subject'''s own org claim keeps
multi-tenancy intact, and the Platform API reads the claim name from
`[platform_api.auth.claim_mappings] organization`.

---

## Reviewing

Roughly in dependency order:

| File | Change |
|---|---|
| `internal/config/config.go` | `TokenExchangeConfig` as the `token_exchange` sub-table on `OIDCConfig`, credential/scope inheritance in `normalize`, `validateTokenExchange` (including the HTTPS-only `token_endpoint` rule). |
| `internal/config/default_config.go` | Defaults — off, WSO2-shaped token types. |
| `internal/auth/tokenexchange.go` | The exchanger: request building per grant, response parsing, error classification, retry on transient failures, credential redaction before logging, redirects refused. |
| `internal/auth/oidc.go` | Exposes `TokenEndpoint()` so the exchange reuses login's discovery. |
| `internal/session/store.go` | `ExchangedToken` on the session and its `Usable` cache-validity rule. |
| `internal/server/server.go` | Builds the exchanger; single-flight map; startup log line. |
| `internal/server/handlers.go` | `upstreamToken`, cache + single-flight, eager exchange at login, exchanged scopes on `/api/session`, 401-vs-502 mapping. `serveProxy` applies the exchange to both the primary and the cloud-analytics hop — both authorize the forwarded token, so the cloud hop must not fall back to the login token. |
| `internal/server/composite_handlers.go` | The two composite endpoints bypass `handleProxy`, so they resolve the upstream token the same way. |
| `configs/config-template.toml` | Operator documentation of `[auth.oidc.token_exchange]`. |
| `configs/config.toml` | Env-driven `[auth.oidc]` + `[auth.oidc.token_exchange]` with inert defaults, so a container deployment can enable the feature without editing a mounted file. |
| `configs/config-debug.toml` | Debug-shaped overrides only (BFF port, local Platform API URL). |
| `platform-api/config/config.toml` | Env-driven `[auth.idp]` + `[auth.claim_mappings]`, defaults inert so `auth.mode` stays `"file"`. |
| `bff/TOKEN_EXCHANGE.md` | Design notes, per-key reference, local-testing recipe, open items. |

### DIFF
--- a/portals/ai-workspace/bff/internal/auth/oidc.go
+++ b/portals/ai-workspace/bff/internal/auth/oidc.go
@@ -123,6 +123,11 @@ func (o *OIDC) Close() {
 	o.closeOnce.Do(func() { close(o.done) })
 }
 
+// TokenEndpoint is the endpoint discovered from the issuer. Exposed so a token
+// exchange configured without an explicit endpoint override can post to the same
+// IDP the user logged in to, without repeating discovery.
+func (o *OIDC) TokenEndpoint() string { return o.disco.TokenEndpoint }
+
 func fetchDiscovery(ctx context.Context, client *http.Client, issuer string) (discoveryDoc, error) {
 	u := strings.TrimRight(issuer, "/") + "/.well-known/openid-configuration"
 	req, err := http.NewRequestWithContext(ctx, http.MethodGet, u, nil)
--- a/portals/ai-workspace/bff/internal/auth/tokenexchange.go
+++ b/portals/ai-workspace/bff/internal/auth/tokenexchange.go
@@ -41,9 +41,11 @@ const (
 	TokenTypeRefresh     = "urn:ietf:params:oauth:token-type:refresh_token"
 )
 
+// The wire values, aliased from config so the URI the request carries and the URI an
+// operator may spell in grant_type can never drift apart.
 const (
-	grantURITokenExchange = "urn:ietf:params:oauth:grant-type:token-exchange"
-	grantURIJWTBearer     = "urn:ietf:params:oauth:grant-type:jwt-bearer"
+	grantURITokenExchange = config.GrantURITokenExchange
+	grantURIJWTBearer     = config.GrantURIJWTBearer
 )
 
 const (
--- a/portals/ai-workspace/bff/internal/auth/tokenexchange.go
+++ b/portals/ai-workspace/bff/internal/auth/tokenexchange.go
@@ -1,0 +1,307 @@
+/*
+ * Copyright (c) 2026, WSO2 LLC. (https://www.wso2.com).
+ *
+ * WSO2 LLC. licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the
+ * License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package auth
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"log/slog"
+	"net/http"
+	"net/url"
+	"slices"
+	"strings"
+	"time"
+
+	"ai-workspace-bff/internal/config"
+	"ai-workspace-bff/internal/session"
+)
+
+// RFC 8693 §3 token type identifiers.
+const (
+	TokenTypeAccessToken = "urn:ietf:params:oauth:token-type:access_token"
+	TokenTypeJWT         = "urn:ietf:params:oauth:token-type:jwt"
+	TokenTypeIDToken     = "urn:ietf:params:oauth:token-type:id_token"
+	TokenTypeRefresh     = "urn:ietf:params:oauth:token-type:refresh_token"
+)
+
+const (
+	grantURITokenExchange = "urn:ietf:params:oauth:grant-type:token-exchange"
+	grantURIJWTBearer     = "urn:ietf:params:oauth:grant-type:jwt-bearer"
+)
+
+const (
+	exchangeTimeout          = 15 * time.Second
+	maxExchangeResponseBytes = 1 << 20
+)
+
+// ErrExchangeUnavailable means the IDP could not be reached or failed server-side;
+// retrying may succeed, so callers keep the session and report 502.
+var ErrExchangeUnavailable = errors.New("token exchange upstream unavailable")
+
+// ErrExchangeRejected means the IDP refused the exchange. The session can never
+// produce an upstream token, so callers destroy it and report 401.
+var ErrExchangeRejected = errors.New("token exchange rejected by identity provider")
+
+// Exchanger trades the login token for one minted for the Platform API. Its
+// settings come from [ai_workspace.auth.oidc.token_exchange].
+//
+// No method returns the unexchanged subject token: a failed exchange is an error,
+// never a fallback. Forwarding the login token instead would reach the Platform API
+// with the wrong audience and, on an IDP that mints no ap:* scopes, no platform
+// authorization at all.
+type Exchanger struct {
+	client          *http.Client
+	cfg             config.TokenExchangeConfig
+	endpoint        string
+	requestedScopes []string
+}
+
+// NewExchanger builds an Exchanger for an already-validated config. endpoint is the
+// resolved token endpoint, so discovery stays with the OIDC client.
+func NewExchanger(client *http.Client, cfg config.TokenExchangeConfig, endpoint string) *Exchanger {
+	return &Exchanger{
+		client:          client,
+		cfg:             cfg,
+		endpoint:        endpoint,
+		requestedScopes: strings.Fields(cfg.Scopes),
+	}
+}
+
+func (e *Exchanger) Endpoint() string           { return e.endpoint }
+func (e *Exchanger) CacheEnabled() bool         { return e.cfg.CacheEnabled }
+func (e *Exchanger) MinValidity() time.Duration { return e.cfg.MinValidity }
+
+// ConfigFingerprint identifies the settings that determine what the IDP mints, so a
+// cached token is not reused after they change. Credentials are excluded: they
+// authenticate the BFF without altering the issued token.
+func (e *Exchanger) ConfigFingerprint() string {
+	return strings.Join([]string{e.cfg.GrantType, e.cfg.Audience, e.cfg.Resource, e.cfg.Scopes}, "\x1f")
+}
+
+// Result is one completed exchange. A zero Expiry means the IDP supplied no lifetime,
+// which callers must treat as uncacheable rather than as expired.
+type Result struct {
+	AccessToken string
+	Expiry      time.Time
+	Scopes      []string
+}
+
+type exchangeResponse struct {
+	AccessToken     string `json:"access_token"`
+	IssuedTokenType string `json:"issued_token_type"`
+	TokenType       string `json:"token_type"`
+	ExpiresIn       int64  `json:"expires_in"`
+	Scope           string `json:"scope"`
+	RefreshToken    string `json:"refresh_token"`
+}
+
+type exchangeError struct {
+	Code        string `json:"error"`
+	Description string `json:"error_description"`
+}
+
+// Exchange trades subjectToken for a Platform API token. subjectToken is a live
+// credential and is never logged.
+func (e *Exchanger) Exchange(ctx context.Context, subjectToken string) (*Result, error) {
+	if subjectToken == "" {
+		return nil, fmt.Errorf("%w: no subject token", ErrExchangeRejected)
+	}
+
+	ctx, cancel := context.WithTimeout(ctx, exchangeTimeout)
+	defer cancel()
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, e.endpoint,
+		strings.NewReader(e.buildForm(subjectToken).Encode()))
+	if err != nil {
+		return nil, fmt.Errorf("%w: %v", ErrExchangeUnavailable, err)
+	}
+	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+	req.Header.Set("Accept", "application/json")
+
+	res, err := e.client.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("%w: %v", ErrExchangeUnavailable, err)
+	}
+	defer res.Body.Close()
+
+	body, err := io.ReadAll(io.LimitReader(res.Body, maxExchangeResponseBytes))
+	if err != nil {
+		return nil, fmt.Errorf("%w: reading response: %v", ErrExchangeUnavailable, err)
+	}
+	if res.StatusCode != http.StatusOK {
+		return nil, e.classifyError(res.StatusCode, body)
+	}
+
+	var tok exchangeResponse
+	if err := json.Unmarshal(body, &tok); err != nil {
+		return nil, fmt.Errorf("%w: decoding response: %v", ErrExchangeUnavailable, err)
+	}
+	if tok.AccessToken == "" {
+		return nil, fmt.Errorf("%w: response carried no access_token", ErrExchangeUnavailable)
+	}
+	// issued_token_type is an RFC 8693 field; Entra's OBO response has none.
+	if e.cfg.GrantType == config.GrantTokenExchange {
+		if err := validateIssuedTokenType(tok.IssuedTokenType); err != nil {
+			return nil, fmt.Errorf("%w: %v", ErrExchangeRejected, err)
+		}
+	}
+
+	claims := session.DecodeJWTClaims(tok.AccessToken)
+	return &Result{
+		AccessToken: tok.AccessToken,
+		Expiry:      expiryFrom(tok.ExpiresIn, claims),
+		Scopes:      e.grantedScopes(tok.Scope, claims),
+	}, nil
+}
+
+func (e *Exchanger) buildForm(subjectToken string) url.Values {
+	form := url.Values{
+		"client_id":     {e.cfg.ClientID},
+		"client_secret": {e.cfg.ClientSecret},
+	}
+
+	// Entra ID does not implement RFC 8693 — it rejects that grant. Its on-behalf-of
+	// flow is RFC 7523: the subject travels as `assertion`, and the target API is
+	// named through `scope`, so audience/resource have no place in the request.
+	if e.cfg.GrantType == config.GrantJWTBearer {
+		form.Set("grant_type", grantURIJWTBearer)
+		form.Set("assertion", subjectToken)
+		form.Set("requested_token_use", "on_behalf_of")
+		if e.cfg.Scopes != "" {
+			form.Set("scope", e.cfg.Scopes)
+		}
+		return form
+	}
+
+	form.Set("grant_type", grantURITokenExchange)
+	form.Set("subject_token", subjectToken)
+	form.Set("subject_token_type", e.cfg.SubjectTokenType)
+	form.Set("requested_token_type", e.cfg.RequestedTokenType)
+	// config.validate permits at most one: IDPs read them differently (PingFederate
+	// treats them as distinct selectors), so sending both leaves the target ambiguous.
+	if e.cfg.Audience != "" {
+		form.Set("audience", e.cfg.Audience)
+	}
+	if e.cfg.Resource != "" {
+		form.Set("resource", e.cfg.Resource)
+	}
+	if e.cfg.Scopes != "" {
+		form.Set("scope", e.cfg.Scopes)
+	}
+	return form
+}
+
+// classifyError logs the IDP's reason and returns a sentinel. The reason stays
+// internal: whether the subject or the target was refused maps out the deployment's
+// trust configuration.
+func (e *Exchanger) classifyError(status int, body []byte) error {
+	var ee exchangeError
+	_ = json.Unmarshal(body, &ee)
+
+	attrs := []any{"status", status, "grant_type", e.cfg.GrantType, "endpoint", e.endpoint}
+	if ee.Code != "" {
+		attrs = append(attrs, "idp_error", ee.Code)
+	}
+	if ee.Description != "" {
+		attrs = append(attrs, "idp_error_description", ee.Description)
+	}
+
+	if status >= http.StatusInternalServerError {
+		slog.Error("token exchange failed: identity provider error", attrs...)
+		return fmt.Errorf("%w: status %d", ErrExchangeUnavailable, status)
+	}
+	// RFC 8693 §2.2.2. An operator must fix this, so it is not a per-user warning.
+	if ee.Code == "invalid_target" {
+		slog.Error("token exchange rejected: audience/resource not accepted by the IDP — "+
+			"register it on the exchanging application, or correct "+
+			"[auth.oidc.token_exchange] audience / resource",
+			append(attrs, "configured_audience", e.cfg.Audience, "configured_resource", e.cfg.Resource)...)
+		return fmt.Errorf("%w: invalid_target", ErrExchangeRejected)
+	}
+	slog.Warn("token exchange rejected", attrs...)
+	return fmt.Errorf("%w: %s", ErrExchangeRejected, ee.Code)
+}
+
+// grantedScopes prefers the response's scope, which RFC 8693 §2.2.1 makes REQUIRED
+// whenever it differs from the request, and falls back to the issued token's claim.
+//
+// The widening check cannot be an enforcement point — only the IDP knows the subject
+// token's grant — but scopes the BFF never asked for mean the exchanging application
+// is over-provisioned, which is worth surfacing.
+func (e *Exchanger) grantedScopes(granted string, claims map[string]any) []string {
+	scopes := strings.Fields(granted)
+	if len(scopes) == 0 {
+		scopes = scopeClaim(claims)
+	}
+	if len(e.requestedScopes) == 0 {
+		return scopes
+	}
+	var extra []string
+	for _, s := range scopes {
+		if !slices.Contains(e.requestedScopes, s) {
+			extra = append(extra, s)
+		}
+	}
+	if len(extra) > 0 {
+		slog.Warn("token exchange returned scopes that were not requested — "+
+			"the exchanging application may be over-provisioned at the IDP",
+			"unrequested_scopes", extra)
+	}
+	return scopes
+}
+
+func validateIssuedTokenType(t string) error {
+	switch t {
+	case TokenTypeAccessToken, TokenTypeJWT:
+		return nil
+	case "":
+		return errors.New("response omitted the required issued_token_type")
+	default:
+		return fmt.Errorf("issued_token_type %q is not usable as an upstream bearer token", t)
+	}
+}
+
+// expiryFrom falls back to the exp claim because RFC 8693 only RECOMMENDS expires_in.
+func expiryFrom(expiresIn int64, claims map[string]any) time.Time {
+	if expiresIn > 0 {
+		return time.Now().Add(time.Duration(expiresIn) * time.Second)
+	}
+	return session.ExpiryFromClaims(claims)
+}
+
+func scopeClaim(claims map[string]any) []string {
+	raw, ok := claims["scope"]
+	if !ok {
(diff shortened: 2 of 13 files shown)


Thushani-Jayasekera:
# Demoing token exchange with two Asgardeo applications

A self-contained way to exercise the BFF's `[ai_workspace.auth.oidc.token_exchange]`
path against a live IdP, using nothing but Asgardeo — no custom STS to write, deploy,
or make publicly reachable.

**The arrangement.** One Asgardeo application mints the *subject token* (standing in for
the corporate IdP that logs the user in). A second Asgardeo application accepts that
token over the token-exchange grant and issues the token the Platform API actually
authorizes against.

```
Org A (issuer)                          Org B (exchange)
┌──────────────────┐                    ┌──────────────────┐
│  App A           │                    │  App B           │
│  client_creds    │─── subject JWT ───▶│  token_exchange  │──▶ access token
│  or password     │    aud = Org B's   │  grant enabled   │    scope = ap:*
│  grant           │    token endpoint  │                  │
└──────────────────┘                    └──────────────────┘
         ▲                                       │
         │        Org B trusts Org A as a        │
         └──── "Trusted Token Issuer" ───────────┘
                (verified via Org A's JWKS)
```

## Why two organizations, not two apps in one

The trusted-token-issuer configuration is keyed on the **`iss` claim** of the incoming
JWT. Every token Asgardeo issues from an organization carries the same issuer —
`https://dev.api.asgardeo.io/t/<org>/oauth2/token` — regardless of which application minted
it. Two apps in the *same* organization therefore produce tokens Asgardeo cannot tell
apart by issuer, and there is nothing meaningful to register as an external trusted
issuer.

So: **App A lives in Organization A, App B lives in Organization B.** A free second
Asgardeo organization is enough. (A root org plus one of its sub-organizations also
works and is closer to a real B2B topology; the endpoint differences are noted in
[Variant: root org and sub-organization](#variant-root-org-and-sub-organization).)

---

## What you need

- Two Asgardeo organizations. Call them `orgathushani` and `orgbthushani` below — substitute your real
  organization names everywhere.
- Admin access to both consoles.
- `curl` and `jq`.

Throughout, the two token endpoints are:

| | |
|---|---|
| Org A token endpoint | `https://dev.api.asgardeo.io/t/orgathushani/oauth2/token` |
| Org B token endpoint | `https://dev.api.asgardeo.io/t/orgbthushani/oauth2/token` |

> Using the developer environment? The host is `dev.api.asgardeo.io` instead —
> that is what the checked-in `config.toml` currently points at.

---

## Step 1 — Create App A, the token issuer (Organization A)

In **Org A's** console:

1. **Applications → New Application → M2M Application**. Name it `Demo Issuer`.
   (An M2M app gives you a token from a single `client_credentials` call, with no
   browser round trip. If you want the exchanged token to represent a *user* rather
   than a machine, see [Getting a user-context subject token](#getting-a-user-context-subject-token)
   below.)

2. Open the app's **Protocol** tab and confirm:
   - **Allowed grant types** includes **Client Credentials**.
   - Under **Access Token**, **Token type** is **JWT**. Token exchange needs a JWT
     subject token; an opaque token cannot be verified by the other organization.

3. Still on the **Protocol** tab, find **Access Token → Audience** and add:

   ```
   https://dev.api.asgardeo.io/t/orgbthushani/oauth2/token
   ```

   This is the step everything else depends on. Asgardeo validates the incoming token's
   `aud` claim against the consuming organization's own issuer value; setting Org B's
   token endpoint as an audience on App A is what makes that check pass. (If the `aud`
   claim does not carry Org B's issuer value, Asgardeo falls back to comparing `aud`
   against the **Alias** you configure in step 2 — either route works, but this one is
   the least surprising.)

4. Click **Update**, then note two things from the app:
   - **Access token type** must be **JWT** (Protocol tab → Access Token). Asgardeo's
     default is opaque, and an opaque subject token fails the exchange with
     `Error while parsing the JWT`.
   - **Client ID** and **Client secret** (Protocol tab) → `ISSUER_CLIENT_ID` / `ISSUER_CLIENT_SECRET`.
   - The **JWKS endpoint** (Info tab) → you will paste this into Org B.
     It looks like `https://dev.api.asgardeo.io/t/orgathushani/oauth2/jwks`.

**Check it before moving on.** Get a token and decode it:

```bash
export ISSUER_CLIENT_ID=5JCoCtfL8JQativQJ25Eupvlaz4a
export ISSUER_CLIENT_SECRET=clge3JMrjGBrLdPRycdfXlMyPKfj1unc1Lx92W6fECMa

SUBJECT_TOKEN=$(curl -s \
  --user "$ISSUER_CLIENT_ID:$ISSUER_CLIENT_SECRET" \
  -H 'Content-Type: application/x-www-form-urlencoded' \
  -d 'grant_type=client_credentials' \
  'https://dev.api.asgardeo.io/t/orgathushani/oauth2/token' | jq -r .access_token)

echo "$SUBJECT_TOKEN" | cut -d. -f2 | base64 -d 2>/dev/null | jq .
```

You are looking for exactly two things in the payload:

```json
{
  "iss": "https://dev.api.asgardeo.io/t/orgathushani/oauth2/token",
  "aud": ["https://dev.api.asgardeo.io/t/orgbthushani/oauth2/token", "<ISSUER_CLIENT_ID>"],
  "sub": "...",
  "client_id": "<ISSUER_CLIENT_ID>"
}
```

If `aud` does not contain Org B's token endpoint, step 3 did not take — go back and fix
it now. If the token is not three dot-separated segments, it is opaque, not JWT; fix
step 2.

---

## Step 2 — Register Org A as a trusted token issuer (Organization B)

Switch to **Org B's** console.

1. **Connections → New Connection → Trusted Token Issuer → Create**.

2. Fill in:

   | Field | Value |
   |---|---|
   | **Name** | `Org A Issuer` (any unique label) |
   | **Issuer** | `https://dev.api.asgardeo.io/t/orgathushani/oauth2/token` — must equal the `iss` claim you decoded above, character for character |
   | **Alias** | `https://dev.api.asgardeo.io/t/orgbthushani/oauth2/token` |

3. On the certificate step, choose **JWKS endpoint** and paste Org A's JWKS URL:

   ```
   https://dev.api.asgardeo.io/t/orgathushani/oauth2/jwks
   ```

   Both organizations are public Asgardeo endpoints, so Org B can fetch this directly —
   this is the whole reason the two-app demo is less work than running your own STS,
   where you would have to expose a JWKS endpoint or manage a PEM certificate by hand.

4. **Finish**.

---

## Step 3 — Create App B, the token consumer (Organization B)

Still in **Org B's** console:

1. **Applications → New Application → Standard-Based Application → OAuth2/OpenID Connect**.
   Name it `Demo Exchange`.

   > Do **not** pick Single-Page Application. Token exchange is not supported for SPAs.

2. On the **Protocol** tab, add **Token Exchange** to **Allowed grant types**. Add
   **Refresh Token** as well if you want a `refresh_token` back. **Update**.

3. Note the **Client ID** and **Client secret** → `EXCHANGE_CLIENT_ID` /
   `EXCHANGE_CLIENT_SECRET`.

4. If you want the exchanged token to carry `ap:*` scopes — which is the point of the
   exercise for the BFF — register them as an API resource and authorize App B against
   it: **API Resources → New API Resource**, add scopes such as `ap:rest_api:manage`,
   then on App B's **API Authorization** tab, authorize that resource and select the
   scopes.

   > This step alone yields a token with **no scopes at all**. Authorizing the
   > application makes the scopes *requestable*; they are granted to a user through a
   > role. Steps 5 and 6 are what actually put them in the token.

5. **User Management → Users → Add User.** Create a user whose lookup attribute —
   email, normally — matches the Org A user you will log in as. The exchanged token is
   an `APPLICATION_USER` token: it represents this local user, and this is the account
   the role attaches to.

6. **User Management → Roles → New Role.** Audience **Application → Demo Exchange**,
   select the API resource from step 4 and tick every scope the workspace needs, then
   **assign the role to the user from step 5**.

7. Back on **Connections → Org A Issuer → Advanced**, enable **Implicit account
   linking** and set the lookup attribute to the one you matched in step 5. Without it
   the exchange fails with `Use mapped local subject is mandatory but a local user
   couldn't be found`.

   > Pick a lookup attribute that is genuinely unique per user. It is trusted as
   > verified, so anyone able to mint a subject token carrying that value becomes this
   > local user.

---

## Step 4 — Exchange the token

```bash
export EXCHANGE_CLIENT_ID=AGCHzdWsdhqWkxIZze0smN1c8doa
export EXCHANGE_CLIENT_SECRET=X

curl -s --location 'https://dev.api.asgardeo.io/t/orgbthushani/oauth2/token' \
  --header 'Content-Type: application/x-www-form-urlencoded' \
  --user "$EXCHANGE_CLIENT_ID:$EXCHANGE_CLIENT_SECRET" \
  --data-urlencode 'grant_type=urn:ietf:params:oauth:grant-type:token-exchange' \
  --data-urlencode 'subject_token_type=urn:ietf:params:oauth:token-type:jwt' \
  --data-urlencode 'requested_token_type=urn:ietf:params:oauth:token-type:access_token' \
  --data-urlencode "subject_token=$SUBJECT_TOKEN" \
  --data-urlencode 'scope=ap:rest_api:manage' | jq .
```

Success:

```json
{
  "access_token": "eyJ4NXQiOi...",
  "scope": "ap:rest_api:manage",
  "token_type": "Bearer",
  "expires_in": 3600
}
```

Decode it and confirm `aud` is now App B's client ID and the scopes are the ones the
Platform API expects. Introspect it if you want the authoritative view:

```bash
curl -s --user "$EXCHANGE_CLIENT_ID:$EXCHANGE_CLIENT_SECRET" \
  -H 'Content-Type: application/x-www-form-urlencoded' \
  -d "token=$ACCESS_TOKEN" \
  'https://dev.api.asgardeo.io/t/orgbthushani/oauth2/introspect' | jq .
```

---

## Step 5 — Point the BFF at it

With both applications working from `curl`, the BFF config is a direct transcription.
No credential goes in a tracked file: `configs/config.toml` reads every value from the
environment, so set them in the gitignored `api-platform.env` that docker-compose
already loads (or export them for `make bff-run`).

```bash
# Login: Org A. Exchange: Org B.
APIP_AIW_AUTH_MODE=oidc
APIP_AIW_AUTH_OIDC_AUTHORITY=https://dev.api.asgardeo.io/t/orgathushani/oauth2/token
APIP_AIW_AUTH_OIDC_CLIENT_ID=<LOGIN_CLIENT_ID>
APIP_AIW_AUTH_OIDC_CLIENT_SECRET=<LOGIN_CLIENT_SECRET>

APIP_AIW_AUTH_OIDC_TOKEN_EXCHANGE_ENABLED=true
APIP_AIW_AUTH_OIDC_TOKEN_EXCHANGE_TOKEN_ENDPOINT=https://dev.api.asgardeo.io/t/orgbthushani/oauth2/token
APIP_AIW_AUTH_OIDC_TOKEN_EXCHANGE_CLIENT_ID=<EXCHANGE_CLIENT_ID>
APIP_AIW_AUTH_OIDC_TOKEN_EXCHANGE_CLIENT_SECRET=<EXCHANGE_CLIENT_SECRET>
APIP_AIW_AUTH_OIDC_TOKEN_EXCHANGE_SCOPE=ap:rest_api:manage ap:project:manage
```

Leave the exchange `audience` **unset** for Asgardeo. The issued `aud` is then the
exchanging application's client ID, which is what `[platform_api.auth.idp] audience`
must carry. Setting it to anything Asgardeo does not accept — including the exchange
org's token endpoint — is rejected as `Invalid audience value provided`.

The login application here is a **separate application from App A**: App A is M2M and
mints subject tokens for the `curl` walkthrough only, while login needs
`authorization_code`, which an M2M app cannot do. Create a Standard-Based or Traditional
Web application in Org A, register the redirect URL, set its access token type to
**JWT**, and add Org B's token endpoint to its **Access Token → Audience** list — the
same `aud` treatment App A got in step 1.

### Platform API side

The Platform API must verify the **exchanged** token, which Org B signs — not its own
key. In `api-platform.env`:

```bash
APIP_CP_AUTH_MODE=idp
APIP_CP_AUTH_IDP_JWKS_URL=https://dev.api.asgardeo.io/t/orgbthushani/oauth2/jwks
APIP_CP_AUTH_IDP_ISSUER=https://dev.api.asgardeo.io/t/orgbthushani/oauth2/token
APIP_CP_AUTH_IDP_AUDIENCE=<EXCHANGE_CLIENT_ID>

# No built-in defaults — unset means the claim is simply not extracted.
APIP_CP_AUTH_CLAIM_ORGANIZATION=org_id
APIP_CP_AUTH_CLAIM_USER_ID=sub
APIP_CP_AUTH_CLAIM_USERNAME=username
APIP_CP_AUTH_CLAIM_SCOPE=scope
APIP_CP_AUTH_CLAIM_ROLES=roles
```

Left in `file` mode it verifies with `auth.jwt.public_key_file` and every request fails
with `crypto/rsa: verification error`.

> **`org_id` is Org B's, not the user's home org.** Asgardeo issues the token in the
> context of the linked local user, so every user of this demo resolves to the same
> Platform API organization. That row must exist, and the collapse is an artifact of the
> two-org topology — an STS that propagates the subject's own org claim keeps
> multi-tenancy intact.

Once it works end to end, set `[auth.authorization] mode = "scope"`: both the BFF and the
Platform API now read scopes off the exchanged token, and the hand-mirrored
`role-to-scope-mapping.yaml` grant table stops being load-bearing.

---

## Getting a user-context subject token

`client_credentials` gives you an application token: `sub` is the client ID and there is
no user behind it. For a demo that exercises user identity, replace App A with a
**Traditional Web Application** or **Mobile Application** in Org A and use either grant:

**Password grant** (fastest for a demo; enable **Password** under allowed grant types):

```bash
SUBJECT_TOKEN=$(curl -s \
  --user "$ISSUER_CLIENT_ID:$ISSUER_CLIENT_SECRET" \
  -H 'Content-Type: application/x-www-form-urlencoded' \
  -d 'grant_type=password' \
  -d 'username=alice@example.com' \
  -d 'password=...' \
  -d 'scope=openid' \
  'https://dev.api.asgardeo.io/t/orgathushani/oauth2/token' | jq -r .access_token)
```

**Authorization code** — the realistic path, and what the BFF actually does.

Either way, note the sharpest limitation in this whole feature: **Asgardeo copies only
the `sub` claim** from the subject token into the exchanged token. No email, no roles,
nothing else crosses the boundary by default. To have the exchanged token represent a
real user in Org B:

1. On App B's **User Attributes** tab, under *Attribute Resolution for Linked Accounts*,
   select **Use linked local account attributes**. Tick **Require linked local account**
   if the exchange should fail outright when no matching Org B user exists.
2. On the trusted token issuer's **Advanced** tab, enable **Implicit account linking**
   and choose a lookup attribute — `email` is the usual choice.

> Pick a lookup attribute that is genuinely unique per user. It is trusted as verified
> by the issuing organization, and links created this way **cannot be deleted by an
> administrator** afterwards.

---

## Variant: root org and sub-organization

If you would rather model the B2B topology than use two independent organizations, the
only change is the exchange endpoint — insert `/o/<ORG_ID>` between the tenant segment
and `/oauth2/token`:

```
https://dev.api.asgardeo.io/t/<root-org>/o/<ORG_ID>/oauth2/token
```

The grant type, parameters, and Basic credentials are identical. Two extra conditions
apply before you get useful scopes back: the application must be shared with that
sub-organization and authorized against the API resources it needs, and the user must
hold roles carrying those scopes **in that sub-organization**. Only scopes granted to
both the application and the user appear in the response.

The same `/o/<ORG_ID>` insertion works for `/oauth2/introspect` and `/oauth2/revoke`.

Thushani-Jayasekera:
@coderabbit review

Thushani-Jayasekera:
@coderabbit review

Thushani-Jayasekera:
@CodeRabbit review

Thushani-Jayasekera:
@coderabbit review

Thushani-Jayasekera:
@coderabbit review

PR #3425, opened 2026-09-11
Let a build be deleted from the build area

dakshina99:
## Purpose
An API keeps a limited number of builds. When automatic cleanup cannot reclaim one
— every stored build is held by a deployment on a gateway somewhere — deploying is
refused, and the deploy page offered no way out of it.

## Approach
Delete lives in the **Build Area**, whose `renderBuilds` is shared by the inline
list and the All Builds drawer, so both get it from one change.

- The trash action asks before acting, inline rather than in a modal (the drawer
  would otherwise have to stack one inside).
- It is disabled, with the reason and the environment named, while a gateway is
  holding the build — `DEPLOYED`, `DEPLOYING` or `UNDEPLOYING`. Suspended and
  failed deployments are deliberately not in that set: their builds are exactly the
  ones automatic cleanup will not reclaim, so they are what this is for.
- The disable is a courtesy, not the gate. A gateway may have claimed a build since
  the page loaded, so the platform's refusal is what decides, and it surfaces
  through the existing action toast.
- A build's `description` now shows under its id, which is what makes a list of
  builds something you can choose from.

`deleteBuild` calls `DELETE /rest-apis/{apiId}/builds/{buildId}` on the API's own
resource rather than through the project path — a build belongs to the API, not to
a pipeline — following the `GET /rest-apis/{apiId}` call already in this client.

The second commit drops the build-id card from the gateway card in favour of a
plain "Deployed <when>" line. The build a gateway runs is already shown once on the
environment, so repeating it per gateway only contributed an `ID` label with
nothing beside it once the build had been reclaimed.

Also removed two comments left behind when the per-gateway redeploy was dropped: an
orphaned block documenting the removed `redeploy` method, and a "Deploys to one
gateway" block sitting above the bulk-deploy function.

## Automation tests
`tsc --noEmit` passes. Verified on a local cluster against an API sitting at the
limit with five builds: the three held only by superseded deployments offer delete,
and the two a gateway is serving are disabled with the reason.

## Related PRs
Stacked on #3421 — its commits appear in this diff until it merges; only
`Let a build be deleted from the build area` and `Show when a gateway was deployed
instead of a build id card` belong to this PR. Needs the build delete endpoint and
the build `description` from #3364.


dakshina99:
Folded into #3421 — both commits are now on `apip-deploy-bulk-ui`, so the deploy page UI work reviews as one PR.

